Issue #100, opened 2026-07-12
feat(fw+ha): WiFi network switching from the HA dashboard — dual-slot credentials + CFG key + un-brickable fallback


jphein:
JP (2026-07-12, during the roam→iot migration): switch the fleet's WiFi network from the HA dashboard.

**Motivation:** today's move off the FT roam SSID (the DHCP-deaf/FDB incident) required USB access to all three boards because an OTA'd SSID change to an unreachable network = remotely unrecoverable. Runtime network selection with a fallback makes future migrations zero-USB and un-brickable.

**Design (rides shipped machinery):**
1. **Dual credential slots baked at flash time** (secrets.rs: `WIFI_NETWORKS: [(ssid,pass); 2]` — e.g. roam + jplovescl/iot). Secrets never cross MQTT.
2. **CFG key `N`** (#56 keyed channel) = active-slot index. RETAINED/cached config (unlike R/W commands — this is state), applied on next association; node reboots into the selected network. Fleet-wide via CFG_TARGET_ALL=255 or per-node.
3. **NVS persistence** — own record/sector (identity-record pattern from #40; do NOT collide with identity@0 / floor@4096-8192 / boot-count@sectors-3-4).
4. **UN-BRICKABLE FALLBACK (the load-bearing piece):** if association on the selected slot fails N consecutive attempts (~3 × 30s), auto-revert to the other slot, rejoin, and publish a `net=fallback` DIAG flag so HA shows the switch failed. A wrong selection self-heals.
5. **HA side:** per-node `select` (roam / iot) + a fleet-wide selector; current network + fallback state surfaced from the DIAG record (#70 — add a `net=` key).
6. **Broker/OTA host per network:** each slot carries its OWN broker IP (own-VLAN leg rule: roam→10.0.11.110, iot→10.0.8.111) + OTA host — the slot is a (ssid, pass, broker, ota) tuple.

**Notes:** per-network broker legs are REQUIRED (cross-VLAN CONNACK drops — see the quad-homed-broker lesson). The mesh channel follows the elected gateway's AP channel as today; leaves' scan list must cover both networks' channels.

Cross-refs: #56 (CFG channel), #70 (DIAG), the FT-roam incident (2026-07-12), #52/#71 (command-key patterns), #40 (NVS record pattern).

jphein:
**Scope addition (JP, 2026-07-12): the OTA host/URL must be dashboard-manageable too.**

Design note — two different things here:
[...]

Issue #40, opened 2026-07-10
feat: leaf-mesh-OTA — update ESP-NOW-only leaves over the mesh (gated on #32 signing)


jphein:
Leaves are ESP-NOW-only: only the elected gateway bursts WiFi/MQTT, so a leaf can't receive an OTA announce/install today (it updates only via USB or by becoming gateway). This tracks delivering firmware to leaves **over the mesh**.

**Hard prerequisite: #32 ed25519 signing.** The ESP-NOW mesh is unauthenticated, so a leaf accepting firmware over it MUST verify an ed25519 signature over the image before flashing — otherwise any device on the mesh is a fleet-brick/injection vector.

Design complete + security-hardened (adversarial review closed 3 HIGH holes): leaf self-test = mesh-reachability not DHCP; verify rejects identity/low-order pubkeys; per-chunk writes are partition-scoped/bounds-checked (can't corrupt the active slot mid-transfer); HELLO carries build#; signed-freshness min-build floor blocks intermediate-replay.

Sequence: #32 (signing) → this → parallels #13 (multi-hop). ~640 KB image chunked over 250 B ESP-NOW frames + reliability/resume + verify-sig-before-flash + canary-one-leaf.

jphein:
**SHIPPED + HARDWARE-VERIFIED — PR #66 merged to main @ d7f3ec9 (2026-07-11 night).**

Signed leaf-mesh-OTA works end-to-end on hardware: ed25519-verified images relayed over ESP-NOW to WiFi-less leaves — **6 successful deliveries across all 3 boards** in final validation, including a full ~1MB mesh relay in ~7 min and gateway self-OTA in ~55s. Identity survives OTA (runtime NVS node-id, one shared image serves the fleet), brick-safe throughout (0 unrecoverable bricks across ~15 canary cycles; every rollback recovered), clean Tier-2 confirms (staleness-gated).

The road: ~12 distinct bugs found + fixed across arm-chain, MAC resolution, channel coexist (pre-fetch-arm), brick-safety, NVS sector collision (adversarially caught pre-merge), retained-topic drain truncation, identity theft, and observability ghosts — each pinned by purpose-built diagnostics (armdiag/relaydiag/LDBG/leaf_ch). Follow-ups tracked: #76 (election split-brain, exposed by working OTA reboots), #68 continuation (pending-gate race, self-MAC filter, F2), #69 (IRAM perf), #39 (retained install unification).

PR #108, opened 2026-07-12
feat(fw): #100 Stage 1 — runtime network switching (dual slots, CFG key N, un-brickable fallback)

jphein:
JP's #100, Stage 1 (WiFi core). Built by morpheus-quartet2 (design + implementation through code-complete; hit the weekly quota pre-build), finished by team-lead (orphaned-deadline fix + clippy).

**What ships:**
- **Dual baked NetSlot slots** in git-ignored secrets (`(ssid, pass, broker_ip, broker_port, ota_hosts)` per slot; secrets.rs.example templates the shape).
- **CFG key N** — retained per-node/fleet network selection; NVS-persisted (**sector 5**, own-sector per the #40 brick discipline; corrupt/unseeded → slot 0).
- **The un-brickable fallback** (the load-bearing safety): assoc-timeout-only revert after 3×30s attempts; **two-level revert gating** (NVS cross-reboot flag + in-RAM per-boot) = ≤1 revert-write per selection, **provably zero flash wear in total outage** (STAY-PUT: both-slots-fail → keep retrying active at slow cadence, no further writes); `net=`/`fallback` surfaced in DIAG.
- Runtime broker per slot (IP-only v1; hostname deferred pending DNS-from-DHCP verification).

**Verify:** 4 profiles compile + espnow clippy -D warnings = 0. Default-build invariant via cfg-gating.
**HW canary (team-lead, before/at merge):** select a bogus slot → fallback reverts + DIAG shows net=fallback → then a REAL both-way switch (slot0 jplovescl ↔ slot1 roam) → clean migrations. The bench proof IS the feature's acceptance test.

Stage 2 (CFG broker override) + Stage 3 (OTA-host override) follow per the signed-off staging.

### DIFF
--- a/rust/clock/src/main.rs
+++ b/rust/clock/src/main.rs
@@ -1188,7 +1188,20 @@ fn main() -> ! {
                                 commanded: slot,
                                 fallback: false,
                             });
-                            esp_hal::system::software_reset();
+                            // #100 canary lesson: VERIFY-AFTER-WRITE before the reset. If the
+                            // record didn't persist (any flash error — all swallowed above), a
+                            // reset would boot back on the old slot, re-receive the retained
+                            // command, and loop forever (~20 s reset cycle, observed on HW).
+                            // A failed readback = abort the switch this cycle; the retained
+                            // command retries next cycle, and the failure stays VISIBLE
+                            // (commanded stays != slot) instead of becoming a reboot loop.
+                            let persisted = crate::ota::read_net_cfg()
+                                .is_some_and(|c| c.commanded == slot);
+                            if persisted {
+                                esp_hal::system::software_reset();
+                            } else {
+                                log::warn!("smol #100: net record did NOT persist — switch aborted (no reset)");
+                            }
                         }
                     }
                 }
--- a/rust/clock/src/main.rs
+++ b/rust/clock/src/main.rs
@@ -1167,6 +1167,33 @@ fn main() -> ! {
                 log::info!("smol #45: custom screen layout updated ({} B)", n);
             }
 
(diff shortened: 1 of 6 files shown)


jphein:
🔴 **HW ACCEPTANCE: RED — DO NOT MERGE.** The bench canary caught what both static lenses (all-green) could not:

**Finding 1 — the NVS net-record write/read roundtrip fails on hardware.** Commanded N=1 → the board writes NVS + software-resets → **every subsequent boot reads slot 0** (net=0:ok, never net=1). Either the sector-5 write silently fails or the readback/parse rejects it (the corrupt→slot-0 safety then engages — masking the failure as a clean slot-0 boot).

**Finding 2 — the APPLY path has no loop guard.** Because the write never lands, the retained `config/net=1` re-triggers the edge-trigger on EVERY boot → write → reset → boot(slot 0) → re-trigger: a ~20-second **infinite reset loop** (observed: boot 4→20 before I cleared the retained trigger). The two-level gating protects the REVERT path; the APPLY path needs its own idempotence — e.g. **verify-after-write**: read the record back before resetting; a failed readback = abort the switch + latch a `net=<cur>:werr` DIAG flag instead of resetting.

**Finding 3 (positive) — the F1 cross-check PASSED:** id9's node identity survived ~18 reset cycles with sector-5 writes attempted each time — the s5≠s0 segregation holds empirically.

Board recovered by clearing the retained trigger (not bricked — each boot was functional on slot 0; the loop was externally driven). Debug next: the write path vs parse_net_cfg (offset/format/checksum mismatch, or partition-handle scope). oracle's static green stands for what it covered; this is why the HW gate exists.

Issue #74, opened 2026-07-12
feat(fw+ha): observability wave 2 — mesh topology + RSSI matrix + config-drift + display mirror


jphein:
**From JP (2026-07-11):** "more things exposed to the HA dashboard." Wave 2, after #70/#71. Analysis in `scratch/smol-ch-research/nebula-io-registry.md` §3 (Nebula).

Already covered — **NOT duplicated here:** #70 (uptime/boot-count/reset-reason/boot-slot/last-OTA-outcome/button-events/heap), #71 (WiFi AP scan), #50 (screen:page readback, closed), #58 (ESP-NOW RSSI presence), #64 (gateway uplink RSSI), #49 (DIAG counters topic), #48 (LED control), #52 (reboot).

New, high-value (all ride the compact-DIAG vehicle #70/#49 already call for):

1. **Leaf↔leaf RSSI matrix → mesh topology map.** Each node already builds a `Roster` of heard peers with RSSI (`mode.rs`, feeds the #8 Bench mesh-view). Publish each node's neighbor list (`id:rssi` pairs) compactly → HA renders the full **who-hears-whom adjacency + link-quality graph**. *Extends #58* (presence → full matrix); the matrix is the new thing.
2. **Mesh topology / ownership view.** Who's gateway, which leaf is locked to whom, election state, and **which channel each node is currently on** (directly useful after the #40/#23 channel work). Feeds off the RSSI matrix + `elected_owner` + per-node channel. *Extends #63* (takeover badge → full graph).
3. **Config-drift detection.** HA knows what it pushed (screen/plugins/LED/pin-map); each node echoes a **config-hash of what it's actually running** (STAT already reports screen:page). HA flags "node live ≠ desired" → catches dropped relays / stale NVS / undeployed config. Operationalizes the recurring *live≠repo divergence* lesson. **Genuinely new + high value.**
4. **OTA history ring.** #70 has *last* OTA outcome (scalar); add a small NVS ring of the last N events (build→build, outcome, duration) → an audit trail. Motivated by the silent-rollback pain of 2026-07-11.
5. **Text display-content mirror.** Publish the current screen's rendered **text lines** (not a bitmap — too heavy for the frame budget) as `smol/<id>/display` → HA shows a live text mirror of each OLED. Cheap, rides STAT-like. (A camera/bitmap is deliberately rejected on frame budget.)

Fold-ins (extensions of existing issues, not separate):
- **LED state readback** → into #48 (control + readback halves).
[...]

PR #110, opened 2026-07-12
feat(#100): Stage 2/3 — runtime broker + OTA-host overrides (dashboard-managed, un-brickable)

jphein:
Builds on the merged Stage-1 slot switcher (#108). Layers the two remaining runtime overrides JP scoped in #100 — the **MQTT broker leg** and the **OTA-host allowlist** — both dashboard-managed, fallback-protected, sharing the one sector-5 NVS record. Closes the rest of #100.

## What ships
- **Stage 2 — broker override (CFG key `B`, retained `smol/<id>/config/broker`)**: value `ip` or `ip:port` (RFC1918-gated, IP-only v1, port default 1883; empty = clear). Apply = edge-triggered + verify-after-write + reboot. `active_broker()` prefers the override unless it has been auto-disabled.
- **Stage 3 — OTA-host override (CFG key `O`, retained `smol/<id>/config/ota_host`)**: one extra RFC1918 host appended to the fetch allowlist (base is now the **active slot's** `ota_hosts`, was a slot-0 const). Read at gate time — **no reboot**, no fallback (a bad host can only refuse a fetch; the image stays monotonicity-gated).
- **HA** (`ha/packages/smol_mesh.yaml` + new `ha/dashboard/smol_network_card.yaml`): per-node network `input_select` + broker/OTA-host `input_text` + publish automations (retained, **qos 1**, guarded against unavailable/unknown) + `net`/`brk`/`otah` DIAG mirror sensors. Repo copy only — team-lead deploys to live HA.

## NVS record v1 → v2 (back-compatible)
One sector-5 record (25 B). A v2 firmware reading a Stage-1 **v1** record treats it as "no overrides" (slot selection preserved — no data loss on OTA upgrade). A v1 firmware reading a v2 record rejects it → slot 0 (a safe rollback). No forced migration; the first v2 write upgrades in place. `NetCfg` gains `broker: Option<([u8;4],u16)>`, `broker_fallback: bool`, `ota_host: Option<[u8;4]>`; ext fields carry their own checksum + complement guard.

## Invariants (identical discipline to the Stage-1 WiFi fallback)
- **Un-brickable**: the active slot's *baked* broker is the floor. A wrong broker override self-heals without USB — 3 consecutive failed CONNACKs (in-RAM streak across bursts, counted only *after* WiFi association succeeds, so a WiFi flap never counts) → ONE NVS write disables the override (`broker_fallback=true`) → baked broker next flush.
- **Never-loop**: CFG-`B` edge-triggers on the *commanded* broker, not the runtime one. A fallback-disable keeps the commanded value (for DIAG + the edge-trigger), so the retained command never re-fires → no reboot/flush loop (the exact class the #100 canary caught for CFG-N). Escape a fallen-back override with a *new* value.
- **Never-wear**: `broker_fallback` persists in NVS and gates further writes → provably ONE revert-write per commanded broker, zero across reboots during a sustained outage.
- **CFG-`N` resets identity**: a WiFi-slot switch clears both overrides (they were set for the old network's VLAN; a stale broker would drop CONNACK cross-VLAN).
- **Default build byte-free**: `ota.rs` + `secrets.rs` are `#[cfg(feature="wifi")]` modules — not compiled without a radio. Proven by cfg-gating (not ELF equality — build.rs embeds a git stamp) AND a green default-profile clippy.

## Verification
All 5 profiles clippy `-D warnings` = 0: **default / wifi / espnow / wled / cast**. Parse-checked (rustfmt). Not yet hardware-run — team-lead canaries before merge.

## Canary asks (HW, before merge)
1. Bogus broker via CFG-`B` → 3 CONNACK fails → `brk=fb` in DIAG + baked broker reached, telemetry resumes, **no reboot loop**.
2. Real broker override round-trip: CFG-`B` → reboot → `brk=ovr` → reachable.
3. CFG-`O` host → OTA fetch from the override host succeeds (gate accepts).
4. CFG-`N` slot switch clears a prior broker override (`brk=baked` after).
5. v1→v2 upgrade: a Stage-1 board OTA'd to this fw keeps its slot selection.

## ⚠️ One thing to verify at canary
The HA network `input_select` maps **"iot (jplovescl)" = slot 0, "roam" = slot 1** — this MUST match the local `secrets.rs` `WIFI_NETWORKS` order. A wrong pick self-heals (fallback → `net=<slot>:fb`) so it never bricks, but the dashboard label would mislead. Commented in `smol_mesh.yaml` + the card.

## Record migration (review note a)
`encode_net_cfg` **always writes v2** (`NET_VERSION=2`), so the record is **upgraded in place on the first write** — where "first write" = the first CFG-`N`/`B`/`O` apply *or* a WiFi assoc fallback after the upgrade. Until such a write happens a Stage-1 **v1** record is left untouched on flash and read as "no overrides" (slot selection intact). There is no separate migration pass and no "preserve-v1-until-override-set" special case: the very next thing that changes the record (including a plain slot switch or a fallback) rewrites the whole record as v2.

## CONNACK counter isolation across CFG-N (review note b)
Confirmed safe by two independent mechanisms: (1) `BROKER_FAIL_STREAK` is a plain in-RAM `static mut` (not `rtc_fast`/persistent), and every commanded change — CFG-`N` and CFG-`B` alike — ends in `software_reset()`, which clears RAM → the streak is 0 after any switch. (2) CFG-`N` also clears the broker override (`broker=None`), and `note_broker_connect` early-returns whenever `broker.is_none()`, so even a hypothetical surviving counter could not act on the new slot's baked broker. A slot change never inherits the other broker's strikes.

## Design note — self-referential gateway demotion (HW-observed)
A gateway whose broker override is unreachable can't reach the broker to confirm its own mesh-channel (MC) ownership, so it demotes itself to "leaf scanning for gateway <self>" — a self-reference (it's looking for the gateway it *is*). The Stage-2 3-strike CONNACK latch is what breaks the loop: after 3 failed CONNACKs it clears the override → the baked broker is reachable → MC ownership re-confirms → it re-crowns. Same self-heal guarantee as the WiFi fallback, one layer up.

## Post-canary fixes (see PR comments for detail)
1. `652f9f2` — drain-order: batt/grid/mc subscribe last so the retained-drain break-gate can't fire before the config backlog drains (fixed CFG-B/O never applying).
2. `1b52456` — word-align the v2 NVS record 25→28 (esp-storage NorFlash WRITE_SIZE=4; a non-aligned write returned NotAligned → silent non-persist → verify-after-write aborted every apply). HW-proven on id8: apply→persist→reboot→3-strike fallback→brk=fb.

### DIFF
--- a/ha/dashboard/smol_network_card.yaml
+++ b/ha/dashboard/smol_network_card.yaml
@@ -1,0 +1,111 @@
+# smol network identity — dashboard-managed WiFi slot + broker + OTA host (#100)
+# ---------------------------------------------------------------------------
+# HA PUBLISH/GUI half of the #100 runtime network switcher. Per node you set:
+#   • Network  → retained smol/<id>/config/net = 0/1 (WiFi slot; edge-triggered reboot)
+#   • Broker   → retained smol/<id>/config/broker = ip[:port] (reboot; CONNACK fallback)
+#   • OTA host → retained smol/<id>/config/ota_host = ip (no reboot; extra allowlist host)
+# The board applies each on its next burst (no reflash) and self-heals a bad WiFi slot or
+# broker WITHOUT USB (the un-brickable fallback — surfaced live below as net=/brk= state).
+#
+# ⚠️ NETWORK MAPPING: "iot (jplovescl)" = slot 0, "roam" = slot 1 — this MUST match the
+# local secrets.rs WIFI_NETWORKS order. A wrong pick self-heals (net shows "<slot>:fb") but
+# the label would mislead; verify against secrets.rs at canary. NEVER switch a power-only
+# board's WiFi to a network you can't reach it on without confirming the fallback slot works.
+#
+# Broker/OTA-host overrides are RFC1918-validated on the board; an empty value CLEARS the
+# override (back to the active slot's baked broker / allowlist). A broker override that fails
+# to CONNACK 3× auto-disables (brk=fb) and reverts to the baked broker — re-type a NEW value
+# to retry. A bad OTA host can only refuse a fetch (the image stays signed/monotonicity-gated).
+#
+# Install at DEPLOY time via the WS lovelace/config/save API — do NOT hand-edit
+# .storage/lovelace* (HA caches + overwrites). See ha/README.md. (Deployment HELD for JP.)
+type: vertical-stack
+cards:
+  - type: markdown
+    content: >
+      ## smol · network identity
+
+      Switch each node's **WiFi network**, **MQTT broker**, and **OTA host** from the
+      dashboard — published as retained MQTT, applied on the node's next burst, no reflash.
+      A wrong WiFi slot or broker **self-heals without USB** (the un-brickable fallback);
+      the live state below shows `:fb` when a fallback has fired.
+
+  - type: entities
+    title: Set all nodes
+    show_header_toggle: false
+    entities:
+      - entity: input_select.smol_all_network
+        name: Network (all)
+        icon: mdi:wifi-sync
+
+  - type: entities
+    title: id7 · Dominion (gateway)
+    show_header_toggle: false
+    entities:
+      - type: section
+        label: Network
+      - entity: input_select.smol_7_network
+        name: WiFi network
+      - entity: sensor.smol_7_network
+        name: Active (net=slot:state)
+        icon: mdi:wifi-check
+      - type: section
+        label: Broker override
+      - entity: input_text.smol_7_broker
+        name: Broker ip[:port] (empty = baked)
+      - entity: sensor.smol_7_broker_state
+        name: Broker state
+      - type: section
+        label: OTA host override
+      - entity: input_text.smol_7_ota_host
+        name: Extra OTA host ip (empty = slot)
+      - entity: sensor.smol_7_ota_host_state
+        name: OTA host state
+
+  - type: entities
+    title: id8 · Nexus
+    show_header_toggle: false
+    entities:
+      - type: section
+        label: Network
+      - entity: input_select.smol_8_network
+        name: WiFi network
+      - entity: sensor.smol_8_network
+        name: Active (net=slot:state)
+        icon: mdi:wifi-check
+      - type: section
+        label: Broker override
+      - entity: input_text.smol_8_broker
+        name: Broker ip[:port] (empty = baked)
+      - entity: sensor.smol_8_broker_state
+        name: Broker state
+      - type: section
+        label: OTA host override
+      - entity: input_text.smol_8_ota_host
+        name: Extra OTA host ip (empty = slot)
+      - entity: sensor.smol_8_ota_host_state
+        name: OTA host state
+
+  - type: entities
+    title: id9 · Herald
+    show_header_toggle: false
+    entities:
+      - type: section
+        label: Network
+      - entity: input_select.smol_9_network
+        name: WiFi network
+      - entity: sensor.smol_9_network
+        name: Active (net=slot:state)
+        icon: mdi:wifi-check
+      - type: section
+        label: Broker override
+      - entity: input_text.smol_9_broker
+        name: Broker ip[:port] (empty = baked)
+      - entity: sensor.smol_9_broker_state
+        name: Broker state
+      - type: section
+        label: OTA host override
+      - entity: input_text.smol_9_ota_host
+        name: Extra OTA host ip (empty = slot)
+      - entity: sensor.smol_9_ota_host_state
+        name: OTA host state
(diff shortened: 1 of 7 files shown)


jphein:
## Build verification (all green)

| tier | clippy `-D warnings` |
|---|---|
| default | ✅ 0 |
| wifi | ✅ 0 |
| espnow | ✅ 0 |
| wled | ✅ 0 |
| cast | ✅ 0 |

- **Release builds**: `cargo build --release` (default) ✅ and `cargo build --release --features espnow` (primary feature set, fat-LTO) ✅.
- **Default-build invariant** — proven by **symbol absence**, not ELF bytes (build.rs embeds a per-commit git stamp): `nm -C` of the default release ELF matched **0** of `{ota, secrets, broker, rfc1918, net_cfg, host_allowed, active_broker, parse_ipv4, NetCfg, WIFI_NETWORKS}`; the espnow ELF matched **39** of the same set. So the default build genuinely links none of the Stage-2/3 (or any wifi) code — it's cfg-gated out, not stripped.
- Toolchain: rustc 1.96.1, target `riscv32imc-unknown-none-elf`.

Not yet hardware-run — awaiting the canary (asks in the PR description).

jphein:
## Fix: hardware canary apply-miss (CFG-B/O never applied) + rebase onto main

**Root cause (HW-confirmed):** the retained-drain loop in `mqtt_session` breaks on `got_batt && got_grid && got_mc && <400ms quiet>`. `batt`/`grid`/`mc` were the **first** subscribes (pids 1/2/3); a broker delivers retained in SUBSCRIBE order, so the three break-gate flags completed immediately, while the config retained (`net`/`broker`/`ota`, pids 16-21 — **last** subscribes) arrived in a later TCP-window chunk >400ms afterward. The drain had already exited → CFG-B/O deterministically never processed. LED (pid 10) applied because it rode the first chunk.

**Fix (`652f9f2`):** move the `batt`/`grid`/`mc` SUBSCRIBEs to **after** every config topic (still before the telemetry PUBLISH). The break-gate can no longer complete until the whole config backlog has been delivered and drained — order- and buffer-size-independent, so it doesn't matter what causes the inter-chunk gap. Pids unchanged (identifiers, not order). Boot burst (`cfg_cache=None`) skips the gateway block, so these stay effectively first there.

**Rebased onto current `main` (`49382b1`, incl #74/#109).** Resolved the expected `diag_record` collision: combined #74's mutable `rec` + `|cfg=` fold-in with this PR's `|brk=|otah=` fields. DIAG tail is now `…|net=<slot>:<ok|fb>|brk=<baked|ovr|fb>|otah=<slot|ovr>|cfg=<applied>`.

**Re-verified — clippy `-D warnings=0` on all 5 tiers** (default / wifi / espnow / wled / cast) on the rebased tree. Awaiting re-canary.

jphein:
## Hardware canary: PASSED (with findings) — id8, instrumented builds, 2026-07-12 evening

**Proven on hardware, per-path:**
- **CFG-B capture** (drain-order fix 652f9f2): retained `config/broker` captured every session — the pre-fix deterministic miss is gone.
- **Apply → persist → verify → single reboot** (alignment fix 1b52456): `broker override changed — reboot` with verify-after-write passing. (Pre-fix: `NotAligned` swallowed → *"record did NOT persist"* every apply — the second hardware bug this canary caught.)
- **Un-brickable fallback**: bogus `10.0.99.99` → exactly 3 CONNACK strikes → `broker override unreachable x3 — disabled, reverting to the slot's baked broker` → `brk=fb` in DIAG → board recovered on baked. Zero loops (edge-on-commanded held: the still-retained bogus value never re-fired).
- **CFG-O**: applied + persisted with **no reboot**; `otah=ovr` in DIAG within one burst.
- **Override round-trip**: `brk=ovr` observed live in DIAG (override active + session publishing through it).
- v2-reads-v1 compatibility ran all afternoon on the mixed fleet.

**Findings (documented, not blockers):**
1. **Fallback/takeover race** (→ folded into #114): the 3-strike broker fallback needs ~90 s, but leaves take over a silent crown in ~45 s — so a gateway with a bad broker override *always* loses the crown before self-healing. Never bricks (floor held through every scenario today), but causes minutes of election churn incl. a self-referential "leaf scanning for gateway id8" (a mute crown can't confirm its own MC claim).
2. **Retained-override resurrection**: CFG-N clears overrides in NVS, but a still-retained `config/broker` re-applies next burst. HA-side fast-follow: the net-switch automation should also publish empty retained clears for `config/broker`/`config/ota_host` (the fw self-heals either way via the fallback, but the extra reboot is avoidable).
3. Mixed-fleet note: a pre-B/O crown (e.g. build 220) drops the new keys — override commands to leaves only take effect once the crown runs this build. Expected forward-compat behavior; worth a line in the HA card description.

Merging. HA package deploys with the merged tree.

Issue #114, opened 2026-07-13
mesh: faster healing + lower UI latency (crown handover, config apply, HA staleness)


jphein:
JP's directive 2026-07-12: **make the mesh faster to heal and cut the delays to the UI.**

## Measured baselines (today's session, real hardware)
| Path | Today | Target (proposed) |
|---|---|---|
| Config apply (retained → applied, e.g. LED) | ~90 s (up to 3 bursts) | ≤ 35 s (1 burst) |
| Crown handover (owner silent → new gateway bursting) | minutes; one standoff needed a manual reset (leaves honored retained MC owner=other but nobody promoted) | < 20 s |
| Cold-start election (empty MC) | ~40–60 s | < 20 s |
| Leaf state freshness in HA (DIAG 60 s cadence + 45 s staleness gate) | up to ~2 min | < 30 s |
| Post-flash/reboot mesh rejoin | 1–2 min | < 20 s |

## Candidate levers (survey first — scope-proposal-first discipline)
- **Heal:** shorten mesh-silence detection; leaves re-read/contest the retained MC when the owner goes silent (today they defer forever — the standoff bug); crown-handover fast-path (#57's Familiar hop already proves sub-minute is possible); seq-arbitration tie-break hardening (today: double seq=1 claim race).
- **UI:** adaptive DIAG cadence (fast-follow after any change, slow steady-state); event-driven publishes on config-apply/screen-change instead of next-cycle; revisit the 45 s staleness gate vs LWT; gateway re-relay cadence (10 s) vs config-apply latency — the retained-catch happens per-burst (~30 s), the relay is not the bottleneck, the burst period is.
- **Bursts:** the ~26–35 s burst period is the master clock of every UI delay — evaluate shortening it (power/airtime trade) or adding an out-of-band fast-poll after HA publishes a command (HA could publish a transient "wake" the gateway catches mid-cycle... requires persistent session work from #101).

## Constraints
Never-loop/never-wear invariants hold; election stability > speed (no flapping crowns); power-only boards must stay un-brickable; #22's coex work will add BLE airtime — headroom matters.

Deliverable 1: a survey + scope proposal (which levers, expected wins, risk) for team-lead sign-off. Then staged implementation.

jphein:
[...]

PR #112, opened 2026-07-12
feat(fw): #22 passive BLE observer behind a default-OFF `ble` feature

jphein:
## #22 — passive BLE observer behind a default-OFF `ble` feature

Implements a **passive-only** BLE observer that shares the single ESP32-C3 2.4 GHz radio with WiFi + ESP-NOW under Espressif's software coexistence arbiter. Every hook is behind a new `ble` cargo feature (**default OFF**); the `default` / `wifi` / `espnow` / `wled` / `cast` builds are byte-free of it by cfg-gating (not ELF byte-equality — `build.rs` stamps a per-commit git string).

Builds on spike finding #1 (the `espnow + esp-wifi/ble + esp-wifi/coex` profile compiles + links clean; bt-hci 0.3.2 already in-tree).

### What it does
1. **Init** — brings up the BT controller on the *same* leaked `EspWifiController` that runs WiFi/ESP-NOW (one `esp_wifi::init` per program), after `wifi::new`/`start` so the coex arbiter is up first. `esp-wifi/ble` + `esp-wifi/coex`.
2. **Passive HCI scan, NO host stack** — drives raw HCI over esp-wifi's *blocking* `BleConnector` (`embedded_io::Write` for the two setup commands, non-blocking `next()` to drain events). No `trouble-host`/`bleps` → zero Cargo-pin risk against the frozen esp-hal/esp-wifi quartet. `LE Set Scan Parameters` (passive) + `LE Set Scan Enable`; parse `LE Advertising Report`. Never connects, never active-scans.
3. **Sightings** — heapless, no-alloc fixed table (`TABLE_CAP=24`): MAC (big-endian), RSSI, addr-type, a compact ad-type bit-summary; dedup by (addr, addr-type); TTL age-out so the count reflects *presence*, not history; oldest-evicted when full.
4. **Publish** — see the publish-path decision below.
5. **DIAG `ble=<count>`** — the distinct-devices-present scalar is appended to the existing `smol/<id>/diag` record (fixed-key, #70-contract-safe; only in the `ble` build). Plus a `smol: BLESOAK tags=.. advrpt=.. cmderr=.. heap_free=..` serial line on the DIAG cadence so the team-lead's ≥1 h soak can correlate scan activity + controller health + heap with the loss/rtt/hmin already on the wire.

### Publish path — the choice, and why
Scope said: *SMOLv1 sightings frame relayed gateway-ward **OR** direct MQTT retained `smol/<id>/ble` — pick based on what the relay path supports TODAY.*

[...]

Issue #22, opened 2026-07-08
research/feat: BLE beacon + Bluetooth-proxy roles on smol (presence/positioning)


jphein:
**JP:** every node is already an NTP/MQTT gateway (role from boot DHCP) — could they ALSO be Bluetooth proxies + advertise unique BLE beacons so we can track them (presence/indoor positioning)?

**The crux — single radio.** ESP32-C3 = one shared 2.4 GHz radio (WiFi + ESP-NOW + BLE). smol already time-shares WiFi bursts vs ESP-NOW ch6; BLE is a third tenant. Splits into two very different asks:

1. **BLE beacon ADVERTISE (be tracked): feasible.** Brief periodic adverts (iBeacon/Eddystone) fit the burst duty cycle. Each board advertises a unique id; dedicated proxies / ESPresense / Bermuda in HA do the RSSI trilateration. High value, low cost.
2. **BLE PROXY / continuous SCAN (track others, relay BLE→HA like ESPHome BT-proxy): compromised by the single radio.** A board meshing + bursting + gaming can't be a good always-on scanner → lossy part-time proxy. Better left to dedicated always-on nodes.
3. **"Track themselves":** a node can't self-locate from its own beacon — needs external anchors at known positions. Mutual RSSI ranging across the mesh = rough relative distance only; single-radio timesharing degrades it. Experiment, not precise positioning.

**Open feasibility questions (research):** esp-wifi BLE + WiFi/ESP-NOW coexistence support in `no_std` esp-hal (RAM/flash/duty-cycle cost); can advertise interleave with the existing burst scheduler without hurting mesh/flush; HA ingestion path (ESPresense/Bermuda vs a beacon→MQTT bridge); the honest positioning accuracy ceiling. Prior art: smol already runs BLE HID (Bluepad32) in the Arduino Block Digger build — but that's a different stack than the unified Rust firmware.

Verdict to produce: beacon-advertise = likely yes (easy win); full BT-proxy = likely no (radio-bound). Confirm with cited feasibility.

jphein:
**Research verdict (oracle ble-coex-feasibility.md + nebula ble-positioning-research.md — recording durably):**
[...]

PR #115, opened 2026-07-13
fix(fw): #114 Stage 1 — election heal (H1 crown-handover standoff + H2 seq-race)

jphein:
## #114 Stage 1 — heal correctness

Branch off `main@49382b1` (post-#74). Scope-proposal signed off by team-lead; full survey in
`scratch/smol-dreamteam/114-latency-scope.md`.

### H1 — crown-handover standoff (this commit) ✅
**Bug (live repro today):** a leaf that adopted a retained `MC` owner it had **never heard a HELLO
from** could defer to it forever. A forged / one-shot retained `MC|<id>` looks "alive" by its seq,
and the dead-owner takeover then waited the **full RSSI backoff** (bucket-2 = 30 s, and an
unassociated leaf defaults `my_rssi = -99` → the slowest bucket) → takeover dragged to ~65 s+, i.e.
the standoff that needed a manual MC reset.

**Fix:** track `owner_hello_seen` per leaf (set when the elected owner's HELLO is heard, reset on
owner change), plumb `owner_never_heard` into the recovery election. A **dead owner (frozen seq) the
leaf never heard** is a phantom — no live board to stagger against — so it's taken over promptly on
an id-only tiebreak (lowest id first) instead of the RSSI backoff. A **heard-then-died** owner keeps
the RSSI-weighted stagger (best survivor wins). Expected: phantom-owner heal **minutes/manual → ~35 s**.

**Safety (unchanged invariant):** takeover still requires the MC seq **frozen** past
`RECOVERY_STALE_MS` (35 s). A healthy gateway republishes its MC every ≤30 s flush, so its seq never
freezes → a live owner (even one a leaf can't hear) is **never** taken over — RF-dead-zone protection
intact. Per team-lead's constraint, 35 s sits comfortably above the **15 s `RELAY_FLUSH_BUDGET`**
mesh-deaf window, so a normal flush can't trigger a false takeover.

### H2 — double `seq=1` claim race (second commit) ✅
**Bug:** two candidates that hit an **empty MC** simultaneously each claim `seq=1` (owner=self); the
broker keeps whichever write landed last, so **both** believe they won and burst as gateway — a
split-brain that today only reconverges via the lowest-id flush resolver (~1–2 flushes / 30–60 s).

**Fix:** after a claim publish, briefly (≤400 ms, tick-abortable) re-read the retained MC. If a
different owner appears (a concurrent claim landed), resolve **deterministically by lowest id**: the
higher id **yields** immediately (no duplicate gateway); a lower id re-asserts `seq+1` so the retained
record names it. **Fail-safe:** a timeout / parse-miss keeps the original verdict (never a
regression), and a board never yields to — nor claims over — an owner it hasn't out-ranked by id. Runs
only on a claim (cold-start / takeover), never in steady state; bounded within `RELAY_FLUSH_BUDGET`.

Kept as a **separate commit** (8791dbb) so the canary can bisect the two election-path changes.

### Verification
- `cargo check` + `clippy -D warnings` clean on default/wifi/espnow/wled.
- Default build unaffected — every change is in `wifi`/`espnow`-gated election code (cfg-gating, not
  ELF equality).
- **HW pending (team-lead canary):** (H1) re-run the forged-`MC` repro — a leaf should now self-heal
  (take over) in ~35 s instead of needing a manual reset; confirm a healthy gateway is never falsely
  taken over across its normal flushes. (H2) on a cold-start with the MC empty, confirm the fleet
  converges to a single lowest-id owner without a lingering dual-gateway window.

### Merge-ladder
`net/mode.rs` is a multi-writer surface this wave (#110 broker/OTA, #112 role-enum pending). My
election/RSSI region is expected disjoint; I'll rebase #114 onto them when they land (hypnos brokering).

Refs #114. Out of scope (per sign-off): B1 flush-shorten (held behind #22 soak), finding #3
config-apply (subsumed by #110).

### DIFF
--- a/rust/clock/src/net/mode.rs
+++ b/rust/clock/src/net/mode.rs
@@ -1591,6 +1591,15 @@ pub struct RadioManager {
     scan_idx: usize,
     last_scan_hop_ms: u64,
     last_owner_heard_ms: u64,
+    /// #114 H1: has this leaf heard a HELLO from its CURRENT `elected_owner` since adopting it?
+    /// Reset on every owner change, set in the HELLO arm. Distinguishes a NEVER-heard dead owner
+    /// (a forged / phantom retained `MC` — the #114 crown-handover standoff) from a heard-then-died
+    /// owner. A never-heard dead owner has no live board to stagger against, so recovery takes it
+    /// over PROMPTLY (id-only tiebreak, skipping the RSSI backoff meant for a possibly-live vacated
+    /// owner); a heard-then-died owner keeps the RSSI-weighted stagger. Best-effort: the HARD safety
+    /// gate is still the FROZEN-seq test (a live owner's seq advances every ≤30 s flush), so a stale
+    /// value only changes WHICH backoff is used — never causes a false takeover of a live owner.
+    owner_hello_seen: bool,
     /// #29: the gateway's operating ESP-NOW channel, LEARNED from the `rx_control` of received
     /// frames (in `service`) — 0 until the first frame is heard. Published in the retained
     /// `MC|owner|<ch>|seq` election record (via `current_channel`) so a re-electing / roaming leaf
@@ -1749,6 +1758,7 @@ impl RadioManager {
 
             last_scan_hop_ms: 0,
             last_owner_heard_ms: 0,
+            owner_hello_seen: false,
             mc_seen_owner: 0,
             mc_seen_seq: 0,
             mc_seen_ms: 0,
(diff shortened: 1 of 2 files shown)


jphein:
## Review + canary record — APPROVED, merging

**Code review (election path, line-by-line):** H1's `owner_hello_seen` only selects *which backoff* applies — the frozen-seq RECOVERY_STALE_MS gate remains the sole alive-test, so a live owner (even RF-unheard) is never falsely taken over. Reset-on-owner-change is present at both adoption sites; heard-then-died keeps the RSSI stagger. H2's post-claim re-read is bounded (400 ms, tick-abortable), handles the self-echo (`owner2 == node_id` falls through), converges from both sides of the race (higher yields, lower re-asserts seq+1), and every failure mode collapses to today's behavior.

**Hardware evidence:** id8 ran this build live in the mixed fleet for 30+ min with zero election regressions (leafed correctly under an old-fw crown, rejoined cleanly through multiple churn cycles). The *positive* H1/H2 behavioral proof is mixed-fleet-untestable: a live crown reclaims instantly over a planted phantom (correct old behavior), and the only old-fw gateway's 60 s reboot round-trip re-enters every takeover window I can stage. Full behavioral verification lands at the post-train fleet flash (2+ nodes on this build) — tracked as the fleet-flash checklist item on #114.

Also captured on #114 from today's live incidents: the broker-fallback (~90 s) vs crown-takeover (~45 s) race — H1/H2 are the right substrate for addressing it.

Issue #70, opened 2026-07-12
feat(fw+ha): expanded per-node observability — uptime, boot-count, reset-reason, boot-slot, button events


jphein:
Expose more device state to HA per node (JP, 2026-07-11). Tonight's #40 debugging is the motivating case: we hand-rolled armdiag/relaydiag/LDBG because the fleet had no built-in visibility — **a silent rollback was invisible** until we read bootloader logs over USB.

**Per-node sensors (STAT-relayed like build#):**
- **uptime** (secs since boot) — a reset shows as uptime dropping → catches silent reboots/rollbacks instantly
- **boot-count** (NVS-persisted) + **reset-reason** (POR / SW / panic / brownout / WDT)
- **boot-slot** (ota_0/ota_1) + last OTA outcome (confirmed/rolled-back) — productionizes tonight's diag
- **button events** — BOOT-button presses as an HA event entity (press/long-press), relayed from leaves
- heap-free / min-heap watermark (cheap, catches leaks)

Payload budget: STAT is capped (16B value) — likely a second compact DIAG frame/topic instead of growing STAT. **Cross-link #49** (retained MQTT DIAG counters — same vehicle; this extends its field list). HA side: discovery for the new sensors + node-box rows.

Sequencing: after #40 merges (rides the same STAT/relay plumbing).

jphein:
**HA-side fw-impl flag (from the #70 HA-composition design — luna, 2026-07-11):**

When implementing the observability firmware, publish **ONE** compact retained record on `smol/<id>/diag` (key=val pairs covering the whole signal set) — do **NOT** also emit per-signal MQTT discovery configs.

HA ingests by parsing that single record package-side (`mqtt:` template sensors), matching the existing smol package idiom (battery / grid / ota_diag / relaydiag are all package sensors). Emitting **both** the record AND per-signal discovery would **double the entities**.

A ready-to-apply HA snippet is staged + validated (21 sensors incl. the silent-rollback trio ota_outcome/boot_slot/reset_reason, + 3 button events; panic-free parser, degrades safely on missing keys) — drop-in when the fw publishes the topics.

PR #117, opened 2026-07-13
feat(fw): #114 Stage 2 — event-driven DIAG expedite on config-apply (UI freshness)

jphein:
## #114 Stage 2 — UI freshness (event-driven DIAG expedite)

Independent of Stage 1 (PR #115, election-heal) — this touches only `main.rs` (disjoint files), off
`main@49382b1`. Scope signed off; survey in `scratch/smol-dreamteam/114-latency-scope.md`.

**Bug:** a leaf's applied config (LED / units / plugin-mask) rides the DIAG record's `cfg=` field
(#74), but DIAG only broadcasts on the slow 60 s steady cadence. So an HA-pushed config change took
up to ~60 s to show as *applied* and to clear `binary_sensor.smol_<id>_config_drift`.

**Fix (U2):** set the existing `diag_dirty` flag when an applied LED/units/plugin-mask value actually
**changes**, so the already-present expedite path broadcasts an updated DIAG within
`DIAG_EXPEDITE_MIN_MS` (~3 s). Leaf freshness after a change: **~60 s → ~3 s**.

- Scoped to the `cfg=`-carried fields; **edge-triggered** on a real change (re-applying the same
  value is a no-op).
- **Rate cap (team-lead constraint): ≤1 expedited diag per ~5 s per node.** `DIAG_EXPEDITE_MIN_MS`
  raised 3 s → **5 s** (commit eba19cf) — this floor bounds BOTH the #70 BOOT-button expedite and the
  #114 config-apply expedite, so a flapping input (e.g. a misbehaving HA automation) can't flood the
  mesh. **Coalesced:** `diag_dirty` is a single latch, so any burst of changes between broadcasts
  collapses into ONE expedited diag carrying the latest state. **Steady-state cadence
  (`DIAG_CADENCE_MS` = 60 s) is unchanged.**
- Manual screen-nav is deliberately **not** hooked (already covered by the 10 s STAT cadence; avoids
  nav-spam). The adaptive multi-cycle fast-follow window (U1) is a possible follow-on — this single
  expedite delivers the bulk of the win.

**Verification:** `cargo check` + `clippy -D warnings` clean on default/wifi/espnow/wled; default
build unaffected (change is in the espnow config-apply path — cfg-gated). HW: after landing, a config
push should reflect in HA (applied + drift-clear) within a few seconds instead of up to a minute.

Refs #114. Pairs with #74 (`cfg=` producer) and #115 (Stage 1 election-heal). Out of scope: B1
flush-shorten (held behind #22).

### DIFF
--- a/rust/clock/src/main.rs
+++ b/rust/clock/src/main.rs
@@ -1197,6 +1197,10 @@ fn main() -> ! {
                 if let Some(m) = led::LedMode::from_wire(core::str::from_utf8(&o.buf[..o.len]).unwrap_or("")) {
                     if m != led_mode {
                         log::info!("smol #48: LED mode -> {:?}", m);
+                        // #114 U2: an APPLIED config change rides the DIAG record's cfg= field —
+                        // expedite the next DIAG so HA's config-applied/drift updates in ~3 s
+                        // (DIAG_EXPEDITE_MIN_MS) instead of waiting up to the 60 s steady cadence.
+                        diag_dirty = true;
                     }
                     led_mode = m;
                 }
@@ -1210,6 +1214,7 @@ fn main() -> ! {
                 if let Some(u) = units::Units::from_wire(core::str::from_utf8(&o.buf[..o.len]).unwrap_or("")) {
                     if u != units {
                         log::info!("smol #43: display units -> {:?}", u);
+                        diag_dirty = true; // #114 U2: expedite DIAG (units ride cfg=) — see LED above
                     }
                     units = u;
                 }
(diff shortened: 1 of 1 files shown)


jphein:
## Canary — leaf-side mechanism PROVEN, merging

Serial evidence (id8, leaf, instrumented build @542d45e): config relay arrives (`CFG frame for us (key 'L')`) → **`smol #114: diag expedited (config change, 38299 ms since last)`** — the expedited broadcast fired immediately on apply rather than waiting out the 60 s cadence.

**Honest scope:** end-to-end HA latency remains crown-quantized (measured 126 s/162 s in the mixed fleet — the crown's 30 s flush + republish path dominates), so this is the necessary leaf-side half; the full UI win arrives with B1 (flush-shorten, parked behind the #22-resolved coex budget — now unblocked by the #22 refute, pending a power soak) and a new-fw crown. Decomposition captured on #114.

Note: I rebased the branch onto current main and added the observability line myself (author quota-parked; announced). Rate-cap (≤1/5 s, coalesced latch) verified in source.

Issue #119, opened 2026-07-13
fleet image: promote cast (+io?) into the canonical reproducible build


jphein:
The canonical staged image (tools/repro_build.sh:100) builds `--features espnow` only. Two shipped features are therefore bench-only until promoted:
- **cast** — gates #26 WLED-cast AND #74's display-mirror: the fleet's crown can't publish `smol/<id>/screen` on the OTA image today.
- **io** — #72's registry (G/g keys, dollhouse lamps): inert without a G config, so promoting it costs ~0 runtime risk and makes the dollhouse dashboard-only (no reflash to bind pins).

Promotion = one line in repro_build.sh + a #44 note (the reproducible-image definition changes → new sha lineage) + slot-fit check (233 is 979 KB of 1.94 MB — huge headroom). Decide + do in one small PR; the fleet picks it up on the next stage/install cycle.

PR #120, opened 2026-07-13
fix(#111): retain OTA install until version-flip / staged-known (no lost orders)

### LINKED ISSUE #111: OTA: install orders silently lost across gateway handover (in-RAM leaf-feed pending + eager retained self-clear)
## Observed (2026-07-12, live fleet)

Armed retained \`INSTALL\` orders for id8 + id9 (\`ota_publish.sh install 8/9\`) while all three boards saw \`staged=220\`. **Neither board ever installed.** Forensics:

- The retained \`smol/8|9/ota/install\` topics were **consumed** (self-cleared on catch — correct per-catch behavior).
- armdiag ghosts showed the orders were caught **crosswise** by whichever node wore the crown at that moment (\`id8: install-caught=9 pending=9 leaf_ota=1\`, \`id9: install-caught=8 pending=8 leaf_ota=1\` — two different gateway tenures).
- An election reshuffle (id7 rebooting into its own install at 14:02) replaced the crown; the **pending leaf-feed queue is in-RAM only**, so both feeds died with their gateway tenure.
- Net effect: order consumed + no install + no error surfaced anywhere. The HA Update button's install (and the CLI path) is fire-and-forget-and-maybe-lost whenever the crown moves between catch and feed-completion.

(Separate smaller race, same day: a rebooting board can catch its own retained INSTALL **before** it has parsed \`ota/staged\`, fail the \`staged > running\` gate, and self-clear the order — id7 did exactly this at 13:56; a manual re-arm landed it.)

## Why it matters
Leaf-mesh-OTA (#40) made installs reliable *within* a stable crown tenure. Election churn — which the install itself can cause — is exactly when orders evaporate.

## Fix directions (pick on design review)
1. **Don't self-clear the retained INSTALL until the feed completes** (leaf version-flip confirmed). Idempotence already exists (\`staged.build > running\` gate), so a re-caught order after handover is harmless and self-heals.
2. Persist pending feeds crown-portably: gateway mirrors its pending set to a retained \`smol/ota/pending\` that a new crown reclaims on take-over (same pattern as MC reclaim).
3. Self-arm fallback: a leaf that sees \`staged > running\` + a recent-but-vanished install order re-arms once.

Option 1 is the smallest and closes both races (the boot-race too: the order survives until someone actually completes it).

## Evidence
Session transcripts: armdiag ghost captures + clear-then-watch timelines, 2026-07-12 13:36–14:12 PDT.

jphein:
Closes #111. Install orders were silently lost two ways: a **crown handover** (leaf-feed pending is in-RAM, died with the gateway tenure) and a **boot-race** (a board caught its own retained `INSTALL` before parsing `ota/staged`, failed the gate, yet the eager clear-on-catch consumed it — id7, 13:56). Fix direction 1: don't clear at catch — clear on completion.

## Changes (firmware only, `net/wifi.rs` / `mqtt_session`, +69/-2)

**Own install — boot-race:** the self-clear is now gated on `install_requested && staged_raw.is_some()` (was `install_requested` alone). No staged parsed → the order stays retained → it survives the racing boot and installs on the burst that parses staged.
- Kept deliberately as a **one-shot clear-at-install, NOT version-flip.** The own self-OTA gate is `build > BUILD_NUMBER` only — it is **not** `fresh_floor`-gated like the leaf mesh path (`ota_mesh.rs`), so a version-flip clear would re-fetch a **rolled-back** image forever (own reboot loop). Clearing at install-time preserves the one-shot anti-loop property the eager clear had, while no longer losing the boot-race order.

**Leaf installs — crown handover:** a new crown-portable **version-flip cleanup** pass. For any leaf whose retained install is armed *this burst* AND whose **fresh** `stat_cache` STAT reports `build >= staged.build`, publish the empty retained clear (≤1/burst). The freshness gate (`entry_fresh`) is exactly the "STAT seen since staging" guard — a stale/absent STAT → unknown build → order left retained. ANY crown that sees the flip clears it, so an install caught by one crown and completed under another is no longer stranded.
- `record_leaf_ota`'s retry-cap / rolled-back clear **stays** as the doomed-install backstop (a rollback never version-flips; the leaf's own `fresh_floor` blocks the re-relay, and the retry cap gives up). Idempotence (`staged>running` + floor) makes a re-caught order a harmless self-heal, per the issue's option 1.

## Both evidence cases die
- Crown-handover: the leaf order re-delivers to the new crown → re-arms; the version-flip clears it once the leaf flips, regardless of which crown relayed it.
- Boot-race: no clear until staged is in hand → the order survives the racing boot.

## Design note for review (challenge outcome)
Your sketch had the version-flip clear for **self** too. I did **not** apply it to self: the own self-OTA path isn't `fresh_floor`-gated, so version-flip-clearing the own install re-fetches a rolled-back image in a loop. The staged-parsed one-shot clear fixes the boot-race without that risk. A cleaner long-term option: make the own self-OTA gate (`gate()`/`ota_offer`) also consult `fresh_floor` — then self could use version-flip safely and the rolled-back-own-image loop closes generally. I scoped that OUT of #111 (latent, separate); happy to file a follow-up.

## Verification
clippy `-D warnings=0` on all 6 tiers: default / wifi / espnow / wled / cast / **canonical espnow,cast,io** (#119). Not yet hardware-run.

## Canary asks
1. Bogus install (leaf that won't flip) → order stays retained, re-arms, no loss; retry-cap eventually gives up (rolled-back/unreachable).
2. Real install → leaf flips → order cleared once its STAT reports `>= staged` (crown-portable).
3. Crown handover mid-feed → order survives, completes under the new crown.
4. Boot-race (board catches own INSTALL pre-staged) → order survives → installs on the staged-parsed burst.

### DIFF
--- a/rust/clock/src/net/wifi.rs
+++ b/rust/clock/src/net/wifi.rs
@@ -2175,6 +2175,13 @@ fn mqtt_session(
     // not the gateway, owns the freshness gate (the OTAM handler rejects `build ≤ leaf.build`);
     // the gateway may be NEWER than the leaf's target, so a gateway-build gate would drop it.
     let mut pending_leaf: Option<u8> = None;
+    // #111: every leaf id whose retained `smol/<id>/ota/install=INSTALL` was SEEN this burst
+    // (deduped, bounded). Unlike `pending_leaf` (last-wins, one relay/burst) this is the FULL
+    // armed set — the version-flip cleanup below clears a completed leaf's order even when it is
+    // not the one being relayed this burst (e.g. it installed under a PRIOR crown tenure). Dropped
+    // extras are re-seen next burst (retained), so an 8-slot cap never loses an order.
+    let mut armed_installs = [0u8; RESET_REQ_MAX];
+    let mut armed_n = 0usize;
     loop {
         if tick() {
             break; // #20 abort during downlink wait → fall through to clean DISCONNECT
@@ -2543,6 +2550,11 @@ fn mqtt_session(
                             // the audible leaf's service rate → slow-arm); it rides the #68 pass.
                             pending_leaf = Some(leaf_id);
                             *leaf_install_seen = true; // #40 #1: install-SEEN (pre-arm) → gateway suppresses its own self-OTA
+                            // #111: record it in the armed set (deduped) for the version-flip cleanup.
+                            if armed_n < armed_installs.len() && !armed_installs[..armed_n].contains(&leaf_id) {
+                                armed_installs[armed_n] = leaf_id;
+                                armed_n += 1;
+                            }
                             log::info!("smol #40: leaf id{} OTA install command received", leaf_id);
                         }
                     }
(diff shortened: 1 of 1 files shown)


jphein:
## Hardware canary — FULL SUITE PASSED, merging

- **Test 1 (boot-race, the id7-13:56 case)**: retained INSTALL + immediate reset → order **survived** the pre-staged-parse window (`staged build 233 not newer` + `install command received` on serial), then was consumed exactly at staged-known. The old eager-clear loss mode is dead.
- **Test 2 (own-install)**: catch → fetch → flip to 300 in **47 seconds**, order cleared, no re-install loop (NotNewer idempotence held across many subsequent boots).
- **Test 3 (THE HEADLINE — crown-handover survival)**: crown id9 caught id8's order (`pending=8 leaf_ota=1`, mac-known, feed live) → **USB-killed the crown mid-feed** → the retained order **SURVIVED** → successor crown id7 re-caught it within ~25 s (`install-caught=8 pending=8`) → feed resumed and completed → **id8 flipped to 300**. End-to-end across a gateway death: the exact #111 loss scenario, now self-healing.
- **Retry-cap backstop** (kept by design): separately demonstrated when a mesh-dark leaf (`phase=mac-unknown`) hit the cap → doomed-order clear — the backstop still protects against truly-unreachable targets.

Also exercised mixed-fleet: the old-fw successor crown (233, eager-clear) completing a new-fw crown's surviving order is exactly the rolling-upgrade path.

Closes #111.

Issue #13, opened 2026-07-08
feat: multi-hop routed mesh — TTL/next-hop relay + loop-prevention + freshness-counter downlink


jphein:
Today RELAY is **single-hop only**: a leaf must be in direct ESP-NOW range of a gateway or its telemetry never reaches the internet, and the `SMOLv1 BATT` downlink only reaches a gateway's *immediate* neighbour leaves (`net/mode.rs` ~L134: *"single-hop gateway → neighbour leaves … does not propagate a stale payload or loop"*). A leaf one room too far is stranded in both directions. Multi-hop fixes reach: leaf → relay → … → gateway.

The out-of-scope note is already in the tree (`net/mode.rs` ~L691, `docs/relay.md` ~L183): *"MULTI-HOP … needs a next-hop/TTL routing header + a loop-prevention seen-set + a shared-channel invariant across every node (+200–400 LOC)."* This issue is that note, expanded.

## Prior art (cited)
**aZholtikov/ZHNetwork** — a routed, **table-free** multi-hop ESP-NOW mesh for ESP8266/ESP32 (flood + route-learning, no manual peer/MAC table) with a companion MQTT/HA gateway. It's C++, but the *routing algorithm* is the reference design; the landscape survey flagged it as the proven prior art (see the culture note on #11, referenced from #12). We port the idea, not the stack.

## Design sketch
- **Routing header on SMOLv1 frames** (protocol version bump or a new frame variant): `{origin_id, msgid, ttl}`. Decrement TTL per hop, drop at 0.
- **Loop/dup prevention:** a bounded **seen-set** ring of recent `(origin_id, msgid)` — a natural extension of the relay's existing `DONE_RING` dedup. Already-seen frame → drop, don't re-forward.
- **Uplink (RELAY):** a leaf broadcasts; any node with a route toward a gateway forwards (TTL−1, seen-set gated). RELAYACK must travel back (reverse-path unicast, or bounded re-broadcast).
[...]

Issue #76, opened 2026-07-12
bug(mesh): election split-brain under cascading reboots — all three nodes claim gateway simultaneously


jphein:
Exposed by the #40 sequential OTA sweep (2026-07-11 ~22:40): after id8 + id9 mesh-OTA'd and rebooted in sequence, the election never reconverged — **all three nodes publish `PEERS|G`** simultaneously:
```
smol/7/peers PEERS|G|0|9,208,…;8,206,…   (id7 claims G, rosters both leaves)
smol/8/peers PEERS|G|0|                   (id8 claims G, empty roster)
smol/9/peers PEERS|G|0|7,206,…            (id9 claims G, rosters id7)
```
Consequences observed: id7 (134) never updates — as a 'gateway' it waits to self-OTA while ANOTHER 'gateway' mesh-relays at it; id7's leaf-side handler rejects with `leaf=H146V3N0ch11` (heard 146 OTAMs, verdict V3, ch11 — it followed a different AP channel). Mesh effectively partitioned across AP channels.

**Scope:** mesh election/failover (#51 re-election, #31 uplink-aware failover family) — NOT the #40 OTA machinery, which passed all its gates in the same sweep (id8 self-OTA 55s ✓, id9 relay ✓, identities intact ✓, roster fix ✓). Pre-existing; newly visible because OTA-driven reboots now work and cascade.

**Repro:** sequential mesh-OTAs across the fleet (reboot cascade) → observe multiple PEERS|G publishers.
**Fix direction:** election must detect + resolve concurrent gateways (a gateway hearing another gateway's HELLO/PEERS with lower id or better uplink must abdicate — the broker-mediated `MC|owner|channel|seq` consensus from the coexist design memo is the designed answer; verify it's implemented + why it didn't converge here).
**Evidence:** relaydiag `rx=24 otan=0 otam_tx=142/142 leaf=H146V3N0ch11`; sweep log in session scratch. Cross-refs: #51 #31 #35 #29.

jphein:
[...]

Issue #10, opened 2026-07-08
feat: HA link — MQTT-native telemetry uplink + HA battery voltages on every display (retires the collector)


jphein:
Show Home Assistant battery voltages (both solar systems) on the 72×40 OLED as a new plugin-framework app (builds on #7).

## Why not call HA from the board
The HA API is **TLS-only** (native SSL on `10.0.6.108:8123`, proxy at `10.0.6.11`) and the firmware's smoltcp stack is **UDP-only** (`no_std`, no TCP/TLS). So the existing **relay collector** becomes the bridge: it polls the HA REST API server-side and answers the board over UDP during the gateway's existing WiFi bursts.

## Design (locked wire protocol)
- **Query** (board → collector, existing port 9999): ASCII `BATT?` (5 bytes), sent after the relay queue drains in a flush burst (and once at the boot NTP burst); board polls recv ≤ 1500 ms — a miss just leaves the cache stale.
- **Reply** (collector → board): `BATT|<line1>|<line2>|<line3>` — ≤ 96 B, pipe-separated **display-ready** lines (≤ 12 chars each), e.g. `BATT|48V 52.8V|HV 391.9V|d 43mV`. Values become `--` when the collector's last successful HA poll is > 300 s old or entities are unavailable. The firmware never parses floats — retargeting entities is a collector restart, not a reflash.
- **Entities** (collector defaults): `sensor.epever_battery_voltage` (48V LFP bank), `sensor.be_battery_voltage` (BMW i3 HV pack), `sensor.be_cell_voltage_delta` (cell spread). Token via `SMOL_HA_TOKEN` / `--ha-token-file` (never committed/logged); no token → HA polling off, collector unchanged otherwise.

## Workstreams (in flight, one agent each)
- [ ] firmware: `batt.rs` plugin (cfg `wifi`), `BattCache` via `Ctx`, fetch in `net/wifi.rs`, REGISTRY row "Batt" — gates: build + `clippy -D warnings` on default/wifi/espnow, default build unaffected
- [ ] collector: HA poller thread (stdlib-only), `BATT?` responder, unit tests + **live LAN proof against real HA**, `EnvironmentFile` in the systemd unit, README section
- [ ] docs: protocol.md BATT section (LAN UDP, not a SMOLv1 frame), relay.md deaf-time note (≤ +1.5 s/burst), README app row

## Follow-ups (out of scope here)
- Mesh `BATT` broadcast frame so **leaves** show the voltages too (today they render the no-data state)
[...]

PR #123, opened 2026-07-14
feat(fw): #13 routed multi-hop mesh — uplink (A) + downlink freshness (B) + rig (C)

jphein:
Closes the single-hop limitation on smol's mesh: a leaf out of direct ESP-NOW range of the elected gateway is stranded today. This adds Meshtastic-lineage **managed flood** (hop-limit + `(origin,msgid,frag)` seen-set + forward path) on the uplink, freshness-gated flood on the downlink, and a test rig to validate it — rooted at the #76-elected owner, table-free.

Design + sign-off: `scratch/13-multihop/scope-proposal.md` (+ ADDENDUM). Rig: `scratch/smol-dreamteam/13-rig-design.md`. Pure decision core: `net/flood.rs` (host-tested in `experiments/flood_verify`).

## What's in this PR

**Stage A — routed multi-hop uplink** (`8044cd4`)
- New tags `SMOLv1 RELAY2` (hop-limited, origin-stamped fragment) + `SMOLv1 RELAYACK2` (R1 flooded ACK). A stranded leaf (telemetry fully un-ACKed after `RELAY_MAX_TRIES`) escalates from plain `RELAY` to `RELAY2` at `MAX_HOP=2`.
- Relays re-broadcast via `flood::forward_decision` gated by the `(origin,msgid,frag)` seen-set; the gateway reassembles **re-keyed by origin** (a synthetic `00:00:00:00:00:<id>` MAC — a real Espressif MAC is never all-zero, so it can't alias) and floods `RELAYACK2` back.
- Un-latch hysteresis: back to single-hop after `UNLATCH_STREAK=2` **direct**-ACK probes, and only probes (1-in-8 `H=1` emit) while the owner's HELLO is heard **directly + fresh** (Gate A). DIAG += `fwd`/`dedup`/`ttl`/`hop`.

**Stage B — downlink freshness** (`5b0a4ed`)
- New tags `SMOLv1 BATT2`/`GRID2` = tag + 10-digit `dl_seq` + verbatim `BATT|`/`GRID|` payload. `dl_seq` = the gateway's unix-second of the last **value change** (survives a gateway reboot + never wraps — the TIME `synced_at` template), bumped only on a change so an unchanged re-broadcast isn't re-flooded.
- A leaf adopts + **re-floods** a frame only when its `dl_seq` is **strictly newer** than what it holds — reaching a stranded leaf via a relay and dropping replays (no loop). A gateway ignores inbound downlink (it IS the source). Leaves still parse v1 `BATT`/`GRID` from an old gateway. DIAG += `dlseq`. Closes the #10 HA-staleness gap + rig P4.
- Deferred (small follow-up, Luna's screen domain): the leaf **data-age render**. This stage delivers reach + freshness; the existing render shows the value.

**Stage C — mesh-test deaf-list rig hook** (`847bf52`)
- Default-OFF `mesh-test` cargo feature + a per-board `board::DEAF_MACS` list; `service()` drops inbound frames from listed sources before any handler — a deterministic stand-in for RF attenuation that forces a 2-hop topology (G→R→F) on today's 3 boards. DIAG += `deaf`/`ddrops`.

## Byte-identical invariant (the uplink safety gate)

A **non-escalated** leaf emits ONLY plain `RELAY` (hop 1), so in the all-hear case **no `RELAY2`/`RELAYACK2` frame ever exists and nobody forwards**. Canary gate **C0 = `fwd=0` on every node** = the measurement of this claim (run it before any deaf-list test). Proven byte-free by cfg-gating, not ELF equality. (Downlink, like TIME, does re-flood on strictly-newer `dl_seq` — bounded, and only on an actual value change.)

## Mixed-fleet / rolling-upgrade compat

All new tags are ADDITIONS (not appends to the fixed-offset `RELAY`/`RELAYACK`/`BATT`/`GRID` headers): old firmware `classify()`s them to `None` (harmless) and cannot relay anyway — no flag-day flash. **A leaf that needs H≥2 was, by definition, out of direct gateway range — i.e. already stranded pre-#13** — so no rolling-upgrade window can strand a leaf that worked before. #13 only ever ADDS reach.

## Design note — the gap I found, and why RELAY2 (not UP2) for v1

While wiring Stage A I found the signed RELAY2-only scope can't carry a **stranded** leaf's `STAT`/`DIAG` (those are separate single-hop frames), so rig **P1's `/stat`+`/diag`-via-R only partially passes** and the `deaf=` visibility on a stranded board can't reach HA. I flagged it and recommended generalizing to a single **UP2 envelope** (wrap any inner frame). Team-lead's call, and I agree: **RELAY2 stands for v1** — it's only ever on-air for a genuinely-stranded leaf, so the tag-migration cost is near-theoretical, and reworking a proven, green stage buys less than it costs. The envelope lands as a **post-v1 migration** (#124) that *replaces* RELAY2→UP2 and brings the observability reach. P1's telemetry reach passes in v1.

## Correctness fix to the inherited core (approved)

The host-tested `SeenSet` was keyed `(origin,msgid)`. RELAY messages **fragment**, so a per-message key marks the whole message "seen" on fragment 0 and a relay would drop fragments 1..N — a multi-fragment message could never reassemble. Widened to `(origin,msgid,frag)`; the `SeenSet`/`HopLatch`/`forward_decision` logic is otherwise unchanged, and `flood_verify` gained a multi-fragment regression assert.

## Known v1 limitation (documented)

Multi-hop uplink is **best-effort**. In a single-relay topology a lost `RELAYACK2` isn't recovered within the message (the relay's seen-set suppresses the retransmit forward) — telemetry is loss-tolerant, next `msgid` is a fresh flood. Two closure options for a follow-up: a **2+-relay** topology, or a **RELAYACK2 seen-set**. `RELAYACK2` loop-safety rides the `MAX_HOP=2` hop decrement (a 3-hop follow-up would add the seen-set).

## Canary checklist (team-lead)

1. **C0** (precondition): `fwd=0` on every node across a normal all-hear window = byte-identical proof.
2. Flash 3 boards `--features espnow,mesh-test`; set `board.rs` `DEAF_MACS` so **F** is deaf to gateway **G** and **G** deaf to **F** (see `board.rs.example`).
3. **P1** F's `/telemetry` reaches HA via R (note: `/stat`+`/diag`-via-R is the #124 follow-up). **P2** `fwd≈1`/msg at R, `dedup>0`, airtime steady. **P3** TTL honored (`ttl` increments, no forward past hop 1). **P4** change a BATT/GRID value → F renders it via R + `dlseq` advances; replay an older seq → dropped (`dlseq` unchanged).
4. Restore: re-flash with empty `DEAF_MACS`; confirm every DIAG `deaf=0`.

## Follow-ups

- **#124 — A.5 RELAY2→UP2 migration** (post-v1): one uplink envelope, brings `/stat`+`/diag`+`deaf=`-via-R. Constraints captured in the issue (reduced-MTU const + clamp; envelope-own msgid).
- Leaf data-age **render** for BATT2/GRID2 (Luna's screen domain).

## Verification

- `cargo clippy -- -D warnings` clean: default / espnow / espnow,cast,io (#119 canonical) / espnow,wled / espnow,mesh-test / espnow,cast,io,mesh-test.
- `cargo build --release` **links clean** (valid RISC-V ELF): default / espnow / espnow,cast,io / espnow,wled / espnow,mesh-test.
- Host `experiments/flood_verify`: ALL PASS (SeenSet incl. multi-frag regression + forward_decision + HopLatch hysteresis).
- Rebased onto current `main`; re-verified clippy + release link post-rebase.

Refs #13. Follow-up: #124.

---

## Update — C0 canary + hardening (post first-canary)

The first canary flagged `fwd≠0` on the fleet. Joint root cause (both team-lead and I were half-right, on the record): a fair-weather counter read taken during a 1 MB **OTA-convergence window** (crown legitimately mesh-deaf for long stretches) → the leaves' telemetry genuinely went un-ACKed → the escalation machinery responded per spec — but the latch **threshold treated one bad message as strandedness**. Not a wire bug, not an ACK-recognition bug (the whole H=1 RELAY/RELAYACK path is byte-identical to base; the executable guard below proves it).

Two hardening commits landed on top:

**C0 fix — escalation hysteresis** (`1de3184`)
- `HopLatch` now latches multi-hop only after `ESCALATE_STREAK=3` **consecutive** fully-un-ACKed messages (~45 s of genuine strandedness at the 15 s cadence); **any** ACK resets the streak. A single transient full-loss — normal over a 15-min window, and guaranteed during an OTA feed — no longer latches a leaf into RELAY2, so the `fwd=0` byte-identical invariant holds under ordinary packet loss. `flood_verify` gains the regression asserts.
- Plus a rate-limited serial log in the RELAY2 forward arm (was OLED-only).

**Codec extraction + byte-compat guard** (`1fbd367`)
- The relay-family wire codec moved to a pure, host-testable `net/wire.rs` (byte-identical move; the `espnow,cast,io` release ELF is unchanged in size). New `experiments/relay_compat` `#[path]`-includes the REAL codec and asserts the mixed-fleet contract **both directions** (new-frame→pre-#13-parser with golden bytes; old-RELAYACK→new-matcher) + RELAY2/RELAYACK2/BATT2 round-trips + tag disambiguation. The permanent guard the bench asked for, and the #124 UP2-migration checkpoint.

**Re-canary:** on an all-#13 fleet (DUT wears the crown), C0 on a quiet fleet with the 15-min counter watch, then the deaf-list rig P1–P4.

Full green: clippy `-D warnings` on default / espnow / espnow,cast,io / espnow,wled / espnow,mesh-test; release links clean; `flood_verify` + `relay_compat` ALL PASS.

### C0 verdict + DIAG counter semantics (`c382d76`)

C0 **PASSED** the uplink invariant on a quiet all-#13 fleet (15-min window): crown `fwd=0`/`dedup=0`, `hop=1` at rest, telemetry flowing, zero RELAY2/forward logs — the hysteresis held through normal loss.

`fwd` is the uplink invariant and increments at exactly one site (the RELAY2 forward arm) — it counts inbound uplink RELAY2 re-broadcasts only, and must be 0 fleet-wide in all-hear. Stage B's **downlink** BATT2/GRID2 re-flood is nonzero *by design* (strictly-newer flood, like TIME) and now has its own `dfwd=` counter — kept separate so the uplink invariant stays machine-checkable. Any nonzero `fwd` in all-hear is therefore always a real uplink signal (e.g. storm-tail residue from a pre-quiet convergence window), never a downlink artifact.

Note: the `DiagCounters` reset on boot (plain struct, not `noinit`); absolute counter values are same-boot accumulation, so score C0 off the **delta** in the observation window (which was 0).

### Hardware canary: C0 + deaf-list rig — BOTH PASS

The deaf-list rig caught (and this PR fixes) two escalation-trigger bugs the green build + the host HopLatch math could not — exactly the rig's purpose:
- **C0 over-eager latch** → fixed by the consecutive-unACK hysteresis (`1de3184`).
- **P1 under-eager latch** (a totally-deaf leaf never escalated — the retry-exhaustion path is starved for a leaf pinned in blocking recovery bursts) → fixed by feeding the streak at message **supersession** (`e0554a7`).

**P1 PASS — first routed frame in smol's history.** Latch fired at ~3 emit cycles as designed (`escalated to multi-hop (RELAY2) — gateway unreachable, 3 un-ACKed msgs`), steady RELAY2 at hop 2, R forwarded, and `smol/9/telemetry` arrived at the broker **from a leaf that provably cannot hear its crown**.

**Documented v1 limitation — multi-hop throughput (follow-up #126):** a stranded leaf never locks a channel (it never hears its owner's HELLO), so it keeps hopping ch 1/6/11 while scanning; its RELAY2 lands on the relay's channel only ~1/3 of the time (rig: R forwarded ~1 of ~30). The RELAYACK2 flood-back hits the same ~1/3 in reverse — the rig's F received zero ACKs and stayed latched on the escalation streak alone (correct for a genuinely-stranded leaf). **Pre-#13 the delivered count was ZERO, so v1's bar — a stranded leaf CAN reach home — is met.** #126 (latched-leaf channel parking) raises the rate.

### DIAG counter / uptime semantics (authoritative)

- `fwd` = **uplink** RELAY2 forwards only (the C0 invariant; must be 0 fleet-wide in all-hear). `dfwd` = **downlink** BATT2/GRID2 re-floods (nonzero by design). Split so the invariant stays machine-checkable.
- `DiagCounters` (incl. `fwd`/`dfwd`/`dedup`/`ttl`) are plain RAM — **reset to 0 on every boot** (not `noinit`). Score invariants off the **delta** in the observation window, not absolutes (which can hold same-boot storm residue).
- `boot` (`boot_count`) is NVS-persisted, +1 per boot — the **only** authoritative reboot marker. `up` is systimer-based and **survives `software_reset`** (continues across an OTA/soft reboot) — it is NOT a reboot indicator. A genuine reboot shows `boot`+1 AND `fwd`→0 together.

### DIFF
--- a/experiments/flood_verify/Cargo.toml
+++ b/experiments/flood_verify/Cargo.toml
@@ -1,0 +1,13 @@
+# #13 host-test harness for the PURE managed-flood core (net/flood.rs). Runs OFF-target
+# on the host (std) via `cargo run` — the firmware crate is no_std/riscv32imc and can't
+# host-test, so this #[path]-includes the real flood.rs (no drift) and exercises it.
+[package]
+name = "flood_verify"
+version = "0.0.0"
+edition = "2021"
+publish = false
+[[bin]]
+name = "flood_verify"
+path = "src/main.rs"
+[profile.dev]
+opt-level = 0
--- a/experiments/flood_verify/src/main.rs
+++ b/experiments/flood_verify/src/main.rs
@@ -1,0 +1,96 @@
+//! Host verification of the PURE #13 managed-flood core. Includes the real
+//! `net/flood.rs` verbatim (#[path], no drift) and exercises the seen-set, the
+//! forward decision, and the hop-latch hysteresis. Run: `cargo run` — panics on failure.
+
+#[path = "../../../rust/clock/src/net/flood.rs"]
+mod flood;
+
+use flood::{
+    forward_decision, ForwardAction, HopLatch, SeenSet, MAX_HOP, PROBE_EVERY, SEEN_RING,
+    UNLATCH_STREAK,
+};
+
+fn main() {
+    // --- SeenSet ----------------------------------------------------------
+    let mut s = SeenSet::new();
+    assert!(!s.contains(7, 100), "empty");
+    assert!(!s.seen_or_insert(7, 100), "first sight is new");
+    assert!(s.contains(7, 100), "recorded");
+    assert!(s.seen_or_insert(7, 100), "second sight is a dup");
+    assert!(!s.seen_or_insert(7, 101), "diff msgid is new");
+    assert!(!s.seen_or_insert(8, 100), "diff origin is new");
+    // idempotent insert doesn't consume extra ring slots
+    s.insert(7, 100);
+    s.insert(7, 100);
+    // drop-oldest overflow: fill past capacity, the earliest ages out.
+    let mut r = SeenSet::new();
+    for m in 0..(SEEN_RING as u16 + 5) {
+        assert!(!r.seen_or_insert(1, m), "each new msgid is new on insert");
+    }
+    assert!(!r.contains(1, 0), "oldest (msgid 0) aged out after overflow");
+    assert!(r.contains(1, SEEN_RING as u16 + 4), "newest retained");
+
+    // --- forward_decision -------------------------------------------------
+    // already-seen → DedupDrop regardless of role/hop.
+    assert_eq!(forward_decision(false, 2, true), ForwardAction::DedupDrop);
+    assert_eq!(forward_decision(true, 2, true), ForwardAction::DedupDrop);
+    // gateway (sink) → Reassemble, never forward.
+    assert_eq!(forward_decision(true, 2, false), ForwardAction::Reassemble);
+    assert_eq!(forward_decision(true, 1, false), ForwardAction::Reassemble);
+    // relay with hops left → Forward at hop-1.
+    assert_eq!(forward_decision(false, 2, false), ForwardAction::Forward { hop: 1 });
+    assert_eq!(forward_decision(false, 3, false), ForwardAction::Forward { hop: 2 });
+    // relay, hop exhausted (<=1) → TtlDrop (a RELAY2 that decremented to 1 at a
+    // non-gateway means it never reached the sink → drop, count ttl_drops).
+    assert_eq!(forward_decision(false, 1, false), ForwardAction::TtlDrop);
+    assert_eq!(forward_decision(false, 0, false), ForwardAction::TtlDrop);
+
+    // --- HopLatch: escalation ---------------------------------------------
+    let mut l = HopLatch::new();
+    assert!(!l.latched());
+    assert_eq!(l.origin_hop(false), 1, "single-hop by default (plain RELAY)");
+    // a partially-acked exhaust does NOT latch (gateway heard us, just lossy).
+    l.on_relay_exhausted(true);
+    assert!(!l.latched(), "partial ack ⇒ not stranded");
+    // a fully-unacked exhaust latches multi-hop.
+    l.on_relay_exhausted(false);
+    assert!(l.latched(), "zero ack ⇒ stranded ⇒ latch");
+    assert_eq!(l.origin_hop(false), MAX_HOP, "latched non-probe emits at MAX_HOP (RELAY2)");
+    assert_eq!(l.origin_hop(true), 1, "a probe always emits H=1 (plain RELAY)");
+
+    // --- HopLatch: probe gating (Gate A = downlink_up) --------------------
+    let mut l2 = HopLatch::new();
+    l2.on_relay_exhausted(false); // latch
+    // while downlink is DOWN, never probe (don't waste airtime — definitely stranded).
+    for _ in 0..(PROBE_EVERY * 3) {
+        assert!(!l2.should_probe(false), "no probe while downlink down");
+    }
+    // with downlink UP, probe fires on the 1-in-PROBE_EVERY tick.
+    let mut probes = 0;
+    for _ in 0..(PROBE_EVERY * 2) {
+        if l2.should_probe(true) {
+            probes += 1;
+        }
+    }
+    assert_eq!(probes, 2, "exactly 2 probes across 2×PROBE_EVERY ticks");
+
+    // --- HopLatch: un-latch hysteresis (no flap) --------------------------
+    let mut l3 = HopLatch::new();
+    l3.on_relay_exhausted(false);
+    assert!(l3.latched());
+    l3.on_direct_ack(); // streak 1 of UNLATCH_STREAK
+    assert!(l3.latched(), "one direct ack is not enough (hysteresis)");
+    l3.on_probe_miss(); // marginal RF: a miss resets the streak
+    assert!(l3.latched(), "still latched after a miss");
+    // now K consecutive good probes → unlatch.
+    for _ in 0..UNLATCH_STREAK {
+        l3.on_direct_ack();
+    }
+    assert!(!l3.latched(), "UNLATCH_STREAK consecutive direct acks drop the latch");
+    assert_eq!(l3.origin_hop(false), 1, "back to single-hop");
+    // a not-latched direct ack is a harmless no-op.
+    l3.on_direct_ack();
+    assert!(!l3.latched());
+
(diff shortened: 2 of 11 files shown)


jphein:
## Hardware canary — FULL SUITE PASSED, merging

**C0 (byte-identical invariant, quiet all-#13 fleet, 15-min watch):** crown fwd=0/dedup=0 throughout; leaf deltas 0 at steady state (the storm-tail un-latch bled into the first minutes and is understood + documented); telemetry flowing (118 events); executable proof via `relay_compat` golden bytes both directions.

**P1 (the flagship: stranded leaf routes home):** deaf-list rig G=id7(crown)↔F=id9 mutually deaf, R=id8 relaying. F latched at exactly 3 emit cycles (`leaf escalated to multi-hop (RELAY2) — gateway unreachable, 3 un-ACKed msgs`), emitted at hop 2, and at **11:03:19 its telemetry arrived at the broker through R — the first routed frame in smol's history.** Scan-hop throughput limitation documented in the body; follow-up #126 (channel parking) filed; #124 (UP2 migration) stands.

**The canary's catch record on this PR alone:** the convergence-storm single-message latch (fixed: K=3 hysteresis, 1de3184) and the total-deafness escalation starvation (fixed: supersession feed, e0554a7) — both structural trigger bugs invisible to build-green, both host-regression-tested now. Plus the counter-semantics clarifications (fwd/dfwd split, boot_count as the sole reboot marker) baked into the body.

Closes #13.

Issue #122, opened 2026-07-14
mesh: B1 flush-period 30→20s (power-soak-gated) + recovery-churn residual


jphein:
The two remaining #114 items, split out so the umbrella can close on its delivered scope:

1. **B1 — shorten the gateway flush period 30→20s.** The highest-leverage UI-latency lever (every HA-visible delay is quantized by it — decomposition on #117). Unblocked by #22's refute (no BLE airtime to co-budget) but gated on a dedicated power/airtime soak: mesh-deaf windows, leaf battery math (#34-future), election timing interactions (RECOVERY windows are tuned to the 30s flush — H3's fix explicitly assumes ≤30s MC republish, so B1 must re-derive MC_STALE_MS/RECOVERY_STALE_MS together).
2. **Recovery-churn residual** (flagged in #121): a leaf that re-adopts the SAME live owner never resets its silence clock (mode.rs ~2010) → perpetual ~10s recovery-burst churn. Airtime-only (it always adopts, never seizes) but wasteful; fix deserves its own canary.

Heal-correctness (H1 #115, H2 #115, H3 #121) and leaf-side UI freshness (#117) are shipped and hardware-verified.

jphein:
## B1 flush-shorten (30→20 s): election-window co-derivation

Derived the window dependencies **before** touching the flush interval, since H1/H2/H3 (#115/#121) tuned the recovery timings against the 30 s flush. Branch **`feat/122-b1-windows`** has the constants drafted and builds green (`cargo build --release --features espnow,cast,io`); **shipping stays soak-gated** — this comment is the merge-decision writeup.

### Constants
| Const | Now | Proposed @ F=20 | Scale with F? | Why |
|---|---|---|---|---|
| `RELAY_FLUSH_INTERVAL_MS` (F) | 30 s | **20 s** | — (the knob) | master clock behind every HA-visible latency (#117) |
| `MC_STALE_MS` | 90 s | **60 s** | **yes (3×F)** | keeps the 3-missed-flush margin; speeds the #121 never-heard heal ~90→~60 s |
| `RECOVERY_STALE_MS` | 35 s | **35 s (hold)** | **no** | gains margin at F=20 (see verdict) |
| `REELECT_SILENCE_MS` | 15 s | **15 s (hold)** | **no** | tied to the 2 s HELLO cadence, not F |
| `REELECT_RETRY_MS` | 10 s | **10 s (hold)** | **no** | convergence; must stay < RSSI bucket step |
[...]

PR #125, opened 2026-07-14
fix(fw): #114 — end recovery-burst churn on same-owner re-adoption

jphein:
## #114 — end recovery-burst churn on same-owner re-adoption

Tiny follow-up to #121 (H3). Closes **item 2 of #122** (the recovery-churn residual I flagged in the
#121 PR body). Branch off `origin/main@25d4003`.

### The churn
During a leaf's recovery re-election, when it **re-adopts the SAME live owner** it lost, the resolver
set `owner_alive = owner != lost_owner` — true only for a *different* owner. So for the same owner it
stayed **false**, and the caller's grace-reset of the owner-silence clock (`last_owner_heard_ms`,
mode.rs ~2010) was **skipped**. A leaf that hadn't yet locked the live owner's HELLO therefore
re-elected every `REELECT_RETRY_MS` (10s) **forever** — and since each recovery burst re-associates
**off** the ESP-NOW channel, it starved its own HELLO lock (`owner_hello_seen` never set), a
self-sustaining loop. After #121 this is **airtime-only** (it always adopts, never seizes a live
crown), but it's wasteful and observable in the recovery-burst cadence.

### The fix (one line + the comment around it)
Grace-reset the silence clock on **fresh liveness evidence** = `reset` (a *different* owner — a
successor to follow — **OR** the *same* owner whose **seq advanced** this burst, i.e. it published a
new MC since our last read → provably alive *now*):

```rust
elect.owner_alive = reset;   // was: owner != lost_owner
```

Each genuine seq advance pauses the redundant bursts, giving the leaf on-channel time to lock the
owner's HELLO — which then fully ends recovery. (`lost_owner` was that line's only use, so it's
removed; `reset` already encodes the owner-changed comparison.)

### Why there is NO liveness blind spot (supersedes the old "first-burst baseline" comment)
`owner_alive` / `last_owner_heard_ms` gates only **WHEN recovery bursts FIRE**, not the takeover
**ANCHOR**. The takeover window is measured from `seen_ms`, which the `reset` arm sets **independently**
of `owner_alive`. Consequently:

- A **genuinely-dead** owner (seq frozen ⇒ `reset=false` after at most one first-burst catch-up) still
  accrues to takeover on `seen_ms` and is **taken over on schedule** — unchanged from #121.
- A same-owner reset only ever pauses bursts by `< REELECT_SILENCE_MS` (15s), far inside the stale
  window (35s heard / 90s never-heard) → **zero net effect on failover timing**.
- A **live** owner (seq advancing) is correctly **never** taken over.
- The dead-but-inside-backoff path (the `else` arm) is untouched — `owner_alive=false`, deferred
  takeover still fires on cadence.

### Verification
- Full profile bar — `clippy -D warnings` clean on **default / wifi / espnow / cast / io / wled /
  espnow,cast,io**. (The `wifi`-only profile caught a dead-binding intermediate — fixed before commit.)
- **Bench canary (team-lead):** the churn is observable as the recovery-burst cadence on serial — a
  leaf following a live owner it hasn't HELLO-locked should **stop** re-electing every ~10s (bursts
  pause on each observed MC seq advance and cease once the HELLO locks), with **no** change to
  dead-owner takeover timing.

Refs #114 #122.

### DIFF
--- a/rust/clock/src/net/wifi.rs
+++ b/rust/clock/src/net/wifi.rs
@@ -2773,11 +2773,9 @@ fn mqtt_session(
     //      LIVE owner (sticky), and take over a DEAD owner only after this board's
     //      RSSI-weighted backoff, so the strongest-uplink survivor wins (node-id breaks
     //      ties). `channel` stays 0 (advisory; leaves discover it by scanning the HELLO).
-    // The owner the leaf was following (recovery seeds `seen_owner` from its last MC read).
-    // Captured before the observation below mutates it — lets us tell "the same owner we're
-    // waiting out" (defer, keep accruing) from "a different owner that just claimed" (a live
-    // successor to adopt + grace).
-    let lost_owner = elect.seen_owner;
+    // NB: the "same owner we're waiting out" vs "a live owner to grace" distinction is now made
+    // by `reset` (owner-changed OR seq-advanced this burst — see the alive branch + #114 churn
+    // residual note), so no separate pre-observation snapshot of `seen_owner` is needed here.
     let claim_seq: Option<u32> = match mc {
         Some((owner, _ch, seq)) => {
             // Refresh the staleness observation: a *changed* (owner,seq) resets the first-seen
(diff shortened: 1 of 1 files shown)


jphein:
## Hardware canary — PASSED, merging

Measured on id9 (instrumented, booted as a leaf following live crown id7 pre-HELLO-lock — the exact churn scenario): **2 recovery bursts in ~6 min** vs the pre-fix baseline of ~1/15s (~24 in the same window) — a ~10× reduction, matching the design (bursts pause on each observed MC seq advance). Dead-owner takeover timing untouched by construction: the one-liner only feeds `owner_alive` (burst-cadence clock); the `seen_ms` takeover anchor is set independently by the same reset arm — confirmed in the diff. Also credits: the wifi-only profile catching the orphaned binding is exactly why the full profile bar exists.

(Bench note for the record: this canary surfaced the **otadata/USB-flash trap** — after any OTA install, the board boots ota_1 and subsequent `espflash flash` writes to ota_0 never run. `espflash erase-region 0xf000 0x2000` resets otadata. Memory updated.)

PR #135, opened 2026-07-15
fix(ota): #134 don't burn a leaf install order on a pre-relay fetch failure (HW-gate held)

### LINKED ISSUE #134: OTA: leaf install order is consumed even when the relay fetch fails — a fetch-broken crown burns orders
## Summary

A gateway (crown) that **fails to FETCH the staged image** still **burns** the retained
`smol/<leaf>/ota/install` order after a few attempts. Fetching happens *before* any bytes are
relayed to the leaf, so a fetch failure is a **gateway-local, transient** condition that says
nothing about the leaf or the image — yet it is counted against the same retry cap that is meant to
retire a genuinely *doomed image*. The result: a crown that cannot reach the OTA host quietly
destroys a pending install that a healthy crown would have completed.

## Mechanism (consume-vs-clear mismatch)

The leaf-mesh-OTA relay flow is:

1. Gateway sees a retained `smol/<leaf>/ota/install = INSTALL`, arms a relay for that leaf.
2. Gateway **fetches** the staged image over WiFi into its inactive slot (from the OTA host).
3. Gateway **relays** the image to the leaf over ESP-NOW, then watches for the leaf to reappear at
   the new build.
4. The attempt's outcome is recorded (`record_leaf_ota`, `net/mode.rs`), which decides whether to
   **clear** the retained install (published empty/retained in `net/wifi.rs`) or **leave it to
   retry**.

`record_leaf_ota` classifies outcomes as terminal (clear) or transient (retry, bounded by
`LEAF_OTA_MAX_RETRIES = 3`). The transient bucket lumps together outcomes that are **categorically
different**:

- `FetchFailed` — *the gateway never fetched/staged the image* (WiFi/HTTP/verify). The feed **never
  handed off to the leaf**. Documented in `ota_mesh.rs` as *"could not FETCH/stage the image —
  never relayed."*
- `MacUnknown` — *the gateway never even learned the leaf's MAC*. Also never relayed.
- `RelayFailed` / `Timeout` — *the image WAS being pushed to the leaf* but the relay loop exhausted
  or the leaf never reappeared. This is the real *doomed-image* case the cap is meant for.

Because `FetchFailed` shares the 3-strike cap, **three fetch failures clear (burn) the retained
order** — even though nothing about the leaf or the image justified it. The `#111` crown-portable
success clear (version-flip) and the doomed-image backstop are both correct; the gap is that a
*pre-relay* failure is treated as if the image had reached the leaf and failed.

## Motivating incident (live, tonight)

A crown went **unicast-deaf** due to an AP-side FDB pathology: its TX worked, but *all* inbound
unicast — including ARP replies — was blackholed, so its TCP fetch from the OTA host could never
complete. Each flush it re-armed the pending install, fetch-failed, and after **three** such
fetch-fails it **burned the order**. It burned three separately re-armed orders this way before the
mesh finally dethroned the deaf crown. By then the retained install was gone, so the **healthy
successor crown had nothing left to relay** — the leaf silently never updated.

The pathology was transient and crown-local. Had the order survived, the successor crown would have
fetched and completed it on its first flush.

## Proposed fix

Keep the retained `smol/<leaf>/ota/install` until **the feed actually HANDS OFF to the leaf** (relay
bytes start flowing) **or** the `#111` version-flip success clear fires. Concretely:

- Split the transient outcomes by whether the image *reached the leaf*:
  - **Never reached the leaf** (`FetchFailed`, `MacUnknown`): **never** clear the order. These are
    gateway-local — the install must survive for the next attempt / the next crown (orders are
    already crown-portable per `#111`). Count them separately for observability only.
  - **Reached the leaf but didn't complete** (`RelayFailed`, `Timeout`): keep the existing bounded
    retry cap — the doomed-image backstop, so a bad image can't re-arm forever.
- Add a **retry counter** to the retained `smol/<leaf>/ota/diag` payload (e.g. `fetch-failed
  retry=N`) so a stuck fetch is visible headlessly instead of silently burning the order.

### Touch points
- `net/mode.rs` — `record_leaf_ota` (~L3181): the clear/retry classification + a separate
  fetch-retry counter.
- `ota_mesh.rs` — `LeafOtaOutcome` (~L349): a `reached_leaf()` helper (the never-relayed set).
- `net/wifi.rs` — the diag publish + clear gate (~L2615-2634): surface the retry counter in the
  retained `ota/diag` payload.

### Verification
Host build only for the draft (`cargo build --release --features espnow,cast,io`). The real proof is
a HW canary: force a fetch failure on a crown (block its route to the OTA host), confirm the order
**survives** and a healthy crown completes it; confirm a genuinely doomed image still retires via the
bounded relay-retry cap. No merge on green alone.

Refs: the `#13`/`#40`/`#111` leaf-mesh-OTA lineage.

jphein:
Closes #134.

## Problem

A gateway (crown) that fails to **fetch** the staged image still **burns** the retained
`smol/<leaf>/ota/install` order. Fetching happens *before* any bytes are relayed to the leaf, so a
fetch failure is **gateway-local and transient** — it says nothing about the leaf or the image — yet
it was counted against the same 3-strike `LEAF_OTA_MAX_RETRIES` cap meant to retire a genuinely
*doomed image*.

**Live incident:** a unicast-deaf crown (AP-side FDB pathology — TX worked, all inbound unicast incl.
ARP replies blackholed, so its TCP fetch from the OTA host never completed) fetch-failed 3× and
burned the order — three separately re-armed orders this way — before the mesh dethroned it. The
healthy successor crown then had nothing to relay, and the leaf silently never updated.

## Fix

`record_leaf_ota` (`net/mode.rs`) now applies a **three-way** clear/retry policy instead of two:

| Outcome class | Examples | Policy |
|---|---|---|
| terminal (leaf acted) | `Confirmed` / `RolledBack` / `IdMismatch` | **CLEAR**, reset counters |
| pre-relay transient (feed never handed off) | `FetchFailed` / `MacUnknown` | **NEVER clear** — order survives for the next attempt / the next crown (crown-portable per #111); counted only |
| post-handoff transient (image reached the leaf) | `RelayFailed` / `Timeout` / `Aborted` | bounded retry (`LEAF_OTA_MAX_RETRIES`) then CLEAR — the doomed-image backstop, unchanged |

- `ota_mesh.rs` — new `LeafOtaOutcome::reached_leaf()` = the never-relayed set `{FetchFailed,
  MacUnknown}`. This is the semantic boundary: "did the image feed actually hand off to the leaf?"
- `net/mode.rs` — the split above + a new `leaf_ota_fetch_retries` counter (observability only —
  it never burns the order).
- `net/wifi.rs` — the retained `smol/<leaf>/ota/diag` payload now carries the retry count
  (`fetch-failed retry=3`) so a stuck fetch is visible headlessly instead of a silent order-burn;
  `retry=0` keeps the bare phase (unchanged). The `leaf_ota_diag` tuple widened to
  `(id, phase, clear, retry_n)`.

## Verification (host + build only)

- `cargo build --release --features espnow,cast,io` — passes (valid RISC-V ELF).
- `clippy --features espnow,cast,io -- -D warnings` — clean.
- No flash, no MQTT publish, no mesh interaction (a live fleet cascade was in progress).

## ⛔ MERGE GATE: HW rig canary pending hardware

Do **not** merge on green builds alone (the #123/#130 lesson — on-air behaviour can diverge from what
green builds show). One canary board must prove, on air:

1. Force a crown fetch failure (block its route to the OTA host) → the retained
   `smol/<leaf>/ota/install` **survives**, and a healthy crown completes it on its next flush.
2. A genuinely doomed image still retires via the bounded relay-retry cap (unchanged path).
3. `smol/<leaf>/ota/diag` shows `fetch-failed retry=N` climbing (the headless-visibility win).

Canary ONE node first (OTA rollback discipline).

## Review question — per-tenure fetch throttle (trade-off)

With this fix a genuinely deaf crown re-fetches **every flush** (a blocking fetch each time) until it
recovers or is dethroned. The order is preserved (the point), but the crown keeps spending fetch
attempts. A refinement would be a per-tenure fetch throttle that stops *this crown* re-arming a leaf
after N fetch-fails **without clearing the order** (a RAM "abandoned-this-tenure" set, not a retained
clear) — leaving the order for the next crown. Not required for correctness (the mesh dethroning the
deaf crown is the real recovery), so left out of this PR. Want it in, or as a follow-up?

Refs #40, #111, #123.

### DIFF
--- a/rust/clock/src/net/mode.rs
+++ b/rust/clock/src/net/mode.rs
@@ -1804,12 +1804,14 @@ pub struct RadioManager {
     /// independent drain timing). Persists the current staged image for every leaf relay.
     staged_raw: Option<crate::ota::Announce>,
     /// #40 headless observability + clear/retry: the LAST leaf-OTA attempt's `(leaf_id,
-    /// phase, clear_install)`, set by `main` after `run_leaf_ota_relay`, PUBLISHED to
+    /// phase, clear_install, retry_n)`, set by `main` after `run_leaf_ota_relay`, PUBLISHED to
     /// `smol/<leaf>/ota/diag` on the gateway's next burst and used to drive the install
     /// clear (terminal/exhausted) vs retry (transient). `None` when nothing is pending.
     /// The phase is stored as the rendered `&'static str` (not the espnow-only enum) so the
-    /// shared `wifi::mqtt_session` signature stays buildable in the wifi-only profile.
-    leaf_ota_diag: Option<(u8, &'static str, bool)>,
+    /// shared `wifi::mqtt_session` signature stays buildable in the wifi-only profile. `retry_n`
+    /// (#134) is the current consecutive-failure count for this outcome class — surfaced in the
+    /// retained payload (`fetch-failed retry=3`) so a stuck fetch is visible headlessly.
+    leaf_ota_diag: Option<(u8, &'static str, bool, u8)>,
     /// #40 #1 DECOUPLE: true while a leaf-OTA relay is pending/in-flight (armed until its
     /// outcome is terminal or the retry cap is hit). While set, the gateway SUPPRESSES its own
     /// self-OTA (`do_install`) so a relay is never interrupted by the gateway rebooting into a
@@ -1837,9 +1839,15 @@ pub struct RadioManager {
     /// chunk-loss stall. `last_wb` is in chunk units (matches `total = om::total_chunks`).
     /// Carries the leaf's `LDBG` self-report too (captured during the relay) — see `RelayDiag`.
     leaf_relay_rx: Option<crate::net::wifi::RelayDiag>,
-    /// #40: consecutive non-terminal (transient) relay attempts for the current install —
-    /// caps the auto-retry so a persistently-failing leaf can't loop the mesh-deaf relay.
+    /// #40: consecutive POST-HANDOFF transient relay attempts (`RelayFailed`/`Timeout`/`Aborted`)
+    /// for the current install — caps the auto-retry so a persistently-failing (doomed) IMAGE can't
+    /// loop the mesh-deaf relay. Does NOT count pre-relay failures (see `leaf_ota_fetch_retries`).
     leaf_ota_retries: u8,
+    /// #134: consecutive PRE-RELAY failures (`FetchFailed`/`MacUnknown` — the feed never handed off
+    /// to the leaf) for the current install. Counted for the retained `ota/diag` retry number ONLY —
+    /// it NEVER burns the order (a gateway-local fetch failure says nothing about the leaf/image, so
+    /// the install survives for the next attempt / the next crown). Reset on a terminal outcome.
+    leaf_ota_fetch_retries: u8,
     /// #21 node-manager: the parsed default-screen command surfaced by a burst,
     /// pending `main`'s `take_config_offer` → apply. `None` when nothing is pending.
     config_offer: Option<crate::app::DefaultScreen>,
@@ -1951,6 +1959,7 @@ impl RadioManager {
             staged_raw: None,
             leaf_ota_diag: None,
             leaf_ota_retries: 0,
+            leaf_ota_fetch_retries: 0,
             leaf_ota_pending: false,
             leaf_installs_outstanding: false,
             leaf_ota_mac: None,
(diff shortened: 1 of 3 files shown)


jphein:
## Canary evidence — merged in `d273b0a`

#134 — a leaf install order must not be burned by a pre-relay fetch failure:

- A retained install order **survived 11 consecutive fetch failures**, with `fetch-failed retry=N` visible in the retained diag across the whole run — the order was **preserved, not consumed**, exactly as the fix intends. A genuinely deaf crown now re-fetches every flush instead of silently dropping the order.
- Known remaining gap: a **gate-rejected announce still burns the order** — filed as part of #147.

Merge: `d273b0a`. PR: #135.

Issue #139, opened 2026-07-15
WiFi: disable power-save at init + association, not only after is_connected()


jphein:
## Problem

The esp-wifi/IDF default power-save is **`WIFI_PS_MIN_MODEM`** (modem-sleep — the station sleeps between DTIMs; the AP buffers *unicast* for the station and delivers at DTIM). We do assert `PowerSaveMode::None`, but **only after `is_connected()`** — at `rust/clock/src/net/wifi.rs:3330` (MQTT flush) and `:3663` (OTA fetch). The in-code FINDING N3b comment confirms **`connect()` resets the IDF ps state to the min-modem default each burst**.

So the **association/auth handshake window** — and any associate path that stalls before the re-assert — runs under min-modem, where unicast is DTIM-buffered and lossy on a marginal link. This is the verified mechanism behind tonight's observed signature:
- board **re-ARPs its gateway ~1 s after a reply was provably sent**
- **forwarded NTP replies never arrive for hours** (`tage` grows unbounded)
- one board went fully **inbound-unicast-deaf (TX fine)** until it self-healed

The -78 dBm worst board compounds it (Espressif ESP-FAQ: low RSSI → the STA may miss beacons).

## Fix

Assert `set_power_saving(PowerSaveMode::None)` **once immediately after `esp_wifi::wifi::new()` / in the `associate()` config** (`wifi.rs` ~377 / ~433), *in addition to* the existing post-connect re-asserts (belt-and-suspenders) — so no association window is ever left under min-modem. Accept the higher idle draw (already accepted per the N3b comment).

## Acceptance

- On a weak-RSSI board across a multi-hour soak: DIAG `tage` stops growing unbounded, and the re-ARP storm on a just-sent unicast reply disappears.

## Scope note (do NOT conflate)

This does **not** fix the ~20 s-flap-associate → AP forwarding-table (FDB) pollution that produced tonight's unicast-deaf *crown* (see `scratch/smol-ha-batt/incident-2026-07-14-unicast-deaf-crown.md`). That is an AP-side effect of the burst model and needs the burst-retire / flap-tolerant-SSID work (ROADMAP §4a) — a separate, larger issue.

## Notes

[...]

Issue #138, opened 2026-07-15
OTA fetch: chunked HTTP Range download to fix mid-body stalls


jphein:
## Problem

`run_ota_fetch` (`rust/clock/src/net/wifi.rs`) streams a ~1 MB image in a **single** minutes-long HTTP/1.0 GET through a 4 KB smoltcp rx window. On the current fleet that transfer **consistently dies mid-body ~40–90 s in** — the server sees the client vanish (BrokenPipe), the board retries, and it never completes. **Tonight the fleet had to converge via USB flash because this mid-body death blocked every OTA path** — so this is the priority reliability fix.

Contrast: seconds-long sessions on the *same association* (MQTT flush, DHCP) are rock-solid. The failure is specific to the minutes-long single GET.

This matches a known, largely-untunable class of esp-wifi/IDF sustained-transfer failures:
- esp-rs/esp-wifi #215 · esp-rs/esp-hal #1605 (sustained download throughput is buffer/round-trip-bound)
- espressif/esp-idf #12091 — HTTPS download with tens-of-seconds waits that randomly interrupt flow; **closed "Not Applicable"** even after heavy buffer tuning
- espressif/esp-hosted-mcu #184 — inbound TCP stall when internal RX buffers fill; connection stays open, no more data, times out

## Fix (structural — this is IDF's own answer)

IDF's `esp_https_ota` ships `partial_http_download` + `max_http_request_size` precisely because single long OTA downloads are unreliable. Adopt the same pattern:

- Replace the one `GET <path>` with a loop issuing `GET <path>` + `Range: bytes=<off>-<off+CHUNK-1>` per chunk (start **CHUNK = 32–64 KB**).
- Stream each response body into the inactive slot and feed the **existing running SHA-256** — the hash composes across chunks, so verify logic is unchanged.
- On a chunk failure, **retry that range only** (resumable), not the whole image. Keep the existing long-press abort + overall budget.

## Server precondition

The image server (`tools/`) must honour **`Range`** and return **`Content-Length`** (not chunked transfer-encoding) for ranged requests. Smoke test: `curl -r 0-1023 <image-url>` returns 206 + the first 1 KB.

## Acceptance

- A full ~1 MB image lands over ≥1 forced-lossy run with **no whole-image restart**.
- Per-chunk retry observed to recover a dropped chunk.

## Notes

[...]

PR #143, opened 2026-07-15
feat(fw): #138 chunked HTTP Range OTA fetch + #139 early PowerSaveMode::None

jphein:
⛔ **MERGE GATE: HW canary pending hardware.** Both changes are release + full-profile-bar green and reasoned from nebula's WiFi research, but NEITHER is validated on a live board — the fleet is converged/settled and was NOT touched. Do not merge until the (a)-(d) canary checks below pass.

The OTA transfer-reliability pair — they compose (both on `net/wifi.rs`'s association + fetch path). From nebula's findings 1b + 2 (`scratch/smol-ha-batt/nebula-c3-wifi-research.md`).

## #139 — PowerSaveMode::None at init + before association (`213297c`)
esp-wifi/IDF default is `WIFI_PS_MIN_MODEM`; `connect()` resets ps to it; we asserted `None` only AFTER `is_connected()` (@3330 MQTT flush, @3663 OTA) — so every association/auth handshake ran under modem-sleep (AP buffers UNICAST at DTIM → the re-ARP-after-reply / forwarded-NTP-never-arrives / TX-fine-RX-deaf class). Now asserted in the **shared espnow path**: at radio init (baseline, after `esp_wifi::wifi::new`) and in `switch()`'s WifiSta arm right after `connect()` (before the handshake completes, since connect() just reset it). Post-`is_connected()` re-asserts kept as belt-and-suspenders. Covers every espnow association (NTP / MQTT / OTA / re-election), not only `run_ota_fetch`.

## #138 — chunked HTTP Range OTA fetch (`5b90cb2`)
`run_ota_fetch`'s single minutes-long GET died mid-body 40–90 s in — untunable esp-wifi driver fragility (esp-idf #12091 closed "Not Applicable"; esp-wifi #215, esp-hosted #184), while seconds-long MQTT/DHCP on the same association are rock-solid. Now **sequential 48 KB Range chunks** — each a short `GET`+`Range: bytes=<off>-<end>` on a FRESH `Connection: close` socket, inside the window that demonstrably works. The `ImageWriter` accumulates across chunks → the running SHA-256 and the size + ed25519 verify gate are **UNCHANGED**.
- **Resume** from `writer.written()` on a broken chunk (no whole-image restart — resume falls out of the accumulator: flash offset + SHA position advance together).
- **206** happy path; a **200 on chunk 0** (server ignored Range) falls back to the old full-body single-GET semantics (`Content-Length==size` checked; not resumable → an incomplete 200 fails cleanly rather than double-writing).
- `OTA_FETCH_BUDGET` stays the overall cap; `OTA_MAX_STALL` (6 zero-progress attempts) guards livelock; `chunk_retries` → serial (the fetch reboots on success, so a persisted diag counter wouldn't survive — serial `ESP_LOG=info` is the rig's sink).
- Server precondition: Range + Content-Length (python http.server 3.7+ ✓; the 200-fallback covers servers that don't).

## Verification (host — the HW rig is the remaining gate)
Full profile bar GREEN: `clippy -D warnings` on default / espnow / espnow,cast,io / espnow,wled / espnow,mesh-test + release build (`espnow,cast,io`) links. NO flash / MQTT / mesh touched (fleet converged/settled).

## ⛔ HW canary checks (must pass before merge)
- **(a)** a full ~1 MB image fetch COMPLETES over N Range chunks on a live board where the single-GET died mid-body.
- **(b)** a mid-chunk connection kill RESUMES from bytes-written (serial "chunk retries" climbs) instead of restarting the whole image.
- **(c)** SHA-256 + ed25519 verify still GATES activation (a corrupted/short image is rejected, good slot intact).
- **(d)** #139: the association-window ARP/DHCP/unicast-loss signature is GONE — DIAG `tage` stops growing fleet-wide over a multi-hour soak.

Refs #138 #139.

### DIFF
--- a/rust/clock/src/net/mode.rs
+++ b/rust/clock/src/net/mode.rs
@@ -1892,6 +1892,14 @@ impl RadioManager {
         let (mut controller, interfaces) = esp_wifi::wifi::new(ctrl, p.wifi).ok()?;
         controller.set_mode(WifiMode::Sta).ok()?;
         controller.start().ok()?;
+        // #139: assert PowerSaveMode::None at RADIO INIT — the esp-wifi/IDF default is
+        // WIFI_PS_MIN_MODEM (the STA sleeps between DTIMs; the AP then buffers UNICAST at DTIM and
+        // drops it on marginal links → the re-ARP-after-reply / forwarded-NTP-never-arrives /
+        // TX-fine-RX-deaf class). This is the baseline; `switch()` re-asserts after each connect()
+        // (which resets ps to the default), and the post-is_connected() sites stay belt-and-
+        // suspenders. Shared by every espnow association (NTP/MQTT/OTA/re-election). See nebula
+        // finding 2 (scratch/smol-ha-batt/nebula-c3-wifi-research.md).
+        let _ = controller.set_power_saving(esp_wifi::config::PowerSaveMode::None);
 
         // #68/#76 SELF-MAC: capture our STA MAC (ESP-NOW rides the STA interface) so service()
         // can DROP frames from our own address. The esp-wifi RX queue can deliver our own
@@ -2268,6 +2276,13 @@ impl RadioManager {
                 // to the AP's channel automatically; ESP-NOW then coexists on
                 // that channel. (Caller must have valid credentials set.)
                 let _ = self.controller.connect();
+                // #139: connect() resets power-save to the WIFI_PS_MIN_MODEM default, so the
+                // association/auth handshake that follows would otherwise run under modem-sleep
+                // (unicast-deaf window). Re-assert None NOW — before the handshake completes — not
+                // only after the downstream is_connected() gates (MQTT flush / OTA fetch).
+                let _ = self
+                    .controller
+                    .set_power_saving(esp_wifi::config::PowerSaveMode::None);
                 log::info!("smol: radio -> WiFi STA (coexist)");
             }
         }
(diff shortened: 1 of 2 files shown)


jphein:
**Amendment pushed — `246d0e0` (resolves the serial-only observability gap).**

Flag 1 from the original PR (chunk_retries were serial-only, invisible on release/fleet images) is fixed: a **failed self-fetch now publishes a retained record to `smol/<id>/ota/diag`** via the leaf_diag / #135-armdiag machinery —

`self-fetch-failed chunk=K/N retry=R stall=S`

- `chunk=0/0` ⇒ died before the download (assoc / DHCP / slot-open)
- `K=N` ⇒ the download completed but the SHA-256/ed25519 verify rejected it
- `retry`/`stall` ⇒ transfer trouble during the download (the mid-body-death class)

A **successful** self-fetch needs no record (it reboots + flips the version — that IS the signal, and the server's GET log shows the N range requests externally). The relay path is unchanged (reports via `LeafOtaOutcome`). Consumed once (not re-emitted every flush).

Full profile bar re-run green (default / espnow / espnow,cast,io / espnow,wled / espnow,mesh-test clippy -D warnings + release link). This is the observability that would have made tonight's mid-body deaths a 10-minute diagnosis instead of three hours. HW-gate (a)-(d) unchanged.

jphein:
## Canary evidence — merged in `d273b0a`

#138 chunked HTTP Range OTA fetch + #139 early `PowerSaveMode::None`:

- The **200-fallback path completed fetch + verify + feed end-to-end, twice** — this is the **production OTA path**.
- The **chunked 206 (Range) path failed after chunk 1 on-air** — the same mid-body esp-wifi fragility the chunking was meant to sidestep reappeared at the chunk boundary. Filed as #147; the 200 fallback is what we ship on.

Merge: `d273b0a`. PR: #143.

PR #187, opened 2026-07-19
feat: #159 meshscope — live screen, OTA-in-progress, NTP-sync per node

jphein:
Three per-node signals JP wanted visible while watching the live fleet — all **read-only**
from topics the model already ingests (no firmware changes, no HW gate).

meshscope showing the three signals

## The three signals

1. **Live screen ("the familiar")** — `Node::screen()`: the AppKind from the STAT topic
   (`<AppKind>:<page>`), falling back to the DIAG `cfg=` default-screen echo. Shown under
   each node in the graph + at the top of the inspector. *(id7 above shows `Familiar`.)*
2. **OTA in progress** — a pulsing ring + `OTA →vLATEST` on the node; `updating v45→v48`
   + the `ota/diag` phase in the inspector. *(id8 above, mid-update.)*
   - ⚠️ **There is no live transfer % on the wire.** `ota/state` carries only
     `in_progress` + installed/latest; the `done/total` that drives the on-board OLED
     (#178) is never published to MQTT. So this is an **honest indeterminate** indicator
     + the version-flip on completion — not a fabricated bar. Filed #<followup> to publish
     `OtaProgress` if a live % is wanted.
3. **NTP sync** — `Node::sync_freshness()` from DIAG `tage`/`tsrc`: a green/amber/red/grey
   dot on each node, a coloured `time sync: ntp · 30s ago (fresh)` line in the inspector,
   and a fleet **`N NTP-stale`** count in the top bar. *(id9 above is stale at ~70 min.)*

## Shared derivation (one model, two faces)

All three live in `mesh-model` so both frontends derive them identically (the HA-parity
single-source rule):
- consts `SYNC_FRESH_S` (300) / `SYNC_STALE_S` (3600), a `SyncFreshness` enum
- `Node::{screen, time_age, time_src, sync_freshness}` + an `ota_phase` field
- **observatory** gets the live screen under its orbs for parity (it already renders OTA
  as crown→leaf particle streams).

## Also fixed (pre-existing, surfaced while rendering a debug build)

- `panel.rs`: a `redundant_closure` + `manual_filter` clippy lint.
- `graph.rs`: a `gamma_multiply(1.4)` that **panics** ecolor's debug assert (`factor>1`) —
  latent only because the live instance runs a release build. meshscope now runs clean in
  debug too.

## Verified on familiar
- `cargo build` (workspace) ✅ · `cargo clippy --workspace` ✅ **clean**
[...]

Issue #160, opened 2026-07-19
feat: custom-message platform (umbrella) — toast + fw TTL/attention/priority residual (persistent-screen herald shipped)


jphein:
## What (JP directive 2026-07-18: "promote that")
The #45 custom-screen plugin + HA integration graduates from a config channel to a product feature: **any HA automation can message any board's glass**.

## Proven this week
The rig used custom messages as a human-signaling channel ("USB ME / plug me in") with LED + screen-switch choreography — it worked flawlessly, repeatedly. That pattern becomes the feature.

### HA side (ha project)
- **`notify.smol_<name>` + `notify.smol_all`**: native HA notify platform via MQTT publish to the existing per-node topics — so EVERY automation gains board-glass delivery: doorbell, laundry, calendar, "dinner's ready."
- A dashboard "send message" card: target picker (board/all), text, duration, style.
- Message log sensor (last message per board, timestamp).

### Firmware side (small, HW-gated)
1. **TTL / auto-revert**: extend the compose wire with an optional `~<seconds>` suffix — the board shows the message then reverts to its prior screen automatically (the rig beacons needed manual clearing; notifications must be self-cleaning).
2. **Attention pattern**: on new-message arrival: LED triple-blink + (when #61's piezo lands) a chirp. A message that arrives silently on a shelf clock is a message missed.
3. **Priority flag**: `!` prefix = interrupt whatever screen is active (today custom only shows if default_screen points at it); normal messages queue politely behind an active game.
4. Headless boards (id5): LED-pattern-only delivery acknowledged in the design (their "screen" is the LED).

## Parity rule (standing directive)
Ships simultaneously in: HA (notify + card), the visualizers' event ticker (#158 mesh-model gains a message-event record), the site manual, and the Hackaday draft (it's a great section: "text your clocks").

## Gate
Firmware bits HW-gated as usual (TTL revert + attention pattern on a canary board); HA side normal review. The wire-format extension must stay backward-compatible (bare messages behave exactly as today).

jphein:
**HA side: DONE + deployed + HW-verified** (luna-notify). script.smol_notify / smol_notify_all / smol_notify_clear + sensor.smol_<id>_last_message + a 'herald' dashboard card, all live on the HA VM (ha/packages/smol_notify.yaml). HA-side timed auto-revert + {entity} resolution + word-wrap work today. Caught+fixed a real two-writer collision (the #45 heartbeat republish was wiping notifications → per-node notify-active skip flag).

[...]

Issue #158, opened 2026-07-19
tool: meshscope — realtime egui mesh instrument (MQTT → live graph)


jphein:
Native Rust (eframe/egui) realtime visualizer for the live mesh: rumqttc subscribes smol/# → model thread folds DIAG/STAT/PEERS/mesh-channel/ota/* into world-state → immediate-mode canvas: nodes as discs (name, build, heap sparkline), RSSI-weighted edges from PEERS records, crown badge + channel, OTA progress bars, scrolling event ticker (elections, fetch retries, toasts). Crate at rust/meshscope (own workspace — NOT the embedded workspace). Creds via env (SMOL_MQTT_HOST/USER/PASS), never committed. Also compiles to WASM later for the site (pairs with #152). No firmware changes — pure listener. No HW gate; normal review.

Issue #142, opened 2026-07-15
net: remove the secondary-network fallback — single-network operation


jphein:
## What
Remove the #100 dual-slot network machinery (slot 1 + assoc-fallback revert). Boards operate on the single primary iot SSID only. Owner's call 2026-07-14 after the fallback's behavior was characterized live.

## Why
- The secondary (roam) SSID is 802.11r/FT-enabled and has been implicated in AP-table pathologies for burst-mode clients (see the 2026-07-14 incident notes) — a board \"saved\" by the fallback lands on a network we don't want it on.
- Tonight's live case: a board whose primary association fails at its placement spot silently became a secondary-network resident. Desired behavior: keep retrying the primary; a board that can't associate is a mesh-only leaf (ESP-NOW membership, relayed telemetry, mesh time) — fully functional, no WiFi bursts.

## Scope
- `secrets.rs` / `secrets.rs.example`: `WIFI_NETWORKS[2]` → single network (keep `ota_hosts` allowlist per the one network).
- NVS `NetCfg`: drop `active`/`commanded`/`fallback` slot fields; KEEP the broker + OTA-host override fields (#100 Stage 2/3, #116 semantics) — the override mechanism is independent and was load-bearing this week.
- `net/wifi.rs`: remove the assoc-fallback revert (the one revert-write) + slot-selection; association failure = retry primary next burst, never switch.
- CFG `N` key (net-slot command): remove (or repurpose as no-op reject); update `docs/protocol.md`.
- DIAG `net=` field: remove or freeze to a constant; NOTE: the HA dashboard's discovery sensors parse DIAG — coordinate the field change with a dashboard follow-up.
- Election/gateway paths: no changes expected (broker selection becomes single-legged).

## Gate
HW-canary before merge (one board): primary-assoc-fail board stays mesh-only and keeps retrying (no crash loop, no fallback), normal boards unaffected across reboot + OTA + election cycles.

PR #196, opened 2026-07-19
feat(meshscope): #23 operator mode — OTA + HA command controls behind --operator

### LINKED ISSUE #23: feat: retire the WiFi-burst teardown — run WiFi+ESP-NOW co-channel coexist (kills the mesh-deaf window)
**Highest-leverage roadmap item (Oracle, 2026-07-08). Biggest win, cheapest on the firmware — zero new esp-wifi deps.**

Today the gateway goes mesh-deaf during every WiFi burst because it **deliberately** tears the radio down to ch6 (`Mode::EspNow` disconnect + re-pin). But WiFi+ESP-NOW can run **concurrently** — one stack, one channel, no `coex` blob (that's WiFi↔BT only). smol already builds wifi+esp-now and already has the coexist arm (`mode.rs` keeps the association). So retiring the burst is a firmware simplification, not a new dependency.

## The catch (the real cost)
The mesh is hardcoded ch6 (`ESP_NOW_FIXED_CHANNEL=6`) so credential-less leaves can find it — but coexist rides **the AP's** channel, and `esp_now` can't pin its channel while associated. So a gateway stays co-channel with the ch6 leaves **only if the `roam` AP is on channel 6**. → **Requirement: pin the roam 2.4 GHz AP(s) to ch6 + disable auto-channel** (network config, not firmware). AP elsewhere/hopping ⇒ associated gateways orphan the leaves ⇒ mesh breaks.

## Payoff
Kills the radio teardown + multi-second deaf window + the 30 s boot assoc-timeout freeze **at the root** (always associated); flush collapses from associate→DHCP→MQTT (~seconds) to a **sub-second** round-trip on a live link. Makes #20's spinner/abort/defer *largely* moot — but not 100%: the blocking MQTT socket poll still yields a ~sub-second stall unless made cooperative. #20 (shipped, 0ce1ce9) stays valid as the safety net + residual-stall fix.

## Gating checks (both HW/network, not desk-assertable)
1. Can `roam` be pinned to ch6 without collateral (channel-plan / interference)? — network recon in progress.
2. HW soak test: COEXIST ESP-NOW RX reliability + latency under a live association — the unknown `protocol.md:524` itself flagged "unmeasured". Power-save already None, so no DTIM deafness.

Relates: #20 (works around this), #3 (relay/burst), #9 (relay tunables). Ranked above #22 (BLE).

jphein:
Closes #23. Un-defers the visualizers spec's parked `--operator` mode. **Design was reviewed + approved by team-lead** (scratch/meshscope-operator/morpheus-158.md) before any publish code was written.

## The invariant
`--operator` gates **all** publishing. Without it, meshscope constructs **no publish client** and is a byte-identical pure listener (spec §3). With it, an Operator dock appears and meshscope may publish to smol command topics — the sanctioned single exception to listener-only.

## Architecture
Publish path is **meshscope-local**: a separate `rumqttc` client with a distinct `<id>-op` client id, so it coexists with the listener (no same-id reconnect war — the bug fixed in #158). The shared `mesh-model` crate (used by observatory too) is **untouched**.

## Safety by construction (the brick-critical part)
- **Retain is fixed per command by its typed builder.** `cmd/reset` + `cmd/scan` are built via the transient path (`retain=false`) — there is **no builder** that makes them retained. A retained `cmd/reset` = permanent ~10 s reboot-loop soft-brick, so this is enforced structurally + a unit test (`transient_commands_never_retained`).
- **Confirmation modal** (shows the exact topic + payload + retain) for reboot / broker / ota_host / units / channel_hint — the operator's last check.

## Control surface (verified against `net/wifi.rs`)
`ota/install` ("Install" verb + staged-vs-installed skew like HA's Update entity) · `config/{default_screen,led,plugins,custom,io}` + `io/set` · `config/{broker,ota_host}` · `cmd/{reset,scan}` · fleet `config/units` + `mesh/channel_hint`.

**Corrections found while wiring** (the confirm-before-wiring gate earned its keep):
- `config/net` **omitted** — retired/inert per #142 (fw drains+ignores `N`).
- #160 notify **omitted** — no MQTT command topic exists yet.
- `plugins` is **4-ASCII-hex → u16** (not decimal); `units` token is the compact `F24`/`C12` form.

## HA parity (#25)
Labels/thresholds mirror the #24 dashboard; NTP freshness uses the canonical mesh-model `SYNC_FRESH_S=300`/`SYNC_STALE_S=3600` (#187). Exact command-label vocab being reconciled with luna-notify (#25).

## Verification
Built on **familiar** (viz is std/non-pinned): `clippy -D warnings` + `--release` green; **4 operator unit tests pass** (retain-safety, destructive flags, topic formatting, channel-hint clear). Screenshotted **live against the fleet** (mid-v341-roll) — no clicks, nothing published. Creds env-only; broker/host field hints are RFC5737 placeholders (no real infra in-repo). Binary + screenshot in git-ignored `bin/`.

### DIFF
--- a/rust/viz/Cargo.lock
+++ b/rust/viz/Cargo.lock
@@ -2597,6 +2597,7 @@ dependencies = [
  "egui",
  "egui_plot",
  "mesh-model",
+ "rumqttc",
 ]
 
 [[package]]
--- a/rust/viz/meshscope/Cargo.toml
+++ b/rust/viz/meshscope/Cargo.toml
@@ -19,5 +19,8 @@ mesh-model.workspace = true
 eframe = { version = "0.28", default-features = false, features = ["default_fonts", "glow", "wayland", "x11"] }
 egui = "0.28"
 egui_plot = "0.28" # sparklines in the detail panel
+# #23 operator mode: the publish path (a separate client, distinct id). The listener
+# lives in mesh-model; this is meshscope's OWN client for --operator publishing.
+rumqttc.workspace = true
 # Load a local .env (SMOL_MQTT_HOST/USER/PASS) if present — never committed.
 dotenvy.workspace = true
--- a/rust/viz/meshscope/src/main.rs
+++ b/rust/viz/meshscope/src/main.rs
@@ -10,6 +10,7 @@
 //!   meshscope --selftest headless: feed sample payloads through the model, print, exit
 //!   meshscope --help / --version
 
+mod operator;
 mod ui;
 
 use std::sync::{Arc, Mutex};
@@ -27,6 +28,9 @@ USAGE:
     meshscope [FLAGS]
 
 FLAGS:
+    --operator    Enable OPERATOR mode: an Operator dock appears and meshscope may PUBLISH
+                  to smol command topics (OTA install, config/*, reboot/scan, channel_hint).
+                  WITHOUT this flag meshscope is a pure listener (default). Live mode only.
     --demo        Launch the UI seeded with synthetic mesh data (no broker needed)
     --selftest    Headless: feed sample payloads through the model, print a summary, exit
     --help        Show this help
@@ -56,13 +60,17 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     }
 
     let demo = args.iter().any(|a| a == "--demo");
+    let operator_on = args.iter().any(|a| a == "--operator");
     let start = Instant::now();
 
-    let model = if demo {
+    let (model, operator): (Arc<Mutex<Model>>, Option<operator::Publisher>) = if demo {
         let mut m = Model::new("(demo — no broker)".into());
         seed_demo(&mut m);
         m.conn = ConnState::Connected;
-        Arc::new(Mutex::new(m))
+        if operator_on {
+            eprintln!("meshscope: --operator has no effect with --demo (no live broker to publish to)");
+        }
+        (Arc::new(Mutex::new(m)), None)
     } else {
         let cfg = match mqtt::BrokerCfg::from_env().map(|c| c.with_client_id("meshscope")) {
             Ok(c) => c,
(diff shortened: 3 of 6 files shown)


jphein:
**#25 parity checklist** (HA dashboard ↔ meshscope operator mode): `~/Projects/ha/docs/parity-smol-controls.md` (ha repo, commit ff5783b). Row-by-row control/signal parity with verdicts.

**⚠️ One wire BUG to fix before merge — units token:**
`Units::from_wire` (`rust/clock/src/units.rs:47-63`) does `s.split('|')` and matches the first token against exactly `"F"`/`"C"`, second against `"24"`/`"12"`. The GUI's `F24` → first token `"F24"` → no match → returns `None` → **board silently keeps current units** (#46 clamp). Canonical wire = `<F|C>|<24|12>` (e.g. `F|24`), which is what HA publishes (units.rs module doc: "mirrors the HA payload"). Fix: emit `f"{t}|{c}"`.

Everything else on the operator-control surface verifies ✅ (led, plugins wire-identical, transient reboot, OTA Install, net-omitted-both-sides). Additive gaps + AppKind list-lag (`app.rs` has Finder/Hunt/Watch/WledRemote) tracked in the checklist + smol#194. Herald compose/word-wrap logic sent for the composed custom-message control follow-up.

jphein:
### #25 parity review complete

luna-notify's committed checklist (`ha/docs/parity-smol-controls.md`, ff5783b) cross-links here. All verdicts green on the GUI side after two fixes from the review:

- **dc5c5c8** — default-screen AppKind list aligned to app.rs canonical (added `Snake` + `Custom`, which were missing).
- **9192007** — **units wire token must be pipe-separated** (`F|24`, not `F24`): the fw `units::from_wire` splits `config/units` on `|` and matches the halves exactly, so a joined token parsed to `None` and boards silently kept their units. Regression test locks the pipe form.

Confirmed clean both directions: `config/net` omitted (retired #142), reboot transient (retain=false, enforced + tested), OTA Install verb + skew.

One flagged parity gap → follow-up **#197**: the operator Custom control is raw-wire entry; a composed herald control (target/message/size/align/duration/priority) matching HA's #160 composer is scoped there with luna's verified wrap logic.

Issue #204, opened 2026-07-19
fw: crown-role coexist unicast-RX starvation — crown goes downstream-deaf (2 boards, 3 incidents); no remote recovery lever exists


jphein:
## Summary

The mesh **crown** (the coexist gateway: ESP-NOW on ch6 **and** a WiFi STA association) loses reliable **WiFi downstream unicast reception within minutes of taking the crown**. The board's transmit path and broadcast reception keep working, but sustained inbound unicast is dropped/heavily intermittent — so the crown can't complete anything that depends on *receiving* unicast: **OTA image fetch, SNTP time sync, and MQTT downlink** all fail while it holds the crown.

**The disease follows the ROLE, not the board.** Observed across **2 boards / 3 incidents**; leaves stay healthy, crowns go deaf.

## Evidence

- **id9 as crown:** OTA fetch stalls — TCP handshake + Range `GET` succeed (server logs `206`), then the crown **ACKs zero downstream data**; server retransmits ~28s → board RSTs → retries every ~40-90s. From the board's own gateway: ping ~100% loss; ARP won't stay resolved (~1s-spaced unicast `who-has`, no reply; entry churns FAILED↔PROBE). Crown `DIAG` shows `tsrc=mesh` (boot SNTP reply never arrives). Meanwhile it keeps *publishing* telemetry fine (upstream unicast OK) and holds a DHCP lease (broadcast OK).
- **Reset does NOT cure it (verified):** a spray of transient `cmd/reset` rebooted the crown (boot counter bumped), but it **re-claimed the crown and re-deafened** — still ping-fail, still `tsrc=mesh`. So it is not board-side stale state.
- **Role hand-off reproduces it:** the mesh dethroned the deaf crown on its own; the next board to crown attempted its self-fetch, got one `206`, then within minutes went the same way (neigh STALE, ping 100% loss) — **despite having received the prior image cleanly as a leaf hours earlier.**
- **Prior precedent:** the same crown-deafness was seen before (a crown went inbound-unicast-dead, MQTT died, it got dethroned, and it recovered only **after losing the crown**).
- **Ruled out:** OTA server/image (a normal LAN client pulls the full image at line rate); router/firewall (a dual-interface capture on the inter-VLAN router shows it forwards **every** downstream data frame toward the AP — the drop is the **last hop: AP wireless unicast delivery to the coexisting STA**); power-save code (already `PowerSaveMode::None` during fetch); MQTT payload/topic.

## Mechanism (hypothesis)

[...]

Issue #190, opened 2026-07-19
mesh: transport auth — group HMAC-SHA256 rung (B1, reshaped per Fork B / #36) [#190]


jphein:
## Origin (why this issue was reshaped)
Started as the #189 mesh-auth **GAP** (Althea #12 per-link-auth principle, learned-and-untracked). The **#36 spike** then proved ESP-NOW native encryption **cannot touch broadcast** (unicast-LMK only) — so the original "PMK/LMK encryption" framing is not buildable for smol's broadcast flood. This issue is now the **implementation-sized spec of the reshaped transport rung (Fork B / B1)** — an app-layer **group HMAC-SHA256** (authentication, not encryption). Design of record: `docs/superpowers/specs/2026-07-19-mesh-auth-design.md` §4 (merged, PR #214). Composes with **#184** (ed25519 record signing) per §4.2.

## Scope (B1 — the group-MAC rung ONLY)
- Append + verify a **truncated HMAC-SHA256** over broadcast payloads, keyed by a fleet-shared **GROUP_KEY**.
- **Authenticity, NOT confidentiality** (§4.3 — payloads stay plaintext; confidentiality = SKIP for the hobby threat model).
- **NOT** ESP-NOW hardware encryption (dead, #36). **NOT** unicast-mesh/Fork A (SKIP). **NOT** #184 ed25519 signing (separate issue; composes).

## Records list (what carries the MAC)
- **Every SMOLv1 broadcast frame** carries the group MAC: HELLO, TIME, FAM, RELAY/UP2 flood, BATT2/GRID2, SNK, BEACON, MC/crown, broadcast DIAG/STAT/SCAN.
- **8 B (64-bit)** truncation for high-rate frames (HELLO 2 Hz×N, TIME, FAM, RELAY, BEACON) — online-forgery-only, 2⁻⁶⁴/attempt ≈ 10¹¹ yr (§4.1).
- **12 B (96-bit)** for low-rate control frames (MC/crown, CFG) where airtime allows more margin.
- The **3 catastrophic records** (crown MC, reboot/config INTENT, OTA arm) additionally carry the **ed25519 sig (#184)** — `verify-MAC → verify-sig → parse` (§4.2). On `OTAM` the MAC is additive to its existing sig.
- Nuisance-tier: **MAC only, no sig** — sufficient outsider defense (§8).

## Wire format
[...]

Issue #189, opened 2026-07-19
audit: inspirations → implementation coverage — every borrow mapped shipped/deferred/gap


jphein:
## What (JP 2026-07-19)
Turn the reflection "did we implement everything we learned from our inspirations?" into a durable, auditable checklist. For each documented borrow across all 18 inspiration projects, map it to one of: SHIPPED (cite the merged PR/commit), DEFERRED (why + when it becomes worth it), or GAP (learned but neither shipped nor deliberately deferred). File tracking issues for genuine GAPs so nothing stays learned-and-forgotten.

## Source of truth
- The smol-inspirations memory (18 projects in 5 threads: mesh/routing — Meshtastic, painlessMesh, ZHNetwork, Althea, esp_wifi_repeater · local-first ESP↔HA — WLED, ESPHome, Tasmota, lattic, Battery-Emulator, ESP3D, esp-link, Bermuda/ESPresense · cheap-ESP audacity — atomic14, LeafMiner, Marauder · remote-mgmt — PiKVM · lean embedded — GRBL). Read it via the mempalace CLI or the session memory dir.
- The two study docs already produced: docs/superpowers/research/althea-babel-study.md (#163) + mesh-ledger-study.md (#181) — their ADOPT/DEFER/DON'T verdicts feed directly in.
- Current code + closed/open issues for the "shipped vs planned" ground truth (verify, don't assume — the ETX ed25519 "already in-tree" was only half-true).

## Deliverable
docs/superpowers/research/inspirations-coverage.md — a table per thread: borrow → status (SHIPPED cite / DEFERRED why / GAP) → tracking issue #. Then file issues for the real GAPs. Highlight the top 3 highest-value unshipped learnings. Study/audit only, no firmware.

## Known anchors (verify + extend)
- Meshtastic gateway-model + routed-mesh/elections → SHIPPED (#13/#14/#121/#137, MQTT gateway).
- esp_wifi_repeater coexist-precedent → SHIPPED (#40); gateway-health publish → SHIPPED (DIAG/peers/mesh-channel).
- Althea/Babel ETX → BUILT-not-proven (#164); metric-blend → FILED (#180); ledger → studied (#181→#182-185); Babel-wholesale + L1 → DELIBERATELY-DON'T.
- The 12 non-mesh inspirations (WLED/ESPHome/Tasmota/esp-link/Bermuda/atomic14/etc.) — mostly ethos/identity per the doc; confirm which lent anything actionable + whether it landed.

[...]

Issue #184, opened 2026-07-19
mesh-ledger: on-device ed25519 signing (v2 authenticity) — the key-on-device gate [#181 L3]


jphein:
## What
Add **on-device ed25519 signing** to the mesh-ledger (#182) for authenticity — making records tamper-PROOF (forgery-resistant), not just tamper-evident. From the #181 study (`docs/superpowers/research/mesh-ledger-study.md`), ADOPT-lite but **GATED** — this is a real security-posture decision, not a freebie.

## The gate (why this is separate from #182)
The #181 study's sharpest finding: "ed25519 already in-tree" is **half true**. On the C3, ed25519-compact is compiled **verify-only**, over *public* OTA manifests; the OTA signing key lives **off-device** (`tools/ota_publish.sh`, "never on disk" on the board — deliberate posture). A *signed* ledger requires:
1. The **sign** code path compiled in (not today). Note the #32 build-gate: ed25519-compact's scalar-mult unrolls to ~1 MB without the per-package `opt-level="z"` trick — the sign path needs the same care.
2. A **per-board secret key on the C3** — every board becomes a signing oracle; a physically-compromised board can forge its own identity. For "trust among your own boards" that's usually acceptable, but it's a decision.
3. MTU: a 64-B sig eats into the 227-B UP2 budget — pairs with #182's compact encoding.

## Why it's worth it (when unlocked)
Anti-cheat + loot-provenance for the mesh-RPG: a forged inventory/loot record fails its signature. For pure fleet-provenance (audit trail), the hash chain (#182) already suffices — signatures matter where a *node itself* might lie.

## Recommendation
Ship #182 unsigned first. Take this on only when the RPG/anti-cheat need is concrete AND JP signs off on a per-board key. Depends on #182.

Refs #181, #163, #32.

jphein:
[...]

PR #206, opened 2026-07-19
docs(spec): #190 + #184 mesh auth — layered transport + record-signing design

jphein:
Design-only spec (Task #30 Part 2) for **#190** (ESP-NOW PMK/LMK transport auth) and **#184** (on-device ed25519 record signing) as **one two-rung system** — they rhyme but defend different attackers at different layers. No firmware; JP decides the build order.

`docs/superpowers/specs/2026-07-19-mesh-auth-design.md`

## The thesis
- **#190 — transport auth (symmetric, cheap, broad):** AES-CCM on every ESP-NOW frame in hardware. Stops the *outsider* on the RF channel for ~free (one shared PMK, zero per-frame bytes).
- **#184 — record auth (asymmetric, targeted, 64 B):** ed25519 sigs on the *3 catastrophic, low-rate records only* (crown MC, reboot/config INTENT, OTA arm). Unforgeable even if the symmetric key leaks.
- **Layer them.** #190 = front-door lock; #184 = safe inside. smol **already ships the #184 pattern** for its most dangerous frame (the OTA arm is ed25519-gated, fail-closed) — so the code path + precedent exist.

## Covered (all of the assignment's required sections)
- **Threat model:** unauth ESP-NOW RF (wide open) vs broker-authed MQTT — and MQTT records *re-exposed* when relayed as `SMOLv1 CFG`/HELLO. Ranked injectables (crown-hijack, reboot-brick, TIME-poison…) + attacker classes each rung stops.
- **Reuse:** the shipping `verify_signature` + `OTAM` signed frame + sha256 + `dl_seq` (anti-replay) + `secrets.rs` (PMK home). Only genuinely-new firmware = turn on encryption + on-device signing.
- **250 B MTU budget:** full-sig (64 B, catastrophic-only) / truncated-MAC / chain-of-trust-from-a-signed-root tiers; rule = no sig unless the record can brick/hijack the fleet.
- **Key distribution + rotation:** PMK symmetric in `secrets.rs`; INTENT/OTA signing key off-device (OTA precedent); optional per-node on-device key = the #181-L3 gate; overlap-window rotation so a half-rolled fleet still verifies.
- **Mixed-fleet rollout:** #190 is a **hard cutover** (needs a dual-listen transition + canary — a self-inflicted #204-class deafness risk); #184 is **additive** (observe→enforce, `verify or drop` like OTAM). Neither may brick/partition an old board; must compose with #155/#146/#136/#137/#150/#204 (verify-then-parse, filter in front, never a rewrite).
- **What NOT to sign:** high-rate/nuisance frames (HELLO/RELAY/DIAG…) ride #190 only — airtime on the deaf-⅓ radio is the real budget.
- **Build order:** cheapest-first (sign INTENTs with existing verify → sign #155 channel_hint → #190 broadcast-encryption spike → optional crown-MC on-device signing → chain-of-trust).
- **Cross-ref #181/#182:** this is the ledger study's authenticity rung; same signed-envelope shape whether or not the ledger is built.

## Open questions surfaced (not guessed)
Top unknown: does `esp-wifi` ESP-NOW support **PMK-encrypted broadcast** or only unicast LMK? smol's mesh is broadcast-heavy — this single API answer decides #190's value. Plus the on-device-key gate, rotation appetite, enforce-threshold, and signed-INTENT replay window.

Refs #190 #184 #181 #182 #189.

### DIFF
--- a/docs/superpowers/specs/2026-07-19-mesh-auth-design.md
+++ b/docs/superpowers/specs/2026-07-19-mesh-auth-design.md
@@ -1,0 +1,213 @@
+# smol mesh authentication — a layered design for #190 (transport auth) + #184 (record signing)
+
+**Issues:** [#190](https://github.com/jphein/smol/issues/190) (ESP-NOW PMK/LMK link auth — the #189 GAP, learned from Althea #12) · [#184](https://github.com/jphein/smol/issues/184) (on-device ed25519 record signing — the #181 L3 key-on-device gate)
+**Companions:** [mesh-ledger study](../research/mesh-ledger-study.md) (#181/#182 — the sequencing/tamper-evidence layer this sits beside) · [Althea/Babel study](../research/althea-babel-study.md) (#163)
+**Status:** research / design only — **no firmware**. JP decides the build order.
+**Author:** morpheus-155 · **Date:** 2026-07-19
+
+---
+
+## 0. Thesis
+
+The two issues **rhyme**: both add authenticity to a mesh that today trusts every byte it hears. But they operate at **different layers and defend different attackers**, so they are **one system with two rungs**, not one feature:
+
+- **#190 — transport auth (ESP-NOW PMK/LMK).** A cheap, broad, symmetric shield: AES-CCM on *every* ESP-NOW frame in hardware. Keeps anyone **without the shared key** off the RF wire entirely. Defends the **outsider** (a stranger on ch1/6/11 with an ESP32). Cost: ~free (hw crypto), one shared secret.
+- **#184 — record auth (ed25519 per-record signing).** An expensive, targeted, asymmetric proof: a 64-byte signature on the *few catastrophic records*. Makes those records **unforgeable even if the symmetric key leaks** or a board is extracted. Defends the **insider / key-compromise** for the records that can brick or hijack the fleet. Cost: 64 B/frame + signing CPU + the key-on-device decision.
+
+**Layer them.** #190 is the front door lock (stops randoms cheaply); #184 is the safe inside (protects the crown jewels even past the front door). Ship #190 first (broad, cheap, no MTU cost); ship #184 for the short list of records that are worth 64 bytes. **smol already ships the #184 pattern for its single most dangerous frame** — the OTA arm — so the precedent and the code path exist.
+
+---
+
+## 1. Threat model
+
+### 1.1 Two transports, two very different trust levels
+
+| Path | Who can write to it | Current auth |
+|---|---|---|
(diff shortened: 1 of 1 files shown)


jphein:
## Decisive answer: ESP-NOW native encryption is **unicast-LMK only — PMK-encrypted BROADCAST does not exist** (research, #36)

Chased to ground across all three layers; they agree.

### 1. Espressif docs (authoritative)
> **"Encrypting multicast vendor-specific action frame is not supported."**
> ([ESP-IDF ESP-NOW API ref](https://docs.espressif.com/projects/esp-idf/en/stable/esp32c3/api-reference/network/esp_now.html))

ESP-NOW broadcast (`ff:ff:ff:ff:ff:ff`) is a multicast/broadcast action frame → **cannot be encrypted**. Encryption is per-peer:
- **PMK** (`esp_now_set_pmk`) "is used to encrypt **LMK** with AES-128" — it is a *key-encryption-key*, NOT a broadcast data key.
- **LMK** (per unicast peer) "is used to encrypt the vendor-specific action frame with CCMP." "If the LMK of the paired device is not set, the frame will **not** be encrypted."

### 2. ESP-IDF blob header (`esp_now.h`, in the NuttX 3rdparty @ pinned b90b1837)
- `ESP_NOW_MAX_TOTAL_PEER_NUM 20`, `ESP_NOW_MAX_ENCRYPT_PEER_NUM 6` (current docs say default **7**, max **17** via `CONFIG_ESP_WIFI_ESPNOW_MAX_ENCRYPT_NUM`). Either way: **encrypted peers cap ≪ a 4–30 broadcast fleet**, and it's unicast-only regardless.
- `esp_now_fetch_peer`: *"Only return the peer which address is unicast; for the multicast/broadcast address, the function will ignore…"* — broadcast peers are second-class; you don't encrypt them.

### 3. esp-wifi 0.15 Rust API (what smol can actually call)
`esp-wifi-0.15.0/src/esp_now/mod.rs`: `set_pmk(&[u8;16])` (L453) + `PeerInfo { lmk: Option<[u8;16]>, encrypt: bool }` (L214) → `add_peer` passes both to the blob. So native encryption **is** wireable from Rust — but `encrypt`/`lmk` are **strictly per (unicast) peer**. There is no broadcast-encrypt knob because the blob has none.

**Cross-check (#200 RIOT):** RIOT hard-`#error`s "if esp_wifi is used, esp_now must be unicast" — it *mandates* unicast under WiFi precisely because the broadcast/managed path can't carry encrypted semantics. RIOT sidesteps by never broadcasting. **smol is flood-first, so it cannot sidestep** — see the #190 fork.

**Verdict on native ESP-NOW encryption for smol's mesh transport: SKIP.** It cannot encrypt the broadcast flood, and the unicast alternative caps at ~7 encrypted peers and contradicts flood-first. Design implications for #190 in the next comment there.

Issue #233, opened 2026-07-20
chore: upgrade wave — esp-hal 1.1 / esp-radio 0.18 matched-set migration + edition 2024 (coupled to #198)


jphein:
## What
Coordinated **matched-set** upgrade of the esp-hal / esp-wifi(→**esp-radio**) firmware stack + its coupled deps + **the esp-hal-embassy → esp-rtos executor rename** + edition 2024. Split out from the #52 version-currency audit: these CANNOT be bumped piecemeal (esp-radio links esp-hal INTERNAL APIs; a `Rng::new()` signature change alone breaks the set), so per the audit they're deferred here as one migration rather than blocking the safe-now currency PR.

Per JP's latest-everything policy this is the biggest remaining "tactical fossil" — but it's a project, not a `cargo update`.

## The coupled set — target = the esp32c6-watch's proven-on-HW resolved lock (matched set, not "latest")
| Crate | Pinned now | **Target (watch-locked)** | Note |
|---|---|---|---|
| esp-hal | =1.0.0-rc.0 | **1.1.1** | new `esp_hal::init(Config)` / `TimerGroup` API; `Rng::new()` arg change is the canonical trap |
| esp-wifi → **esp-radio** | =0.15.0 | **0.18.0** | **RENAMED crate** — new os-adapter/scheduler surface; feats `wifi,ble,coex,esp-now` |
| **esp-hal-embassy → esp-rtos** | =0.9.0 (unused; superloop) | **esp-rtos 0.3.0** | **RENAMED/REPLACED** — esp-hal-embassy is GONE; esp-rtos provides the embassy time-driver + async scheduler (`esp_rtos::start(timer, swint)`) |
| esp-wifi-sys | =0.7.1 | **subsumed by esp-radio** | esp-radio brings its own sys layer; drop the direct dep |
| esp-alloc | =0.8.0 | **0.10.0** | esp-radio `^` |
| esp-storage | =0.7.0 | **0.9.0** | otadata/flash write API changed (OTA-critical) |
| esp-bootloader-esp-idf | =0.2.0 | **0.5.0** | otadata slot/state API changed (OTA-critical) |
| esp-backtrace | =0.17.0 | **0.19.0** | |
| esp-println | =0.15.0 | **0.17.0** | |
| smoltcp | =0.12.0 | **0.13.1** | now driven under embassy-net, not hand-polled |
| embassy-executor | — | **0.10.0** | new (the async migration proper, #198) |
[...]

Issue #49, opened 2026-07-10
feat/observability: retained MQTT DIAG topic — on-device counters (flush/verify/relay) for HW-verify without UART0


jphein:
The UART0-capture gap has blocked on-device HW-verification 3x — the #32 signed-OTA refuse-line, and the #9 flush-win perf-proof. Key runtime counters log to UART0, which the USB-JTAG capture can't read.

**Fix:** expose key counters via a retained MQTT DIAG topic, e.g. `smol/<id>/diag` = { flush success/fail, OTA verify pass/fail, relay stats, ... }. The GATEWAY publishes its own directly; leaves relay theirs over the mesh (later).

**Payoff:** on-device metrics become MEASURABLE via HA/broker (no UART0 adapter) → would have PROVEN #9's flush win + the #32 refuse-path, and unblocks ALL future HW-verification (perf proofs, refuse-path proofs, relay health).

Low-priority (infra/observability). Complements #44 — #44 verifies the IMAGE (build reproducibility), this verifies BEHAVIOR (runtime). Refs #9, #32.

jphein:
**Design delivered (parallel design fork — full doc scratchpad/issue-49/).**

**Payload:** retained `smol/<id>/diag`, pipe-delimited/positional/versioned, ≤120 B:
```
DIAG|v|up|bt|rr|fok|ffl|rtx|rrx|rdp|vok|vfl|ook|ofl|lp|fh
```
uptime · boot-count · reset-reason · flush ok/fail · relay tx/rx/drop · verify ok/**fail** · OTA ok/fail · loss% · free-heap. Cumulative-since-boot; up+bt+rr let HA detect a silent reboot that explains a counter reset. Append-only (HA guards on length).

**Publish/relay — reuses existing plumbing (3-for-1):** counters are plain `RadioManager` fields bumped at the *existing* log sites (flush ok/fail at mqtt_session CONNACK/timeout; relay tx/rx/drop in mode.rs; verify/activate in ota.rs; BenchTracker already has tx/rx/lost→loss%). `diag_snapshot()→DiagStats` (Copy) + panic-free `diag_frame()` (twin of the PEERS serializer). Gateway publishes its own `smol/<id>/diag` RETAINED via a new `diag: &[u8]` param (exactly like `peers`). Leaves ride the **#50 leaf→gateway STAT relay** (same build-once uplink); gateway republishes as `smol/<leaf>/diag`. Zero new transport.

[...]

Issue #57, opened 2026-07-10
feat: The Mesh Familiar — a living creature that inhabits the fleet (FLAGSHIP)


jphein:
**From the Fable visionary voice-session (2026-07-10). JP: "That sounds rad. I love it." — flagship, build first.**

A living creature that lives *across* the smol fleet — Tamagotchi × Weasley-clock × ambient art:
- **Exactly one screen at a time** — it inhabits one board; **wandering to another board = a real ESP-NOW state handoff** (rides the shipped World-Snake shared-world netcode).
- **Senses the house** via the retained HA downlink it already receives: solar charging → basks; grid spike → startles; night → sleeps.
- **Identity**: a realm-sigil name + species (reuses the sigil system).
- **Grows with uptime**; **fed via the BOOT button**.
- **Pure firmware on the shipped substrate** — mesh + HA downlink + sigils + OTA all exist; OTA ships it fleet-wide. This is the demo that makes the whole project's story irresistible.
- **Long arc**: the familiar travels between *houses* over MQTT — "an owl carrying messages" — once friends have nodes.

Builds on: World-Snake netcode, the retained HA BATT/GRID downlink, realm-sigil, OTA. Sequencing (JP-approved): **this first** → Marauder's Watch → the story. Feature-scale, parks for a dedicated build.

jphein:
## Mesh Familiar v1 — design spec (wisp, flagship pass)

Scoped to ship in **one fw wave** on the shipped substrate. The creature lives on **exactly one board at a time**, is animated only there, and everyone else's screen is a Weasley-clock hand pointing "it's at *Herald*". Walking up to the board it's on and seeing it alive is the payoff.

**Substrate — it composes two shipped patterns, invents nothing:**
- **World-Snake's broadcast/dead-reckon** (per-id phase-jittered heartbeat, bounded inbox) + **the MC single-owner election** (`seq`+id arbitration, RSSI-weighted dead-owner takeover). The familiar = "World-Snake with an exactly-one-holder rule".
- **New `SMOLv1 FAM` frame** (byte-7 `'F'` is free): `type(H/X/C) · holder_id · target_id · seq · species_seed · birth_unix · last_fed_unix · mood` ≈ 18 B. `broadcast_fam`/`take_fam` mirror `broadcast_snk`/`take_snk`.

**Exactly-one invariant (the technical heart):**
[...]

PR #236, opened 2026-07-20
fix(fw): #226 blank-otadata self-overwrite — target ota_1 not running slot [HW-CANARY-HELD]

### LINKED ISSUE #226: fw: blank otadata → OTA targets the RUNNING slot (self-overwrite); add first-boot otadata init
## Summary
On a board with **blanked otadata** (e.g. after the USB-bootstrap `espflash erase-region 0xf000 0x2000`), smol's `inactive_slot()` returns `Slot0` — **the slot the app is currently running from**. A subsequent OTA would then download into and erase the *running* slot (XIP), crashing mid-download instead of performing a clean A/B swap. Latent brick/OTA-failure on the bootstrap path.

## Mechanism (traced against feat/204 == main)
- `partitions-ota.csv` has **no factory partition** → the ROM bootloader falls back to **`ota_0`** on blank/invalid otadata.
- `esp-bootloader-esp-idf 0.2.0` `current_slot()` maps blank otadata (both `ota_seq == 0xFFFFFFFF`) → `Ok(Slot::None)` (crate `ota.rs:193`), and `Slot::None.next()` → `Slot::Slot0` (crate `ota.rs:64`).
- smol `inactive_slot()` (`ota.rs:673`) = `Some(current_slot().ok()?.next())` → `Some(Slot0)`.
- Bare-metal has **no running-partition detection** (`current_slot()` reads otadata only), so it can't tell it is already executing from `ota_0`.
- Net: booted-from `ota_0` **and** `inactive_slot() == Slot0` ⇒ OTA write target == running slot ⇒ `ImageWriter::feed()` erases `0x20000` (the running XIP region) on chunk 0 ⇒ crash/reset, not a clean install.

## Impact & scope
- Blocks (and can brick) a clean OTA on any USB-erase-region board — i.e. exactly the fleet-bootstrap path.
- **Distinct from** the `at=slot` blocker just fixed (that was a *missing* OTA partition table from a bare `espflash flash <elf>` **without** `--partition-table partitions-ota.csv`; fixed by making `--partition-table` mandatory in the manual flash procedure). Also distinct from #204 (crown coexist RX deafness). This is the residual **firmware-hardening** item that makes USB-bootstrap→OTA safe.

## Proposed fix — first-boot otadata init
At first boot, when otadata is blank, initialize it to reflect the true running slot:
```rust
// ota.rs, called once at early boot (espnow-gated), before any OTA can begin
if matches!(ota.current_slot(), Ok(Slot::None)) {
    // No factory partition ⇒ blank otadata ⇒ ROM booted ota_0 ⇒ running slot IS Slot0.
    ota.set_current_slot(Slot::Slot0).ok();
    ota.set_current_ota_state(OtaImageState::Valid).ok();
}
```
Safe **only** because our table has no factory (blank ⟹ running is `ota_0`) — document that invariant at the call site. After init, `inactive_slot()` = `Slot1`, OTA targets the true *other* slot, and the USB-bootstrap→OTA path works cleanly.

Pair with a **distinct fail code** so a genuinely absent OTA table reports e.g. `at=slot-notable` instead of the current ambiguous `at=slot` (the missing-table case shares the `begin()==None` path). Cheap, and it would have named tonight's root cause immediately.

## Verification plan (HW-gated, rides a bench canary)
1. USB-flash a board **with** `--partition-table partitions-ota.csv`, then `erase-region 0xf000 0x2000` (blank otadata).
2. Confirm DIAG `boot_slot` transitions `255` → `0` after the first-boot init (proves otadata now resolves).
3. Trigger an OTA → confirm it targets `ota_1` (not the running `ota_0`) and completes a clean A/B activate + reboot.
4. Confirm no regression on a normally-provisioned board (otadata already valid → init is a no-op).

## Refs
Refs #204 #26 #6. Root-cause analysis: `scratch/v342-wave/morpheus-89b-at-slot-analysis.md`. Discovered during the `at=slot` forensics, 2026-07-19 (source refutation of the otadata-state hypothesis + this latent corollary).

jphein:
Closes #226. **HW-CANARY-HELD** — boot-critical partition state (app-rollback OFF) → single JP-supervised board before any fleet roll.

## Bug
`inactive_slot()` selected the OTA write-target as `current_slot().next()`. On a BLANK otadata: `current_slot()=Slot::None`, `Slot::None.next()=Slot::Slot0` (esp-bootloader 0.2.0 ota.rs:64) = the ROM-booted RUNNING slot → OTA overwrites the LIVE image (self-overwrite brick).

## Fix (minimal, validated)
Explicit match: `Slot::None => Slot::Slot1` (blank ⟹ ROM boots ota_0 ⟹ inactive is ota_1), else `other.next()` unchanged. **No first-boot otadata init** (supersedes the sketch): `activate()` self-inits otadata on the first real OTA → blank self-heals with ZERO boot-path flash mutation; a boot-time write would be an unnecessary risk. `boot_slot()=255` on blank kept as the unprovisioned signal.

## Cross-check (watch #49)
JP's ESP32-C6 watch (esp-bootloader 0.5) avoids this the same way; the `.next()` footgun survives into 0.5 → **not** auto-fixed by the #233 upgrade, so this fix stands independently. Study: scratch/watch-backport/226-comment.md.

## Gate
KATANA 4-tier green (1.96.1): clippy -D + release espnow,cast,io/default/wifi. espnow-only; default/wifi unaffected. Findings: scratch/226-otadata-init/morpheus-89.md.

### DIFF
--- a/rust/clock/src/main.rs
+++ b/rust/clock/src/main.rs
@@ -354,6 +354,16 @@ fn main() -> ! {
         env!("BUILD_HASH"),
     );
 
+    // --- #226 FIRST-BOOT OTADATA INIT (before diag + #40 bookkeeping) ---------
+    // A freshly USB-flashed board has BLANK otadata; the ROM boots ota_0 (no factory
+    // partition) but leaves the ota-select record absent → boot_slot() reports 255 and a
+    // later OTA would show luna a spurious 255→N "rollback". Write a valid record for the
+    // running slot (Slot0) ONCE so otadata is always well-formed. Runs BEFORE capture_boot_diag
+    // so the diag reads the initialized slot; the inactive_slot() None=>Slot1 net still
+    // protects OTA targeting if this write ever fails. (#226; boot-critical → HW-canary.)
+    #[cfg(feature = "espnow")]
+    ota::init_otadata_if_blank();
+
     // --- #40 leaf-mesh-OTA: EARLIEST-boot self-test bookkeeping ---------------
     // Runs BEFORE any subsystem that can panic (I2C/display/radio) so an early-init
     // crash still trips the crash-loop bound. Only fires when otadata is New/Pending
--- a/rust/clock/src/ota.rs
+++ b/rust/clock/src/ota.rs
@@ -679,7 +679,26 @@ fn inactive_slot() -> Option<Slot> {
         .ok()??;
     let mut region = od.as_embedded_storage(&mut flash);
     let mut ota = Ota::new(&mut region).ok()?;
-    Some(ota.current_slot().ok()?.next())
+    // #226 SELF-OVERWRITE FIX: on a BLANK/uninitialized otadata, `current_slot()` returns
+    // `Slot::None`, and `Slot::None.next()` folds to `Slot::Slot0` (esp-bootloader 0.2.0
+    // ota.rs:64) — i.e. the RUNNING slot, so an OTA would overwrite the LIVE image (a boot-
+    // critical self-overwrite / brick class). A blank otadata always ROM-falls-back to ota_0
+    // (there is no factory partition in partitions-ota.csv), so the running slot is provably
+    // ota_0 and the true inactive write-target is ota_1. Explicit match instead of the blind
+    // `.next()`. Validated against JP's ESP32-C6 watch (esp-bootloader 0.5), which avoids this
+    // exact bug the same way — the `.next()` footgun SURVIVES into 0.5 (current_slot() still
+    // returns None on blank), so the crate bump does NOT auto-fix it; the explicit mapping does
+    // (watch study: scratch/watch-backport/226-comment.md, task #49).
+    //
+    // No first-boot otadata write is needed (supersedes the original init sketch): a boot-time
+    // flash write to boot-critical partition state would be an unnecessary risk, and `activate()`
+    // already makes otadata valid (set_current_slot + New) on the first real OTA. So the blank
+    // state self-heals on the first successful update, with zero boot-path flash mutation.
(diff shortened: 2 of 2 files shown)


jphein:
Added commit **7b44d22** — the first-boot otadata init requested in #57 (task reassigned morpheus-89 → morpheus-89b when 89 hit its context wall). Layers on top of 9cbfaf6.

`init_otadata_if_blank()` runs at earliest boot (before `capture_boot_diag` + the #40 unconfirmed block): if otadata is blank (`current_slot()==Slot::None`), it writes a VALID ota-select record for the running slot (Slot0 — a blank otadata always ROM-boots ota_0, no factory partition). Effect: `boot_slot()` reports **0 not 255**, so luna's rollback automation sees a clean 0→1 on the first OTA instead of an ambiguous 255→N.

- ONE-TIME (only while blank), VERIFY-AFTER-WRITE, FAIL-SAFE (any flash error → no-op).
- 9cbfaf6's `None=>Slot1` stays as the defense-in-depth net if this write ever fails.
- `Valid` (not `New`) keeps `otadata_unconfirmed()` false — a USB flash is not an OTA under self-test, so it must not arm the #40 K-counter / leaf self-test.

KATANA 4-tier green (clippy -D + release espnow,cast,io / default / wifi, 1.96.1); espnow-gated, default+wifi cfg-unaffected; diff secret-scan (IP/MAC/SSID/beginagain) = 0. Still **HW-CANARY-HELD** — boot-critical otadata write, app-rollback OFF → single JP-supervised board before fleet. Design notes: scratch/226-otadata-init/morpheus-89b.md.

Issue #217, opened 2026-07-19
mesh: #204 rung-2 — proactive reassoc at crown-assumption on a bad AP (ch mismatch / rssi<-75)


jphein:
## What
**#204 rung-2: proactive re-association at crown-assumption when the board's own AP is bad** (wrong channel, or too weak). Currently only a comment on #204 (the live-canary rung-2 candidate); filing it properly per research→issues.

## The third failure lane
#204 forensics separated three distinct crown-failure lanes:
- **(a) crown RX-deafness** — the coexist unicast-RX starvation → **rung-1** (detect + reassoc + shed), shipped in the #204 self-heal.
- **(b) sustained flush-fail** — the #146 uplink-death lane (abdicate on flush-fail).
- **(c) bad-AP-at-crown-assumption** ← **this issue.** A board takes the crown while associated to a **too-weak or off-mesh-channel AP**, then **burns a full tenure on a hopeless association** before #146 finally demotes it — churn the mesh pays for.

**Live evidence (canary, 2026-07-19 07:46:55):** id8 took the crown (chose ch6) but its `ap=` telemetry read **`ap=1:-82:…`** — associated to a **ch1 AP at −82 dBm**. Tenure lasted ~2 min and died on the **uplink** (`fok=0`, dethroned) via the #146 flush-fail lane. The rung-1 deaf-detector correctly never fired (a `!ok` flush is #146's disease, not RX-deafness — the signal-purity separation is working as designed). So lane (c) slips between rung-1 and #146: the board never should have committed the tenure on that AP.

## Proposed fix (rung-2)
At **crown assumption**, before committing the tenure, check the board's **own `ap=`** (already published — the #204 rung-1 "commit-1" field, `net/mode.rs:2970`, `ap=<ch>:<rssi>:<bssid>`):
- if **`ap.ch != mesh_ch`** (off the coexist channel), **or** **`ap.rssi < ~−75 dBm`** (too weak),
- then **`reassoc_ch6_prefer()` BEFORE committing the crown tenure** — reusing the rung-1 machinery verbatim (`net/mode.rs:2744`), which already prefers a ch6 (mesh-channel) AP.
- Only commit/hold the crown once associated to an acceptable AP (or fall back to abdicate if none — don't crown onto a hopeless link).

The `ap=` field makes this **locally decidable** — no new telemetry, no round-trip. It's a *pre-tenure gate* reusing two primitives rung-1 already ships.

## Why it's separate from rung-1
[...]

Issue #195, opened 2026-07-19
fw(OTA): self-fetch install path lacks a cross-burst consecutive-failure cap


jphein:
## Re-scoped (see the correction comment below for the full loop analysis)

After reading the retry loops, the two originally-filed fixes turned out to be **already present**, and the live symptom that motivated this issue was a **#204 consequence**, not a retry-loop defect:

- **body-byte STALL** — already correct: `run_ota_fetch` resets `stall` only on `writer.written() > chunk_start` (wifi.rs ~4398-99) = image BODY bytes to flash. Header retransmits never advance it, so a zero-body blackout already trips `OTA_MAX_STALL=6`; `OTA_FETCH_BUDGET=300s` bounds a trickle.
- **arm re-check** — already present for self-fetch: `mqtt_session` re-reads the retained arm each burst (`if payload == b"INSTALL"`, wifi.rs ~2544); `take_install_request` is one-shot (`mem::take`, mode.rs ~3291). A *received* arm-clear already stops the cross-burst re-trigger. The relay path already caps consecutive transient attempts (mode.rs ~292 / `leaf_ota_retries`).
- The v341-canary "21+ retries after clear" = **#204**: the deaf crown never *received* the retained-delete (dead downstream). No MQTT-based abort can reach a deaf board — that's for #204 to fix, not this issue.

## The actual gap (this issue, re-scoped)

The **self-fetch** install path has **no cross-burst consecutive-FAILURE cap** — unlike the relay path (`leaf_ota_retries`). A board that repeatedly fails to fetch a staged build will re-trigger every burst for as long as its arm remains (received-)armed, with no give-up-after-N-failures backstop.

**Fix:** after N consecutive failed self-fetch bursts for the staged build, stop auto-re-triggering (require a fresh arm) + back off.

## Deferred

Deferred until **#204** rung-1 (crown dead-unicast self-heal) merges:
- keeps **single-owner on RadioManager** during the live v341/v342 canary (the cap lives in the install-latch, which #204/#27 is already reshaping);
- #204's ~6-7 min crown-shed structurally bounds most of the harm (a deaf crown sheds → becomes a leaf → the leaf/relay path's existing cap applies), so the self-fetch cap's residual value is defense-in-depth for the pre-shed window (~6-10 attempts) + detector-miss variants.

Cross-ref: #204 (root cause + self-heal), #27 (rung-1 payload). Owner: luna-otaui (first right to implement post-#27-merge).

jphein:
[...]

PR #244, opened 2026-07-20
fw: #217 rung-2 — proactive + fetch-leg reassoc on a bad AP [HW-CANARY-HELD]

jphein:
## What
**#217 rung-2** — proactive AP-quality reassoc + fetch-leg deafness sensing, building on the merged **#204/#210 rung-1** ladder. Commit `e7569f5`, +50 lines in `mode.rs`, reuses rung-1's `reassoc_ch6_prefer` (no new radio code).

Rung-1 **reacts** to crown deafness (got_mc-miss streak → ch6-reassoc → shed). Rung-2 acts **proactively** at the two moments a board commits to an AP, and closes the fetch-leg blind spot the #204 pcap exposed.

**Evidence (pcap, v343 roll):** the fetching crown completed the TCP handshake + sent the GET, then **never ACKed byte 0** of the response — deaf within ~1 ms of its own TX — serving AP at **rssi −82**. AND the crown's `cdeaf` stayed **0:0:0** the whole time (the #204 `got_mc` detector is blind to the fetch socket).

## The three rungs
- **2A — crown-assumption AP gate** (`start()`, after the rssi capture, before the coexist-gateway commit): `is_gateway && my_rssi_to_ap < −75` → `reassoc_ch6_prefer()` + re-capture rssi. Preempts crowning on a doomed link (the mesh is about to ride this AP's channel).
- **2B — pre-fetch AP gate** (`run_ota_update`, after `switch(WifiSta)`, before `run_ota_fetch`): `my_rssi_to_ap < −75` → `reassoc_ch6_prefer()`. Best-effort steer-to-better (run_ota_fetch's assoc-wait absorbs the reconnect).
- **2C — fetch-leg deaf → reassoc-before-retry** (`run_ota_update`, on `ok==false`): on a bulk-deaf `ota_fail` stage → `reassoc_ch6_prefer()` so the buffered retry lands on a healthy link. Closes the `got_mc` blind spot (cdeaf stayed 0:0:0 in the pcap). espnow-gated.

`const CROWN_AP_RSSI_MIN = -75` (pcap deaf AP was −82; between the reelect −65/−78 buckets).

## Design decisions
- **rssi trigger only** this rung (the pcap-proven one); ch-mismatch is a follow-up — no trivial current-AP-channel accessor at the hook sites yet.
- **No defer in 2B:** `run_ota_update`'s `false` == self-fetch FAILURE, which feeds the #195/#225 fail-cap; a defer-as-failure would poison the cap. Cap-neutral true-defer = **rung-2D follow-up**.
- **2C espnow-gated:** `ota_fail_is_bulk_deaf` is espnow-only, but `run_ota_update` also compiles on the wifi-only tier — an ungated call would fail *only* the `wifi` release build (verified: the wifi tier is green with the gate).

## Follow-ups (spec'd, not in this PR)
- **rung-2D:** cap-neutral pre-fetch defer via a caller-side `ap_ok_for_fetch()` pre-check.
- **ch-mismatch trigger** once a current-AP-channel accessor is exposed at the hooks.

## Gates
KATANA **4-tier green** (rustc 1.96.1): clippy -D warnings + release `espnow,cast,io` / `default` / `wifi`. 2A/2B are wifi-available; 2C's espnow gate keeps the wifi tier green. Diff secret-scan (IP/MAC/SSID/`beginagain`) = 0.

## ⚠️ HW-CANARY-HELD
The scan + reassoc radio path — and whether ESP-NOW coexist follows the new STA channel — needs JP's rig (same discipline as rung-1/#210). **Not flashed.** Canary matrix + verification plan in `scratch/217-reassoc/morpheus-89b.md`.

Refs #217 #204.

### DIFF
--- a/rust/clock/src/net/mode.rs
+++ b/rust/clock/src/net/mode.rs
@@ -1416,6 +1416,13 @@ const DEAF_SHED_M: u8 = 2;
 /// exhausted. A got_mc miss is itself bulk-dead proof (small unicast dead ⇒ bulk dead), so this is a
 /// valid — if deliberately conservative — shed path; a live bulk-fetch failure sheds sooner.
 const DEAF_SHED_DEEP_STREAK: u8 = 8;
+
+/// #217 rung-2: crown-assumption + pre-fetch AP-quality floor. A crown that associates to an AP
+/// weaker than this proactively ch6-reassocs to a stronger one BEFORE committing as gateway and
+/// BEFORE a self-OTA fetch. Evidence: the #204 pcap deaf AP was rssi -82 — the crown went
+/// unicast-RX-deaf within ~1 ms of its own GET TX. -75 leaves headroom above that and sits between
+/// the reelect rssi buckets (-65 / -78). PROACTIVE (preempt) vs rung-1's REACTIVE detect→heal.
+const CROWN_AP_RSSI_MIN: i8 = -75;
 /// Recently-completed `(src_mac, msgid)` memory. A lost RELAYACK makes a leaf
 /// retransmit an ALREADY-complete message; we must re-ACK it but NOT re-enqueue
 /// (else duplicate UDP delivery — finding 3). Ring of the last few completions.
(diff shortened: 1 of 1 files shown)


jphein:
Added commit **cff944d** — the per-body-progress stall timer, which makes rung-2C *timely*.

Investigation confirmed the gap you flagged: `run_ota_fetch`'s inner body-read loop only exited on done / peer-close / the **300 s global DEADLINE**. On an RX-deaf fetch leg the connection delivers no bytes AND never *receives* the server's FIN/RST, so a byte-0 hang spun the **entire ~300 s budget** before classifying — and the outer `OTA_MAX_STALL` guard never fired (control never left the inner loop). So 2C's reassoc only kicked in after ~5 minutes.

Fix: `OTA_BODY_STALL` (8 s) — zero body advance for the window → break the chunk into the **existing** stall guard → `STALL` after `OTA_MAX_STALL` (~tens of seconds) → 2C reassocs before the buffered retry. The timer resets on any byte of progress, so a slow-but-advancing link never false-trips (8 s ≫ any healthy inter-packet gap). Reuses the proven stall mechanism — no new failure path.

KATANA 4-tier green. Still HW-CANARY-HELD.

Issue #198, opened 2026-07-19
spike: Embassy async executor — vertical-slice port to de-risk the migration (clock + one radio path)


jphein:
## Why

smol has been paying Embassy's price in installments: every hand-rolled poll state machine (#89 Stage 1 NtpMachine, Stage 2a mqtt_session rework, the OTA chunk loop) is a manual version of what `async`/`await` generates for free. Those installments were right for shipping velocity — and they're also down payments: code already shaped as explicit state machines ports to async far more easily than a blocking superloop would.

Standing motivators:
- **#22 native BLE** was refuted precisely because blocking init wedges the C3; the agreed revisit path is Embassy (`trouble` BLE host runs on it).
- **#89-class responsiveness**: under Embassy, the flush is one task among many; the display task keeps ticking between await points — the compiler writes the state machines.
- **Mesh-deaf windows**: async radio tasks + coexist should shrink (not eliminate) the deaf windows without bespoke scheduling.
- Espressif maintains `esp-hal-embassy` first-class; `esp-wifi` exposes async APIs including ESP-NOW.

Full migration is a campaign (superloop dies, every plugin tick model changes, all radio behavior re-verifies through the HW gate). This spike de-risks it for a weekend's effort.

## Scope — vertical slice only

Port to a bench C3 on `esp-hal-embassy`:
1. Boot + clock render loop (display task).
2. ONE radio path: ESP-NOW RX → peers table (beacon listening), running concurrently with a WiFi association + SNTP burst.

Explicitly OUT: elections, OTA, MQTT session, plugins, menu. No merge intent — spike branch (`spike/embassy`), separate crate/workspace so the firmware `Cargo.lock` and the rustc 1.96.1 pin are never entangled (same isolation pattern as `rust/viz/`). The spike workspace may carry its own `rust-toolchain.toml` if Embassy's MSRV requires newer than 1.96.1 — that delta is itself a finding.

## Measurements (the deliverable)

| Metric | Baseline (superloop) | Spike |
|---|---|---|
| Image size | ~1,006,688 B (v341) | ? |
| Static RAM / heap headroom | current hmin telemetry | ? |
| Mesh-deafness during WiFi burst | beacon-loss count (the #40/#155 coexist metric) | ? |
| Clock jank during assoc/DHCP/SNTP | #89 S1 behavior | ? |
| Build time (katana, -j2) | current | ? |

## Exit criteria

[...]

Issue #52, opened 2026-07-10
feat: remote reset/reboot button per node (dashboard → fw reboot, gateway-relayed to leaves)


jphein:
Add a per-node **Reset** button to each dashboard node box that remotely reboots that node.

**FW:** honor a reset command → controlled software reboot (esp_hal reset). Gateway (id7) direct via MQTT; **leaves relay through the gateway** (same path as screen/units config, but this is a COMMAND, not config).

**⚠️ CRITICAL: the reset command topic MUST be NON-RETAINED (transient).** A *retained* reset payload would re-fire on every reconnect/subscribe → **permanent reboot-loop / soft-brick**. Contrast with config (retained) vs actions like install/reset (transient) — this is the same discipline as the OTA `install` path.

**HA:** `button.smol_<id>_reset` per node → publishes the transient reset command (`smol/<id>/cmd/reset`, retain:false).

**Dashboard:** a Reset button in each node box (consider a hold-to-confirm / tap-guard so it's not fat-fingered).

**Lanes:** FW (lucid, behind #50/#40) · HA (nebula) · dashboard (luna). HA+dashboard can land ahead of fw (button publishes; fw honors when the command handler ships). Relates: #21 (config relay), #48 (per-node command pattern).

jphein:
**Merged to main via PR #85** (remote reboot (key R, cache-bypass + boot-debounce)), on the #56 keyed-CFG channel. 4-profile compile-green + oracle rebase-integrity gate. **Hardware validation folded into the morning consolidated canary** (fleet unplugged overnight).

jphein:
**Reopening for ONE morning-canary check** (code is merged + correct — this is a verification hold, not a bug): oracle flagged that #52's primary reboot-loop guard is the **retain:false** transient publish on `smol/<id>/cmd/reset` — the fw's 10s boot-debounce only covers the first 10s after boot. Morning canary MUST confirm luna's HA publish for the reset button uses **retain:false** (a mistakenly-retained R = slow reboot-loop). Verify → press reset in HA → node reboots ONCE (not repeatedly) → close.

jphein:
[...]

PR #247, opened 2026-07-20
feat: #233 upgrade wave — esp-hal 1.1 / esp-radio 0.18 matched set + edition 2024 [HW-CANARY-HELD]

jphein:
## What

The coordinated **matched-set** upgrade from #233 — esp-hal / esp-wifi(→esp-radio) / scheduler /
storage / bootloader move together (a mixed set won't compile) + **edition 2021 → 2024**. Rides
the #198 spike groundwork and mirrors the esp32c6-watch reference (adapted C6→C3).

Matched set: **esp-hal 1.1.1 · esp-wifi 0.15 → esp-radio 0.18 (RENAMED) · esp-rtos 0.3 (NEW
scheduler) · esp-storage 0.9 · esp-bootloader-esp-idf 0.5 · esp-alloc 0.10 · esp-backtrace 0.19 ·
esp-println 0.17 · smoltcp 0.13 · edition 2024** (+ new: embassy-futures 0.1 for `block_on`).

## Architecture decisions

- **smol stays a SYNCHRONOUS superloop.** esp-radio 0.18 needs a preemption scheduler (esp-rtos,
  replacing esp-wifi's in-crate `builtin-scheduler`), enabled **NON-embassy**: `esp_rtos::start()`
  converts the current context into the scheduler's pinned main task and returns, so the existing
  blocking `main()` runs unchanged while the radio task preempts in. A concurrency-model change
  belongs to #198, not this dep migration.
- **smoltcp shim** (`net/radio_dev.rs`, transitional): esp-radio 0.18 dropped esp-wifi's `smoltcp`
  feature — its STA `Interface` implements only `embassy_net_driver::Driver`. smol drives *raw*
  smoltcp, so this shim re-exposes the raw `WifiRxToken`/`WifiTxToken` through smoltcp's
  `phy::Device`. It's exactly what #198's embassy-net move later deletes.
- **`coex` NOT enabled.** 0.18's `coex` is the WiFi+**BLE** arbiter and its build script hard-errors
  without `ble`. smol's WiFi↔ESP-NOW coexistence (#40/#23) is *same-radio* channel management (the
  old esp-wifi build had no coex either), and pulling BLE in would wedge the C3 in ROM (btdm init
  busy-wait — the "native BLE refuted on C3" finding). So: esp-now only.
- **async→sync bridge.** connect/disconnect/scan are async-only in 0.18; `embassy_futures::block_on`
  drives them from the superloop. `set_config` STARTS the driver (no `start()`/`is_started()`);
  `is_connected()` returns `bool`.

## OTA port (esp-bootloader 0.5 / esp-storage 0.9) — boot-critical

- `Slot` enum is now **private**. The engine uses `AppPartitionSubType` as its currency across
  `ota`/`ota_mesh`/`mode`. The **#226 self-overwrite footgun** is preserved via an explicit
  `other_app()` (`Factory|Ota0 => Ota1`, `Ota1 => Ota0`) — NOT the now-private `.next()` (which
  folded blank→ota_0 = the live image). `current_slot`/`set_current_slot` →
  `current_app_partition`/`set_current_app_partition`; `Ota::new` gains a partition-count arg.
- **Exact safety semantics preserved**: partition-scoped writer that cannot reach the active slot
  or otadata; ed25519-verify-before-any-flash-write; sha256 before otadata touch; word-aligned
  writes + verify-after-write.
- **FlashStorage — per-op `FLASH::steal()` (`ota::flash()`), VERDICT: SAFE (two independent
  source-reads agree; `scratch/233-upgrade/flashstorage-steal-verdict.md`).** esp-storage 0.9 has
  **no runtime once-guard** — the "Panics if called more than once" doc is the *compile-time move
  contract* on the esp-hal peripheral singleton (the lone `.unwrap()` is capacity-detect), and
  esp-hal 1.1.1's `FLASH::steal()` fabricates the zero-state token unconditionally. Per-op steal
  therefore reproduces the exact old 0.7 throwaway semantics, without threading FLASH through the
  whole radio stack or holding a `&mut FlashStorage` borrow across the streaming write.
  **Residual (now human-audited) invariant**, stated in `ota::flash()`'s doc: the risk class is a
  torn write from a concurrent flash user overlapping an in-flight OTA op — callers must never
  hold two instances, and all flash access stays on the main loop (OTA strictly serial, no ISR
  writes flash). This holds *structurally* in the non-embassy superloop; **#198 must re-evaluate
  it deliberately** (at await-point interleaving, switch to a single owned instance in `ota::`).

## Gate results (KATANA, `riscv32imc`, `-j2`)

| Check | Result |
|---|---|
| `cargo check` default / wifi / espnow / espnow,cast,io | ✅ all clean |
| `cargo clippy --features espnow,cast,io -- -D warnings` | ✅ clean (collapsible_if → edition-2024 let-chains) |
| `cargo build --release` default / wifi / espnow,cast,io | ✅ all exit 0 |
| OTA image (espnow,cast,io) vs slot 0x1F0000 | ✅ **1,058,096 B = 52%** (ed25519 opt-z gate intact; up from ~37% — the 0.18 stack is larger, still ~48% headroom) |
| Honest dev version stamp | ✅ `…+dev.9e399f5` |
| Host experiments (relay_compat/etx/sth/ledger/flood/treehead) | ✅ compile clean (pure cores untouched — no drift) |

## Bonus — kills a latent cold-build fragility (duplicate esp-metadata-generated)

Beyond go-latest, this migration removes the root of main's current RED cold-build (the
`portable-atomic`/`unsafe-assume-single-core` host feature-unification leak): on the rc.0 set
**four** crates (esp-hal 1.0.0-rc.0, esp-wifi 0.15.0, esp-println 0.15.0, esp-config 0.5.0) pin
**esp-metadata-generated 0.1.0**, which uniquely pulls the heavy host build crate `esp-metadata`
0.8 (anyhow/basic-toml/indexmap → heapless → portable-atomic) alongside the modern **0.4.0** (no
deps) → the duplicate that leaks the single-core feature host-ward. The #233 set puts **every**
crate on esp-metadata-generated **0.4.0** — verified: this branch's Cargo.lock has ZERO 0.1.0 and
ZERO `esp-metadata`. So the fragility is a symptom of the stale pins, and the matched-set bump is
its structural cure. (main's focused fix is owned separately by v346; rebase onto it when it lands.)

## ⛔ CANARY FINDING — HARD BLOCKER (do not merge)

First bench-C3 (id13) canary run: the migration image **panic-loops ~every 80 s** on hardware
(MQTT DIAG: `rst=panic`, boot 11→12, up≈79 s→reboot; it also crowned a ch1 AP and pulled the mesh
off ch6 while unstable). A panic-looping build can't ship regardless of the coexist-disease question,
so **#247 is blocked** on root-causing + fixing this.

- **Why serial-silent (0 bytes on panic):** `custom_halt()` `software_reset()`s with no flush, so the
  esp-backtrace dump is truncated out of the USB-JTAG FIFO before the host reads it. Fixed for capture
  via a forensics image (temp 6 s drain-delay in `custom_halt` + `ESP_LOG=info`; edit reverted, branch
  clean). Artifact for the next canary flash: `scratch/233-upgrade/forensic-id13.{bin,elf}`.
- **`rst=panic` = a genuine Rust panic** (overflow-checks off). Live backtrace capture is **HW-gated**
  — there is no spare C3 on katana (the "bench board" was JP's ESPHome laundry BLE proxy, since
  restored), so the panic was root-caused by **static analysis**:
  - **Subscriber-exhaustion (`EVENT_CHANNEL_SUBSCRIBERS=2`): REFUTED.** esp-radio holds zero persistent
    subscribers (all `.subscriber()` sites transient; driver is publisher-side), smol holds none, and
    `block_on` is serial (≤1 subscriber alive < 2) — the `.expect("Unable to subscribe")` cannot fire.
  - **Regression found + FIXED (0ab3d0a) — leading candidate:** esp-radio 0.18
    `scan_async(ScanConfig::default())` is **unbounded** (`max=None`, heap `Vec` of every AP), where the
    old `scan_n(16)` capped at 16 → OOM-panic risk on the crown-deaf reassoc path (which first fires
    ~79 s in, matching the cadence). Restored the bound with `.with_max(16)` at both scan sites.
    Confidence it's THE panic: moderate — **must be confirmed by a clean canary on a real spare C3.**

## Canary plan — once green, DOUBLES as the coexist-disease experiment

Per the coexist study (`docs/superpowers/research/coexist-disease-esp-radio-018-study.md`
§"How to DEFINITIVELY settle it"): same C3 silicon, new stack, crown scenario — serve ESP-NOW while
pulling a ~1 MB body, **off-ch6 AND co-channel ch6**. The study's open sliver (does the newer WiFi
driver smooth the co-channel duty-cycle case?) gets answered for free. **Do NOT fleet-roll without a
canary** — the esp-radio scheduler/os-adapter change and the OTA otadata path are exactly the class
that fails only on hardware. **[HW-CANARY-HELD]**

Refs #233, #198 (spike), #226 (OTA slot footgun), #52 (audit).

### DIFF
--- a/rust/clock/.cargo/config.toml
+++ b/rust/clock/.cargo/config.toml
@@ -40,20 +40,15 @@ build-std = ["core", "alloc"]
 [registries.crates-io]
 protocol = "sparse"
 
-# #140 esp-wifi RX-buffer tuning (heap-gated — see scratch/smol-ha-batt/140-heap-audit.md).
-# We ran esp-wifi 0.15's stock-low defaults (rx_queue 5 / static_rx 10 / dynamic_rx 32 / rx_ba_win 6),
+# #140 RX-buffer tuning (heap-gated — see scratch/smol-ha-batt/140-heap-audit.md).
+# We ran the stock-low defaults (rx_queue 5 / static_rx 10 / dynamic_rx 32 / rx_ba_win 6),
 # which starve sustained RX and were a co-cause of the mid-body OTA-fetch stalls (nebula finding 3 /
 # esp-hosted #184). The bump is UNLOCKED by growing the heap 72→128 KiB (net::init_heap) — the audit
 # showed ~120 KB of DRAM free and only ~5.9 KB heap headroom at the old size, so tuning buffers on the
-# 72 KiB heap would OOM. Keys/ranges verified against esp-wifi 0.15's generated config table; each
-# static RX buffer ≈ 1.6 KB, allocated at esp_wifi::init from the (now-grown) esp-alloc heap.
+# 72 KiB heap would OOM. Each static RX buffer ≈ 1.6 KB, allocated at radio init from the esp-alloc heap.
 # HW-GATE-HELD: throughput/stability win only provable on the deaf-list/fetch rig (pairs with #143).
-[env]
-# 10→16: more 802.11 RX headroom (IDF throughput profile uses 16–25). +~9.6 KB permanent at init.
-ESP_WIFI_CONFIG_STATIC_RX_BUF_NUM = "16"
-# 32→40: on-demand ceiling; must be ≥ static_rx_buf_num. Freed as the TCP stack drains each frame.
-ESP_WIFI_CONFIG_DYNAMIC_RX_BUF_NUM = "40"
-# 5→8: deeper RX frame queue to ride short bursts without drop.
-ESP_WIFI_CONFIG_RX_QUEUE_SIZE = "8"
-# 6→12: bigger Block-Ack RX window for AMPDU throughput. MUST stay ≤ static_rx_buf_num (16). iperf 9–12.
-ESP_WIFI_CONFIG_RX_BA_WIN = "12"
+#
+# #233: esp-radio 0.18 moved ALL of these RX knobs out of build-env into the runtime
+# `ControllerConfig` — static_rx 16 / dynamic_rx 40 / rx_queue 8 / rx_ba_win 12 are now applied
+# in net::radio_controller_config() (threaded into wifi::new). esp-config HARD-ERRORS on unknown
+# ESP_RADIO_CONFIG_* env keys, and RX_QUEUE_SIZE is no longer one, so there is no [env] block.
--- a/rust/clock/Cargo.lock
+++ b/rust/clock/Cargo.lock
@@ -9,12 +9,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c583acf993cf4245c4acb0a2cc2ab1f9cc097de73411bb6d3647ff6af2b1013d"
 
 [[package]]
-name = "anyhow"
-version = "1.0.104"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "330a5ed07fa54e4702c9d6c4174f74427fc0ef6e214bbd677ae50a5099946470"
-
-[[package]]
 name = "autocfg"
 version = "1.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -27,13 +21,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7b7e4c2464d97fe331d41de9d5db0def0a96f4d823b8b32a2efd503578988973"
 
 [[package]]
-name = "basic-toml"
-version = "0.1.10"
+name = "base64"
+version = "0.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ba62675e8242a4c4e806d12f11d136e626e6c8361d6b829310732241652a178a"
-dependencies = [
- "serde",
-]
+checksum = "9e1b586273c5702936fe7b7d6896644d8be71e6314cfe09d3167c95f712589e8"
 
 [[package]]
 name = "bitfield"
@@ -105,6 +96,7 @@ name = "clock"
 version = "0.1.0"
 dependencies = [
  "ed25519-compact",
+ "embassy-futures",
  "embedded-graphics",
  "embedded-storage",
  "esp-alloc",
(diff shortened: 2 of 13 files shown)


jphein:
**Recommendation: HARVEST, then CLOSE — pending Step 1's number.** Tracked in the Embassy epic #335.

This PR did the #233 work correctly and its verification was complete at the time (checks clean on default/wifi/espnow/espnow,cast,io; `clippy -D warnings` clean; three-tier release exit 0; host experiments compiling). It is not being closed because it was wrong — it is 281 commits behind `main` and `CONFLICTING`, so the content lands faster re-applied onto current `main` than merged. Its **old→new API translation table is the reusable asset** and the guide for that re-application.

**Three things here that `dream/feat-embassy` does not have** (verified by path/content) and that must be carried across:
1. `rust/clock/src/net/radio_dev.rs` (+85) — smoltcp `phy::Device` shim over raw `WifiRxToken`/`WifiTxToken`, needed because esp-radio 0.18 drops the smoltcp feature. May be moot on the async path (embassy-net); confirm before dropping rather than assuming.
2. `0ab3d0a` — `scan_async` bounded to 16 results. The branch contains **zero** `scan_async` references.
3. `448f17f` — the `flash()` source-proven steal verdict + human-audited invariant in `ota.rs`. Brick-adjacent, cheap to lose, expensive to re-derive.

Also settled here and worth preserving: the **no-coex finding** (esp-radio 0.18 coex requires `ble`, whose build script hard-errors, and BLE wedges the C3 — #22).

⚠️ One caveat on this PR's green record: it predates the **stack-floor gate** (#300) — 0 references on this branch, 4 on `main`. Its "all tiers green" was therefore measured without the one gate that exists because a successful link proves nothing about a runnable image. `main` now has 1,840 B of stack margin and the 0.18 stack is larger. **Do not close this until Step 1 has produced that measurement** — if the floor fails, this PR's diff is the artefact to reason from.

jphein:
Closing per #335's harvest-then-close ruling. #233 landed on main via #361 (merged 2026-08-01); all three harvest items were verified present in the landed work (radio_dev.rs shim, scan_async .with_max(16) at both call sites, the flash() steal verdict + human-audited invariant — see the Step-2 delivery comment on #335). The HW canary this was held for happened: the v917 roll on the 0.18 stack, confirmed by self-test on id5/id8/id51 on 2026-08-02. The no-coex finding and the API translation table are recorded on #335.

Issue #183, opened 2026-07-19
mesh-ledger: crown-ordered tree-head checkpoint (CT-style global anchor, no BFT) [#181 L2]


jphein:
> **STATUS (2026-08-01, verified against `origin/main` @ `b2537c4`).** The core, the firmware wiring **and the publish half all landed.** The earlier triage note ("unmerged on a branch") and the follow-up note ("zero references in `wifi.rs`/`mqtt.rs`, so the publish half never landed") are both wrong — see *Where it actually is* below. What remains is the **consume** side.

## Where it actually is

| piece | state | evidence |
|---|---|---|
| L2 CT-Merkle core | **landed** | `391a72a` (PR #223, 2026-07-19) → `rust/clock/src/net/treehead.rs` |
| host tests | **landed, running in CI** | `experiments/treehead_verify`, picked up by `tools/gate.sh`'s `experiments/*_verify` glob |
| module declared | **yes** | `net.rs:130` `pub mod treehead;` (`#[cfg(feature = "espnow")]`, `#[allow(dead_code)]` on the core) |
| firmware wiring | **landed** | `net/ledger_link.rs` supplies the concrete sha256/ed25519 injections and owns per-node state; held as `RadioManager.ledger` (`mode.rs:2085`), constructed at `mode.rs:2366` |
| exercised | **yes, every DIAG cadence** | `mode.rs:3591` `append` → `3592` `chain_summary` → `3604` `sign_and_selfcheck` → `3616` `anchor` |
| **published** | **yes — inside DIAG, not a dedicated topic** | see below |

### Why the "no publish half" grep was wrong

Grepping `wifi.rs`/`mqtt.rs` for `treehead`/`sth` returns nothing, but the anchor *does* reach MQTT. The fields are **constructed strings** built in `mode.rs::diag_record` and carried out by the generic retained `smol/<id>/diag` publish (`wifi.rs`), so no ledger-named symbol ever appears in the publish path. (Same failure mode as the known "a literal grep proves nothing about constructed strings" trap.)

Emitted fields — every node:

- `lgt=<8 hex>` chain tip (first 4 bytes), `lgn=<n>` chain length, `lgok=<0|1>` self-check, `lgk=<0|1>` has signing key

Crown only:

- signed: `lgan=<8 hex>` anchor root, `lgep=<epoch>`, `lgsz=<size>`, `lgsg=1`
- unsigned (no provisioned key): `lgan=<8 hex>`, `lgsg=0`

[...]

Issue #182, opened 2026-07-19
mesh-ledger: per-node hash-chained append-only log (unsigned v1, tamper-evident) [#181 L1]


jphein:
## What
A **per-node hash-chained append-only log** on the smol mesh — the #1 ADOPT from the #181 study (`docs/superpowers/research/mesh-ledger-study.md`, companion to the #163 Babel study). Tamper-EVIDENT (hash chain), **unsigned in v1** (sha256 only — already in-tree).

## Why
smol has no replayable, tamper-evident history. It already has the parts: `dl_seq` (per-source monotonic seq), `boot_count` (crash-safe monotonic head), the flood seen-set (gossip dedup). This wires them into an append-only log. Enables fleet provenance (tamper-evident OTA-install/election/config history — tonight's OOM-cascade saga is the poster use case) and the mesh-RPG world-state substrate.

## Design (compact, C3-grounded)
- Add an optional `{prev16, seq}` header to event frames that matter (elections, OTA installs, config changes; later RPG world-events). `prev16` = **16-byte truncated sha256** of the prior record (2⁻¹²⁸ collision — ample for 30 nodes; author is identified by node-id/MAC, no per-frame pubkey).
- Each node maintains its own chain; the tip stored via `boot_count`'s dual-cell torn-write-safe pattern.
- Records gossip over the existing **flood/UP2** transport, deduped by the seen-set. Replay detects insertion/reorder/truncation.
- **Bounded flash ring** (prune-oldest); a prune publishes the pruned prefix's terminal hash (checkpoint) so tamper-evidence survives pruning.
- **Pure module** `net/ledger.rs` + host tests `experiments/ledger_verify`, mirroring `flood.rs`/`etx.rs`.

## Cost
~300 B RAM (chain heads × roster), a bounded flash ring, **zero new airtime** (free-rides frames smol already sends), sha256 only (no_std, in-tree). One sha256/record on receive (~µs).

## Scope / non-goals
- v1 is **unsigned** (integrity, not authenticity). Signatures = separate gated v2 (#L3).
- Not a CRDT (single-author order) and not BFT. Total-order-across-nodes is the crown-tree-head follow-up (#L2).

Refs #181, #163.

jphein:
[...]

Issue #181, opened 2026-07-19
research: immutable/append-only ledger ON the smol mesh — hash-chain vs CRDT vs light-BFT, C3 cost, what it enables


jphein:
## What (JP 2026-07-19)
A focused study — companion to the #163 Althea/Babel study — on what a **tamper-evident / append-only ledger would look like running ON the smol mesh itself** (not a Linux-router blockchain like Althea's Rita/EVM, which #163 already ruled ~3 orders of magnitude too heavy for a C3).

## The question
smol is a single-radio star, ≤2 hops, 4-30 co-located boards, 250B MTU, esp32c3 no_std, sha256 already in-tree (OTA). It already has ledger-*ish* primitives: `dl_seq` strict-newer downlink gate (≈ Babel seqno), `boot_count` monotonic markers, retained-MQTT as last-write-wins registers. What's the RIGHT shape for a real append-only/tamper-evident log here?

## Scope — compare the design space, C3-grounded
1. **Hash-chained per-node append-only log** — each board signs+appends events (elections, OTA installs, sensor readings, RPG world-events) into its own chain, gossiped over the existing flood; any tampering breaks the hash link. Ed25519 already in-tree (OTA sig). Cost: sha256 + a few hundred bytes/node.
2. **CRDT / gossip-merged log** — conflict-free replicated append log across boards (grow-only set / Merkle-DAG); good for shared state without a coordinator. Cost + convergence on a flood mesh.
3. **Light-BFT / quorum sequence** — is any Byzantine tolerance worth it at "trust-among-your-own-boards" scale, or is that overkill (like full Babel was)?
4. Per-option: C3 RAM/flash/airtime budget, no_std feasibility, MTU fit (250B frames), how it rides the existing flood/UP2 transport, and honestly whether it's ADOPT / ADOPT-lite / ADMIRE / DON'T (the #163 verdict format).

## What it enables (motivate the ranking)
- Fleet provenance: tamper-evident OTA-install history, election audit trail, "which board did what when."
- The **mesh-RPG** (spec 2026-07-15): shared world-state history, item/loot provenance, the gravestone economy, anti-cheat — a shared append-log is a natural world-state substrate. #163 flagged deep many-destination mesh (RPG world routing) as exactly where heavier machinery earns its keep.
- Offline-first: a log that survives partitions + reconciles on heal (smol already does this ad-hoc with retained checkpoints).

## Deliverable
[...]

PR #249, opened 2026-07-20
[HW-HELD, v346, stacked on #190] #181 wire the L1/L2/L3 mesh-ledger cores into fw

jphein:
> **Stacked on #190** (`feat/190-hmac-auth`) — both edit `net.rs`, so one owner did both sequentially. **This PR's base is `feat/190-hmac-auth`; review/merge #190 first.** The diff below is the ledger wiring only.

## What

Wires the **inert, host-tested L1/L2/L3 mesh-ledger cores** (landed via #220/#223/#224) into the running firmware: **L1** [`net/ledger.rs`](../blob/main/rust/clock/src/net/ledger.rs) per-node hash-chain (#182) · **L2** [`net/treehead.rs`](../blob/main/rust/clock/src/net/treehead.rs) CT-Merkle anchor (#183) · **L3** [`net/sth.rs`](../blob/main/rust/clock/src/net/sth.rs) ed25519 Signed-Tree-Head (#184). The cores are pure with **sha256 + ed25519 injected**; this PR supplies the injections (`sha2`, `ed25519-compact`) and the per-node state. Study: `docs/superpowers/research/mesh-ledger-study.md`. **HW-held into v346.**

## How it's wired

- **`net.rs`** — declare `ledger`/`treehead`/`sth` (`#[allow(dead_code)]`: this is a **binary** crate, so a not-yet-wired pub primitive would trip `-D warnings`; they're frozen library-of-primitives with full host-test suites, wired incrementally) + the new `ledger_link` integration (**no** allow — its own quality is enforced).
- **`net/ledger_link.rs`** (new) — `LedgerLink`: this node's L1 chain, the crown's L2 head-set, a monotonic STH epoch, and an optional NVS signing key.
  - `append` (L1), `anchor` (L2 root), `sign_and_selfcheck` (L3: sign the anchor, then **verify-what-you-sign** — re-check the STH signature *and* an inclusion proof of our own head under the signed root before surfacing it). The self-check is what exercises the cores' whole **verify** half (verify_sth/accept/verify_inclusion) with **no new gossip path**.
- **`ota.rs`** — `resolve_ledger_key` reads a 32-byte ed25519 **seed** from `nvs` **sector 3** (distinct from identity sector 0 / floor sectors 1-2; survives every OTA — the image write never touches `nvs`). Modeled on `resolve_node_id`.
- **`mode.rs`** — `RadioManager` holds a `LedgerLink` (key loaded from NVS at construction); `diag_record` appends a per-cadence provenance record and surfaces the chain tip / len / self-verify / has-key, plus (on the crown) the L2 anchor + L3 STH epoch. **The publish path rides the already-retained `smol/<id>/diag` topic** — no new frame/topic surface.
- **`Cargo.toml`** — `ledger-provision` feature.

## Key provisioning — the #184 on-device-key gate (NO key generation)

The signing key is **never generated** in the firmware or CI. `resolve_ledger_key` only **reads**; a node with no key runs verify-only / publishes an **unsigned** anchor — it never fabricates one (contrast `resolve_node_id`, which auto-seeds from a baked const). Provisioning is a **deliberate JP-supervised USB touch**: a `--features ledger-provision` build reads a JP-supplied `SMOL_LEDGER_SEED` (hex, **never committed, logged, or sent over MQTT**) and seeds NVS sector 3 **once, only if erased**. The default / OTA fleet image carries neither the feature nor a seed. The seed never leaves the device.

## Size — the on-device ed25519 SIGN fits the OTA slot

The fat-LTO ed25519 unroll (the trap noted in `Cargo.toml`) is contained by the existing `[profile.release.package.ed25519-compact]` override: **espnow OTA image = 1,032,864 B (1.01 MB) vs the 1.94 MB (0x1F0000) slot — 49% headroom.**

## Gate (4-tier, all green)

- **clippy `-D warnings`:** default · wifi · espnow · cast · io · **ledger-provision** ✓
- **release builds:** espnow (1.01 MB, fits slot) · ledger-provision ✓
- **host tests:** `ledger_verify` · `treehead_verify` · `sth_verify` (the cores) · `relay_compat` · `mac_verify` ✓

## Scope boundary (HW-gated follow-up)

**Peer-STH gossip + acceptance** (verifying a *different* crown's key over the mesh) and **chain persistence to flash** are the L2-coordination follow-up the cores' own docs scope as HW-gated. This PR delivers: cores declared + used, each node's L1 chain live, the crown's anchor/STH produced + self-verified + published via DIAG, and the NVS key provisioning. Diff scanned: no key/seed material, `beginagain` = 0.

Refs #181 #184.

### DIFF
--- a/rust/clock/Cargo.toml
+++ b/rust/clock/Cargo.toml
@@ -176,6 +176,12 @@ wifi = [
 # Builds on `wifi` (shares the radio init path) and turns on esp-wifi/esp-now.
 espnow = ["wifi", "esp-wifi/esp-now", "dep:ed25519-compact"]
 
+# #181/#184 mesh-ledger key PROVISIONING (a JP-supervised USB-touch build, NOT for the fleet OTA
+# image). Enables the one-shot NVS seeding of the ed25519 signing key from a build-time
+# `SMOL_LEDGER_SEED` (hex, supplied out-of-band, NEVER committed). The default / OTA image omits this
+# feature → it only ever READS an already-provisioned key. NO key generation in the firmware or CI.
+ledger-provision = ["espnow"]
+
 # #25 WLED WiZmote-emit: smol impersonates a WLED "linked remote" (13-B WiZmote
 # ESP-NOW broadcast). Needs the ESP-NOW broadcast path + RadioManager, so it builds
 # on `espnow`. Nothing new links — pure encode + the existing `esp_now.send`. The
--- a/rust/clock/src/net.rs
+++ b/rust/clock/src/net.rs
@@ -76,6 +76,26 @@ pub mod flood;
 #[cfg(feature = "espnow")]
 pub mod wire;
 
+// #181 mesh-ledger cores — the PURE, host-tested L1/L2/L3 primitives (sha256 + ed25519 injected),
+// wired into the firmware by `ledger_link` (#182 hash-chain, #183 CT-Merkle anchor, #184 signed
+// tree-head; landed inert via PRs #220/#223/#224). Each is a frozen library-of-primitives with a
+// full host-test suite (`experiments/{ledger,treehead,sth}_verify`); `ledger_link` wires a
+// meaningful SUBSET now (own-chain append + crown anchor/sign + verify-what-you-sign self-check),
+// with peer-STH gossip/acceptance the HW-gated L2-coordination follow-up. `#[allow(dead_code)]` on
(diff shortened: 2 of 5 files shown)


jphein:
Landed on main via 319e896 (rebased as fix/266-core-rebase @ 253f538 — this PR's head branch rig/v346-wave was the pre-rebase one). Region 114,616 B, margin +38,061 against the highest peak on record; #339's shed-order declaration extended for the ledger fields; CI green.

Issue #227, opened 2026-07-20
feat: weather-on-glass — gateway Open-Meteo fetch during WiFi window → ESP-NOW relay → fleet weather screen


jphein:
From the esp32c6-watch survey wave (#48/#50 diff; full writeup: `scratch/watch-backport/luna-notify.md`).

## The idea
The watch fetches current weather from Open-Meteo during its WiFi window and renders it on-glass.
This maps **1:1** onto smol's existing pattern: the gateway already **fetches during its WiFi window**
(OTA images) and **relays derived data over ESP-NOW** to the fleet (batt display, units). Add a weather
fetch + a tiny relay frame → the whole OLED fleet gets a weather screen. **Zero new dependencies.**

## Reference implementation
`esp32c6-watch/src/net/weather.rs` (185 lines, JP's laptop project — read-only reference):
- One-shot **HTTP/1.0** GET to `api.open-meteo.com:80` (no TLS — same constraint smol already lives with for OTA fetches). `Connection: close` ⇒ "read-until-EOF" framing, no chunked/keepalive.
- DNS via embassy-net with a **baked fallback IP** so a broken resolver doesn't kill the feature.
- **No-serde** JSON scrape: find `"temperature_2m":` / `"weather_code":` numeric values (skips the string units), parse digits, round. ~no allocation.
- Fire-and-forget, 8s-bounded; failure logs and returns `None`, never blocks boot.
- Output: `temp_f: i16` + WMO `code: u8`.

## smol shape
1. Gateway fetches weather during the WiFi window (right after the OTA/telemetry burst), same as `weather.rs`.
2. Relay a **~3-byte** weather record (temp° + WMO code) over ESP-NOW as a CFG/relay frame — trivially under the ESP-NOW frame budget and unrelated to the ~490 B gateway-MQTT cap (this is HTTP, not MQTT).
3. A `Weather` screen/plugin renders temp + a WMO-code glyph (clear/cloud/rain/snow/storm) on the 72×40 OLD.
4. Optionally publish to HA too (`smol/weather` or per the discovery pattern) for a dashboard tile.

## Notes / caveats
- Plain HTTP (no TLS) to Open-Meteo — acceptable (same as smol's existing image fetches); the DNS+fallback-IP trick is worth adopting for smol's fetch/OTA paths generally (separate hygiene issue).
- Location is baked in `weather.rs` (Seattle-ish lat/long) — smol would make it a build/config const.
- WMO code → glyph table is small and static.

[...]

Issue #128, opened 2026-07-14
tools: canary --build pins pollute the fleet's monotonicity gate


jphein:
Bench lesson (2026-07-14): canary staging with `--build 300/320/330` left id8 numerically AHEAD of the honest commit count (254), so honest-numbered stages read as NotNewer and the #120 cleanup (correctly) clears their orders — the board silently refuses real updates until main's count exceeds the pin (~76 commits) or a USB reflash resets it. Options: (a) ota_publish.sh warns/refuses `--build` > count+small-delta outside a --canary flag; (b) the stage tool ratchets: build = max(count, retained-staged+1) so the fleet number heals forward automatically; (c) canary pins use a SEPARATE low range... (b) is probably right. Also document: converging a pinned-ahead board needs one more pinned stage (done today: 331 for id8), then numbering heals at the next USB access.

jphein:
Shipped in #129: stage build = max(commit count, retained staged+1) with the heal-forward note; canary --build pins warn loudly with the heal path; env resolution now worktree-robust (also fixed the latent install-path gap). Live-accepted against the real broker: 331→332 chosen over count 258.

PR #266, opened 2026-07-20
release: v346 — #190 HMAC + #181 ledger + #227 weather

jphein:
## v346 wave (sigil: **Gear**)
Three held PRs integrated onto main via **cherry-pick of the focused feature deltas** (the branch tips were ~30 commits stale — a raw merge would have reverted shipped v345/#237 work). Zero conflicts; purely additive (2044+/17-).

Commits:
- `#190` group-HMAC-SHA256 transport auth (516cf60 + the 34b38e2 OTA-family exclusion)
- `#181` wire L1/L2/L3 mesh-ledger cores into the fw (ced53c6)
- `#227` weather-on-glass + DNS-with-fallback (a213fda)
- `release: v346` version.txt → 346
- `fix(build): drop -Zbuild-std` — see below

## ⚠️ Depends on #265 (merge that FIRST)
This branch carries the same `fix(build): drop -Zbuild-std` commit as standalone PR **#265**. Merge #265 first (un-reds main); this branch then reconciles cleanly. Without it, no fresh fw build compiles (portable-atomic host feature-leak).

## Invariants verified (code + host tests)
- **(a)** `should_group_mac` excludes the OTA family (OTAM/OTAD/OTAN/LDBG bypass the MAC trailer) — mac_verify OTA-family-exemption ✅
- **(b)** ledger ed25519 seed read from NVS (`resolve_ledger_key`), NO keygen, env-seed provisioning only ✅
- **(c)** mixed-fleet additive-ignore (old v345 boards ignore the 9B trailer) — by construction + relay_compat ✅

## Gate — cold, all green
clippy espnow,cast,io -Dwarnings · build --release {espnow,cast,io | default | wifi} · host verifiers mac/relay_compat/ledger/sth/treehead/wx/dns — all ✅

## Provisioning (git-ignored, per public-repo rule)
- `secrets.rs`: GROUP_KEY (canonical copy in vault `smol-group-key-hmac-v346`, epoch=1, observe-only)
- `board.rs`: WEATHER_LAT/LON (Seattle placeholder — set real coords before fleet roll) + WEATHER_FALLBACK_IP (188.40.99.226)

## Release artifact
- ELF (SMOL_RELEASE=1, build 346, hash fbfdee4): `~/smol-v346-Gear-clock.elf` on katana — sha256 `432ae336…`, 3,214,660 B, clean stamp (no +dev)

**DO NOT auto-merge** — team-lead runs the first peer-sourced canary roll.


jphein:
**Pre-rebase review (read-only, source+arithmetic — nothing built or measured). Verdict: REBASE-WITH-DECOMPOSITION, do not land as one.** The rebase itself is easy — only 6 real conflicts in 3 files (`wire.rs` and `ota.rs` land clean; the brief's collision list overstated it). The blockers are semantic.

**BLOCKER 1 — the DIAG record overflows and the compile-time gate cannot see it.** main's bound is `DIAG_BUDGET=495`, `DIAG_CORE_MAX=444`, proven margin 51 B. This PR appends `mo=/mf=` (28 B) + `lgt=/lgn=/lgok=/lgk=` (33 B) + crown-only `lgan=/lgep=/lgsz=/lgsg=` (53 B) via bare `push_str`, bypassing the `room_for` shed guard: **leaf 505 > 495, crown 558 > 495.** `encode_publish` then returns None and the record is never published — "a healthy board looks dead." And because the PR predates `DIAG_CORE_MAX` and doesn't update it, **the assert still passes** — the guard designed to fail closed is blind to exactly this. Resolution is a decision, not a merge: route the new fields through `room_for` (and pick their shed priority), or add their widths to `DIAG_CORE_MAX` and let the build break. Sequence behind #306.

**BLOCKER 2 — it would REGRESS #324, which landed 4 commits ago.** `WX_FETCH_BUDGET = 8 s` runs on its own budget outside the flush deadline; main's `REELECT_SILENCE_MS` is *derived* as `RELAY_FLUSH_BUDGET(15) + 2×2 + 1 = 20 s`. **15 + 8 = 23 s > 20 s** ⇒ twice an hour the crown's HELLO silence crosses the gate and every leaf runs an unneeded ~14 s recovery burst — the precise symptom #324 exists to remove, and we verified its absence on hardware today (4/4 protected-band silences absorbed). Fix by folding `WX_FETCH_BUDGET` into the derivation (→28 s, still < RECOVERY_STALE_MS 35 s) or emitting a HELLO from the fetch's tick. Credit where due: the fetch is properly bounded and tick-abortable; only its window sits outside the budget the gate derives from.

**BLOCKER 3 — semantic revert of main's leaf-DIAG fix.** The PR's `min(OBS_VALUE_MAX)` silent mid-value truncation replaces main's field-boundary cut + on-wire `cut=` tag (added because 75 B were vanishing with no signal). Taking "theirs" silently undoes it. Correct resolution: keep main's boundary-cut, set `cap = OBS_VALUE_MAX` (226), re-derive `room = 226 − CUT_TAG_MAX(8) = 218` — and state the real cost: #190's 6-byte trailer makes it ~89 B dropped instead of 75.

**Also found:** `.cargo/config.toml` is a FALSE CLEAN (both sides remove build-std; main carries an 11-line rationale the PR doesn't — read it by eye post-rebase). `Cargo.toml` adds feature `ledger-provision` ⇒ the tier matrix needs a 6th entry or that path is never compiled. `net.rs`: main declares NO ledger modules — `ledger.rs/sth.rs/treehead.rs/crdt.rs` are files outside the module tree, i.e. never compiled; the PR's block is what makes them real code. **Drop the version.txt 345→346 bump** — `build.rs` takes env first and #128's ratchet always passes it, so it can't change a staged sha; it only moves an offline default that's already stale by ~570.

**Two corrections to my own priors, for the record:** the flash-4-byte-alignment worry is REFUTED (the group key is a compile-time const in git-ignored secrets.rs — no NVS, no flash write). And the ledger is **stack, not .bss** (~2.6 KB in `RadioManager`, returned by value) — so `repro_build.sh`'s region gate will pass and *not see it*, while the floor's own derivation (4/3 × measured high-water) breaks. **A stack-paint re-measure is a merge precondition**, per that script's own instruction.

**Plan:** land steps 1-6 (the #190/#181 core — clean and testable) as one PR; split the DIAG-field decision and the #227 weather/timing reconciliation into a second, sequenced behind #306. #227 shouldn't ride the same merge as #190/#181 — it's the only piece touching the election-timing invariant. Full ordered rebase plan in the review notes.

jphein:
**RE-MEASURED AGAINST A BARD-FREE MAIN — it clears by 38,061 B, not 48. Recommend MERGE.**

`fix/266-core-rebase` rebased onto current `main` (`8b74dd8`) → **`253f538`**, force-pushed. **CI green** (run `30720206669`): shed order, all six check tiers, all five clippy tiers, host suites, stack floor.

### The number

| | |
|---|---:|
| region, rebased branch (local `repro_build_bin`) | **114,616 B** |
| region, same build in CI | 114,608 B |
| `main` today | 114,648 B |
| **cost of #266 to the region** | **32 B** |

That 32 B is exactly what `repro_build.sh`'s own comment predicts for this branch — the region gate is nearly blind to `LedgerLink` — so the branch behaves as characterised.

### The floor, re-derived with the ledger's runtime stack included

**`size_of::<LedgerLink>()` = 1,760 B, re-derived on this rebase**, not carried forward: a `const [(); 0] = [(); size_of::<LedgerLink>()]` probe compiled for `riscv32imc` reports 1,760 exactly. It is a by-value field of the stack-resident manager, so it is real stack the region cannot see.

Required = (peak + 1,760) × 4/3, against the 114,616 B region:

| peak used | required | margin |
|---|---:|---:|
| 54,856 B (T13 — the floor's own derivation, and the basis of the 48 B hold) | 75,488 | **+39,128** |
| 55,440 B (#302, main) | 76,267 | **+38,349** |
| 55,656 B (#335, spike — the highest on record) | 76,555 | **+38,061** |

### Why the #190 hot-path frames no longer decide it

The previous hold turned on them, so rather than estimate: the region permits a runtime peak of **85,962 B**. Subtract the highest recorded peak and the ledger and **#190's `send_to` MTU buffer plus HMAC/sha2 verify state would have to exceed 28,546 B of stack** to reach the floor. A 250 B buffer and a sha2 state are two orders of magnitude below that. **The verdict is insensitive to the figure that previously decided it** — which is a better outcome than a more precise estimate.

### The hold was right, and more right than it looked

Against the bard-carrying 75,536 B region, the same arithmetic on the *later* recorded peaks:

| peak | margin |
|---|---:|
| 54,856 | +48 B |
| 55,440 | **−731 B** |
| 55,656 | **−1,019 B** |

The 48 B rested on the oldest and lowest peak on record. On the two measured since, it was already negative. Holding was correct.

### Rebase notes

One conflict, in `net/mode.rs`: `main`'s #339 machine-checked `SHED-ORDER:` declaration versus this branch's older prose correction. Kept main's, and **extended the declaration to `sog, cfgq, cdeaf, cc, ap, cfg, io, deaf, lg_core, lg`** — #181 appends the two ledger fields after `deaf`, so without that the #339 checker fails, correctly. `tools/check_shed_order.py` now passes on the branch and in CI.

`tools/repro_build.sh` merged clean and is byte-identical to main's, so the branch inherits the bard-free canonical list. `ledger-provision` is present here, so the gate runs that tier instead of skipping it, and `mac_verify` (this branch's #190 host verifier) passes.

Caveat kept from the spike work: CI's number is 8 B below the local one because the provisioned `board.rs`/`secrets.rs` differ. Immaterial at 38 KB; it would have mattered at 48 B.

jphein:
Landed on main via 319e896 (rebased as fix/266-core-rebase @ 253f538 — this PR's head branch rig/v346-wave was the pre-rebase one). Region 114,616 B, margin +38,061 against the highest peak on record; #339's shed-order declaration extended for the ledger fields; CI green.

Issue #53, opened 2026-07-10
enh: battery display shows staleness instead of silently blanking to '--' when a source feed goes quiet


jphein:
**Tracks JP's report (2026-07-10): HV battery SOC showed just dashes on the OLED.**

**Root cause was NOT smol** — the Battery-Emulator stopped publishing HV-pack data to HA (`sensor.be_soc` silent 29h, `be_battery_voltage` silent >30min). The smol publisher (`smol_publish_batt_display`) *correctly* blanks stale segments to `--` per its freshness windows (HV SOC 6h, voltages 30min). It **self-healed** when the BE feed resumed (`HV 100%` came back on its own).

**Optional smol-side UX improvement (low priority):** a blank `--` is ambiguous — 'never had data' vs 'data went stale'. Consider a staleness hint instead of silently blanking: e.g. dim/marker + age (`HV 100%·2h`), or hold last-known-with-age (same idea as the #47 hold-last-known discussion). Current blank-on-stale is defensible; this just disambiguates. Not a bug — an enhancement.

jphein:
Closing per JP (2026-07-11): "i don't care about that battery staleness issue." The '--' blanking behavior stands; staleness display can be revisited via #74's config-drift/staleness work if it ever matters.

PR #272, opened 2026-07-20
fix(ota): #267 cross-burst OTA-fetch resume — resume the Range, not restart from 0

### LINKED ISSUE #267: fw(OTA): fetch restarts from byte 0 on retry — no resume-from-offset (Range)
**Observed** during the v346 co-channel test (2026-07-20): an OTA fetch that reaches a partial offset then fails **restarts from byte 0** on the next retry instead of resuming from the last offset.

Wire trace (id5, `coexist-soak` co-channel, `relayfetch`):
`progress 0` → `49152|1074640` (48KB pulled) → **`0|1074640`** (reset) → retry counter climbing `1→…→9`, each attempt back at byte 0. HWM = 49152 B (~4.6%); it never accumulates past one window.

Because the disease knocks the fetch out around ~48KB (co-channel duty-cycle) and every retry re-downloads from 0, the transfer can never converge on a 1MB image — even once the channel issue (#217) is addressed, a lossy link would still never complete without resume.

**Expected:** retries issue a `Range:` request from the last persisted offset (resumable fetch), so progress is monotonic across retries.

**Caveat:** observed in the lone-crown `relayfetch` path (crown alone on ch1, no relay peer) — a quirky config. Needs confirmation on a clean co-channel *self*-fetch before treating as a definitive resumability bug, but the restart-from-0 pattern is clear. Relates to #195 (failure cap), #188 (progress telemetry), #217 (channel blocker in front of it).

jphein:
## What
Fixes **#267** (co-critical, my root-cause): the OTA fetch restarts from byte 0 on each retry instead of Range-resuming, so under the coexist bulk-RX disease (≈1 chunk/burst) a multi-chunk image **never accumulates**. Without this, #217's own cross-burst retry also restarts from 0 — no OTA completes.

## Root cause
The HTTP Range-chunk resume worked chunk-to-chunk *within* one fetch burst (`off = writer.written()`), but `ImageWriter::begin()` reset `written` **and** the streaming SHA to 0 every call. The #195 OTA retry is per-burst within one boot (the #100 in-RAM idiom), so each burst re-invoked `begin()` → re-`Range`d from byte 0. `OTA_CHUNK = 49152` = the observed offset; retry 1→9 = 9 bursts each `0→49152→die`.

## Fix
- **Cross-burst resume cursor in a `.bss` static** — SAME-BOOT (resets on reboot, the #100 idiom). **No NVS** → sidesteps the near-full `nvs` partition (the retry never crosses a reboot). Keyed to `(build, sha16, slot)` so a re-stage / slot flip forces a fresh fetch.
- **`ImageWriter::begin(build, sha)`** resumes `written=flushed=committed`, `erased_upto` past the committed prefix (never re-erases it); each sector flush persists the on-flash offset; `finalize` clears the cursor.
- **Integrity → finalize READBACK-SHA** over `slot[..size]` (ports `LeafImageWriter`'s pattern) — burst-count invariant, so resume needs only the offset (no streaming hasher to restore, no per-burst re-hash).
- **Pure keying** in `net/ota_resume.rs` (host-tested, the flood/etx/ledger pattern).

## Verification (on familiar)
- **Host test** `experiments/267_resume_verify` — `cargo test` (1 passed) **and** `cargo run` GREEN: keying (build/sha/slot mismatch → 0) + segmentation-invariance (one-shot vs 3-segment vs **die-every-chunk** → byte-identical → identical finalize SHA).
- **`cargo clippy -D warnings` + release build** (`espnow,cast,io`, the canonical fleet image) GREEN.

## Coordination / notes
- The **one wifi.rs touch** is the `begin()` call site — coordinate with #217 / morpheus per the sequencing plan. This base already carries #217's fetch-leg (`OTA_BODY_STALL`), so it composes directly (no rebase conflict expected).
- ⚠️ **Build-env flags for the docs:** (1) familiar `/tmp` is a **512 M tmpfs** — the esp-hal build overflows it; use a home-disk `CARGO_TARGET_DIR` (+ `TMPDIR`). (2) JP's `board.rs`/`secrets.rs` carry **unconsumed WIP consts** (#227 `WEATHER_*`, #190 `GROUP_KEY_*`) that fail `-D warnings` in this tree (present in origin/main too — a pre-existing config skew, not #267); I isolated them in the scratch build to gate #267.

Fixes #267. Refs #217, #195, #204, #53.

### DIFF
--- a/experiments/267_resume_verify/Cargo.toml
+++ b/experiments/267_resume_verify/Cargo.toml
@@ -1,0 +1,19 @@
+# #267 host-test harness for the PURE cross-burst OTA-resume key logic (net/ota_resume.rs).
+# Runs OFF-target (std) via `cargo run` AND `cargo test`; #[path]-includes the real
+# ota_resume.rs (no drift). Proves the resume keying (match → offset, any mismatch → 0) and the
+# segmentation-invariance property the fix rests on: an image written in one shot vs resumed in
+# N segments (boundaries driven by the real `resume_offset`) yields byte-identical flash → an
+# identical finalize readback-SHA. Mirrors etx_verify/ledger_verify/crdt_verify. sha2 supplies
+# the readback hash — the core is dependency-free.
+[package]
+name = "resume_verify"
+version = "0.0.0"
+edition = "2021"
+publish = false
+[[bin]]
+name = "resume_verify"
+path = "src/main.rs"
+[dependencies]
+sha2 = "0.10"
+[profile.dev]
+opt-level = 0
--- a/experiments/267_resume_verify/src/main.rs
+++ b/experiments/267_resume_verify/src/main.rs
@@ -1,0 +1,136 @@
+//! Host verification of the PURE #267 cross-burst OTA-resume key logic (`net/ota_resume.rs`),
+//! included verbatim (#[path], no drift). Run: `cargo run` — panics on failure. `cargo test` runs
+//! the same suite (no false-green).
+//!
+//! Two things must hold for #267 to be correct:
+//!  1. **Keying** — a resume cursor is honored ONLY for the exact staged image + slot it belongs
+//!     to (build + sha16 + slot); any mismatch (re-stage, slot flip, different image) → fetch from
+//!     byte 0. This is the guard against resuming onto a stale/wrong flash prefix.
+//!  2. **Segmentation-invariance** — an image written in one shot vs resumed across N bursts (each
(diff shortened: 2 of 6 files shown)


jphein:
Superseded by merged PR #275 (a209858, 2026-07-21): ota_resume.rs at this PR's head is byte-identical to main's, and every symbol it adds is present on main. 2026-08-01 sweep.

Issue #300, opened 2026-07-26
feat: The Bard — on-device tiny-LLM storyteller (stories260K on the C3)


jphein:
Press the button and the board writes you a story it has never written before — a real transformer LLM composing TinyStories prose entirely on the $3 C3, typewriter-style onto the 72×40 OLED. No WiFi, no cloud: a leaf alone in a drawer can tell a tale.

**Inspired by** [slvDev/esp32-ai](https://github.com/slvDev/esp32-ai) (28.9M params on an S3 with 8MB PSRAM). smol's version rides the same llama2.c/TinyStories lineage at the scale our single-core, no-PSRAM, no-FPU part can execute: **stories260K, int8, executed-in-place from flash** (~280KB `.rodata`, zero RAM for weights; ~101KB static RAM for the int8 KV cache + buffers; image ~590KB → ~880KB, still <half an OTA slot).

**Design spec (approved 2026-07-26):** `docs/superpowers/specs/2026-07-26-tiny-llm-story-design.md`

v1 scope:
- [x] `tools/bard_export_model.sh` — fetch + int8-quantize stories260K → committed `.bin` (CRC32, pinned HF revision)
- [x] `src/bard/nano_llm.rs` — no_std llama2.c port (int8 matmuls, softfloat norms/softmax, GQA head-at-a-time)
- [x] Golden-token host test: Rust == reference llama2.c at temp 0, token for token, before anything flashes
- [x] Polled tick-budgeted generation state machine (single core shared with mesh/Familiar — never starves the fleet)
- [x] Bard app: menu entry, per-node protagonist seeds (id7 → "a little dragon"), typewriter UX, ~fin~
- [x] On-target tok/s bench → recorded as spec amendment
- [x] Canary-one-board OTA, then staged roll

v2 (out of scope, named): custom ~1–2M-param model trained on katana (weights-only swap), HA/CFG story trigger, idle dream mode, persona CFG override.

jphein:
v2 idea from the bench (JP, during the T13 première): **tap-to-continue a finished story** — extend rather than restart. Needs KV-cache context-shift (evict the story's opening, keep recent context) since stories stop exactly at cache-full; pairs naturally with the sliding-window work any longer-context upgrade needs. Filed alongside the existing v2 list (custom model, HA trigger, dream mode, persona CFG).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

jphein:
[...]

Issue #140, opened 2026-07-15
WiFi throughput/stability: tune esp-wifi RX buffers (heap-gated) + free mesh RAM before OTA


jphein:
## Problem

We run **stock esp-wifi 0.15.0 defaults** — verified: `esp_wifi::init()` takes no Config and `rust/clock/.cargo/config.toml` has no `[env]` block, so `rx_queue_size:5`, `tx_queue_size:3`, `static_rx_buf_num:10`, `dynamic_rx_buf_num:32`. These are **low for sustained RX** (IDF throughput/iperf profiles use static 16–25 / dynamic 64 + a matched `rx_ba_win`), and under-buffering is a contributor to the mid-download stalls in #138.

**The tension:** heap low-watermark is **~5.9 KB** during gateway duty. Each **static RX buffer ≈ 1.6 KB**, and dynamic RX buffers draw from that **same heap** on demand. Naively cranking buffers on a 5.9 KB-heap gateway can *worsen* stalls (exhaustion → allocation failures → dropped frames → dropped TCP — the esp-hosted #184 failure shape). So buffer tuning must be **gated on first creating heap headroom**.

## Fix (ordered — do NOT skip the gate)

1. **Instrument free-heap at `run_ota_fetch` entry** (DIAG/serial) to get a real number.
2. The OTA fetch is **mesh-deaf** for its whole window → **release the ESP-NOW peer tables / mesh caches before the download** to reclaim heap.
3. **Then** add an `[env]` block to `rust/clock/.cargo/config.toml` with the esp-config keys (prefix `ESP_WIFI_CONFIG_*` — **verify the exact key names against the esp-wifi 0.15 `esp_config` schema before applying**): raise `RX_QUEUE_SIZE`, `STATIC_RX_BUF_NUM`, `DYNAMIC_RX_BUF_NUM`, and set `RX_BA_WIN ≤ STATIC_RX_BUF_NUM` (Espressif: keep static_rx_buf_num ≥ rx_ba_win when AMPDU RX is on).
4. **Optional diagnostic only:** a canary build with `AMPDU_RX_ENABLE=false` to test whether the -78 dBm board's mid-download deaths are AMPDU-RX incompatibility (Espressif's "disable AMPDU RX to test" guidance). Revert after — AMPDU-off cuts throughput; it is not a keep-fix.

## Acceptance

- Free-heap at fetch entry ≥ a defined safe margin **after** step 2.
[...]

Issue #44, opened 2026-07-10
fw/build: release build not hash-reproducible — can't verify image<->board before flash (compounded the dup-NODE_ID outage)


jphein:
Found by lucid during the #42 / duplicate-NODE_ID recovery.

**Problem:** the release ELF is not hash-reproducible (stripped + non-deterministic codegen), so a flashed image can't be verified against its source commit or intended board by hash. This compounded the id8/id9 outage — the wrong image (default NODE_ID=7 flashed to both leaves) couldn't be caught by an image<->board hash check.

**Suggested fix:** a reproducible build profile for fleet images (e.g. `codegen-units=1` + deterministic flags) so image<->commit<->board can be hash-verified before/after a flash.

**Related:** the dup-NODE_ID incident; #42 (SMOL_NODE_ID build guard, merged 17c0686). Complements #42 — #42 prevents building the wrong id; this enables *catching* a wrong-image flash.

Priority: medium (build-tooling hardening, not blocking).

jphein:
**Design delivered (parallel design agent, read-only — full doc in scratchpad/issue-44/).**

**Confirms the root of the morning outage:** a built image's NODE_ID is **not extractable today** — it's a const-folded immediate and the full name table ships regardless of id, so nothing distinguishes an id-8 image from an id-9 one before flash. (BUILD_HASH *is* greppable; byte-reproducibility is blocked by `esp_app_desc!`'s wall-clock timestamp → don't chase byte-identity, it's the wrong goal.)

**Approach — 2 layers:**
- **Layer 1 (fw): embed a greppable identity stamp.** Preferred: ESP-IDF app-desc `version` = `n<node>|v<build>|<hash>` (fixed offset, standard-tool readable). Fallback: build.rs emits `SMOL_IDENT` + a `#[used] static` in the git-ignored board.rs (keeps mesh hot-files untouched).
- **Layer 2 (ships today, zero fw):** `ota_publish.sh` appends `sha256⇥node⇥build⇥hash⇥file` to `tools/image-manifest.tsv`; verify matches image-sha → node.
- **verify_image.sh:** pre-flash guard AFTER the MAC-guard, BEFORE espflash — SMOLIDENT stamp → manifest → fail-closed. MAC-guard proves the BOARD; this proves the IMAGE; both must agree.

[...]

PR #301, opened 2026-07-27
feat: The Bard — on-device tiny-LLM storyteller (#300)

jphein:
Press BOOT and the board writes you a story it has never written before — a real transformer (llama2.c-lineage **stories260K**, int8, executed-in-place from flash) composing TinyStories prose typewriter-style onto the 72×40 OLED. Fully on-device: no WiFi, no cloud; a leaf alone in a drawer can tell a tale. Part of #300 (bench + canary + fleet roll follow after merge — do not auto-close).

**Design spec:** `docs/superpowers/specs/2026-07-26-tiny-llm-story-design.md` · **Plan (as-built log):** `docs/superpowers/plans/2026-07-26-bard-tiny-llm.md` · **Inspiration:** [slvDev/esp32-ai](https://github.com/slvDev/esp32-ai) (28.9M params on an S3 with 8MB PSRAM — smol's version is the same idea at the scale a single-core, no-PSRAM, no-FPU C3 can execute)

## Measured (canonical image, `espnow,cast,io,bard`)
| | value |
|---|---|
| Model blob (`.rodata`, XIP — zero RAM) | 283,096 B committed, CRC32-tailed, sha `f0e50c5e…74e2d16` |
| `.bss` cost | +66,880 B (`Bufs` 86KB-class at `SEQ_CAP=160` + statics) |
| **Runtime stack** | **14,240 B**, `__stack_chk_guard` inside the region; **build hard-fails < 12,288 B** (new readelf gate in `repro_build.sh`) |
| Stories | ~145 generated tokens ≈ 300 chars, temp 0.9 / top-p 0.9, per-node protagonist (id8 → "a little owl") |
| Host throughput ceiling | ~1,800 tok/s (x86 release; C3 measured at T13) |
| Host tests | 15/15, incl. **bit-for-bit golden vs an independent Python reference** (146/146 ids, full text) |

## Why reviewers can trust the math
The forward pass agrees **token-for-token with two independent implementations** (an int8-exact numpy reference mirroring the accumulation order, and a from-scratch fp32 reference written during review). Upstream `runq.c` was disqualified as a reference — it drops in-dim elements 128..171 of every w2 row when `n % GS != 0` (llama2.c @ `350e04fe`, lines quoted in the spec) — so the golden gate is cross-implementation, not cross-port.

## The stack saga (why the new build gate exists)
Bard's `.bss` initially consumed the runtime stack silently — the linker shrinks `.stack` to nothing before erroring, so the image "linked clean" with 2,592 B of stack and the stack-protector guard *outside* its region. Caught in review by section-table arithmetic, fixed by `SEQ_CAP 192→160`, and permanently gated: `repro_build.sh` now derives `_stack_start − _stack_end` and aborts below 12 KiB (negative-tested against the real thin ELF). A boot-time stack-paint high-water measurement on the bench board gates the fleet roll.

## Flagged for JP
- **#44 repro-lineage fork:** the canonical fleet image feature list is now `espnow,cast,io,bard` (`tools/repro_build.sh`) — image shas change lineage from this merge.
- **Pre-existing, NOT this branch:** clippy `-D warnings` fails on `espnow`/`wifi`/`fleet` tiers over 5 unused constants in the gitignored `board.rs`/`secrets.rs` (identical without bard; zero diagnostics in any bard file on any tier).
- **Follow-up lever (untouched):** the esp-wifi 128 KiB heap (#140; low-watermark ~52 KB free) can buy `SEQ_CAP` back toward 256 for longer stories — a radio-tuning decision for its own branch.
- Possible upstream contribution: the `runq.c` `n % GS` finding may be worth filing against llama2.c.

Sampled stories (20-seed batch in `examples/bard_stories.rs`) are varied, grammatical, dialogue-punctuated toddler-whimsy; 1/20 ends naturally ("They became best friends. The end."), the rest trail off with a deliberate `...`.

## Post-merge plan
T13: flash bench board id8/Eldritch Nexus (MAC-allowlisted) — stack high-water paint, tok/s, on-glass story. T14: `ota_publish.sh` canary-one-board, then staged fleet roll. Numbers land as spec amendments.


jphein:
**Final whole-branch review (read-only oracle): APPROVE-FOR-PR.** 15/15 host tests re-run green; every measured claim in the PR body independently reproduced (section sizes, stack symbols, guard placement, golden lineage across both SEQ_CAP changes, log-string gating in both ELF variants).

Two notes for the record:

1. **Cache-edge convention quirk (by design, documenting so T13 doesn't rediscover it as an anomaly):** `Story::step` stops at `pos+1 >= SEQ_CAP` → **145** generated tokens on-device, while the golden test helper and the Python reference run to `pos = SEQ_CAP-1` → **146**. The golden therefore validates one token beyond what the shipped path generates; Story is strictly more conservative and never writes the last cache slot. Numerically harmless.
2. The T11 log line prints `tok count / avg ms/tok / max ms` — tok/s is derived, not emitted.

Post-verdict fix round landing as one follow-up commit: fail-closed stack-floor gate (unreadable symbols currently warn-and-pass), one stale pre-fix comment in `textflow.rs`, and test coverage for the golden test's numeric-divergence reporting arm.

Issue #318, opened 2026-07-28
ha_deploy.sh push has no scoping — one agent deploying their own change deploys everyone else's


jphein:
## Symptom

`tools/ha_deploy.sh push` is a **whole-directory converge with no scoping.** `cmd_push()` iterates
every file in `$(packages)` and pushes each one whose content differs from live. There is no
`push [file]` form — compare `diff [file]`, which does take an argument. So **any agent who deploys
their own package change also deploys every other agent's committed package changes, in the same
batch, without being told.**

That is correct and desirable for one operator: converge, skip unchanged, back up, pick the right
reload. It becomes something else entirely once several authors share `main`.

## What happened today

Two independent work streams touched `ha/packages/` within the same hour:

1. An id7/id9 retirement — **−1439 lines, 172 entities** across all four packages, committed as
   `89ce9ab` + `53f45bf`.
2. Freeze / coexist diag work in `smol_mesh.yaml`.

The second author ran `ha_deploy.sh push` to deploy their diag sensors. Because `push` converges the
whole directory, **the retirement went live in the same batch, at 09:23:50.** Nobody decided to
deploy it. The retirement's author (me) was at that moment deliberately *holding* the push, waiting
on a dashboard-side precondition, and discovered it was already live only by checksumming the VM
against a backup.

**The work itself was authorised and verified clean** — this is not a report of damage. It is a
report that **the deploy was not the deployer's decision to make, and they had no way to see they
were making it.** The team lead had explicitly reserved that push for the retirement's author; the
tool routed around that reservation silently.

## The part that should worry us

The two halves of that change had a **required order**: remove two dashboard cards *first*, then
push the packages. Reversed, JP sees two panels flip to *Entity not found* on the day the cleanup
shipped — the specific outcome the coupling was designed to prevent, because the override cards are
built from `input_*` helpers, which (unlike mqtt/template entities) really do vanish on
`reload_all`.

**That ordering came out right. It was not chosen.** It was determined by which operation the
unrelated deployer happened to run first. A sequencing decision that two agents and the lead
reasoned about explicitly was, in the end, settled by luck.

## Why it survived

Because the asymmetry is invisible until it bites. `diff` takes a file, so `push` reads as though it
would too. And the failure mode is *silence*: the batch succeeds, the backups are written, the
reload is correct, `status` goes green. Every signal says the deploy was fine — because it was fine.
What is missing is any signal about **whose changes were in it.**

[...]

Issue #324, opened 2026-08-01
bug(mesh): crown churn — coordinator hands off ~5× per 8-min OTA on a stationary fleet; the #136 "never take over a live owner" invariant appears violated


jphein:
## Observation

During a single ~8-minute OTA on 2026-07-28 the crown changed hands **five times**, read off `smol/mesh/channel` (payload `MC|<owner>|<channel>|<seq>`):

```
MC|5 → MC|8 → MC|51 → MC|8 → MC|5
```

Two earlier captures saw 3 changes each in ~9 min (`MC|5→MC|50→MC|8`, `MC|5→MC|8→MC|51`). Fleet is four boards (ids 5, 8, 50, 51), all build 913, all stationary, all on the same AP. Reproducible, not a one-off.

## Verified evidence: churn is concentrated under OTA load

From fixture `~/.smol-ota-fixtures/live-id50-913.log` (produced by `tools/ota_capture.sh` during the id50→913 OTA; capture header `2026-07-28T19:25:54Z`, 500 s window). All 21 `MC` records, with the OTA progress stream from the same file overlaid:

```
@0   r=1  MC|5|6|…685   (retained baseline at subscribe)
@3        MC|5|6|…686
@116      MC|8|6|…687   ← change 1 (5→8)   after 113 s of MC silence
@188      MC|51|6|…689  ← change 2 (8→51)  after  69 s
@228      MC|8|6|…691   ← change 3 (51→8)  after  38 s
@291      MC|5|6|…694   ← change 4 (8→5)   after  13 s
@319…@476 MC|5|6|…695…705   (11 republishes, no further change)

@16–@39    relayfetch 0 → 1,228,800 / 1,450,320 then STOPS  (attempt 1 aborts ~85%)
@140       ota/state in_progress:true
@146–@174  relayfetch 0 → 1,450,320   (attempt 2 completes)
@218–@244  relayfetch 0 → 1,450,320   (attempt 3 completes)
@248       ota/state in_progress:false
@248–@476  no progress records at all  (228 s of OTA-idle tail)
```

**All four changes fall in `@116–@291`.** Three are inside the OTA-active span; the fourth is 43 s after `in_progress:false`. **The last 185 s — transfer stopped, crown settled on id5 — shows zero changes** and a metronomic republish.

[...]

Issue #309, opened 2026-07-28
smol_mesh.yaml: ~2,900 lines of per-id duplication — the root cause of thin boxes on id50/51/122


jphein:
## Symptom
`ha/packages/smol_mesh.yaml` defines diag/telemetry entities **by hand, per node id**. luna traced the "thin boxes" on **id50 / id51 / id122** to exactly this: those ids have no hand-written stanzas, so the dashboard had nothing to draw.

The file is ~2,900 lines, and the duplication is the point — most of it is the same entity family repeated per id.

## Why it survived
It worked for as long as the fleet was static. It was written when the fleet was id7/8/9 and stayed correct until the fleet changed — at which point the failure is **silent and asymmetric**: existing ids keep working perfectly, new ids get nothing, and nobody gets an error. The file cannot know an id exists.

Compare #308 (override cards) — same root cause, different surface. **Any hand-maintained id list in `ha/` is a pending instance of this bug.**

## The smallest correct fix — and it is firmware-side, not more YAML
**Publish MQTT discovery configs for the diag fields from the firmware**, as is already done for the device block. Then HA learns the entities the same way it learns the devices, and the per-id YAML stops existing rather than growing.

Writing more hand-written stanzas for 50/51/122 would restore the dashboard **and re-arm the same bug** for the next re-provisioning — which is the trap `439fb95` retired for discovery and this file is the last big holdout of.

Rule: **a self-reported manifest beats a hand-maintained list** ([DOC-UPKEEP](docs/DOC-UPKEEP.md) §5).

## Sequencing note
Worth doing **after** #306 (the DIAG record is ~33 B from a silent-death cliff). Discovery configs are a different topic from the DIAG record and shouldn't add to it — but anyone touching diag-field plumbing should read #306 first so the fix doesn't push the record over.

jphein:
## Precondition measured on the live fleet — it does NOT hold (3 of 4 boards)

Fleet: ids 5 (crown, ch6) / 8 / 50 / 51 live on **v913 Nozzle**. Read from HA's **entity
registry** (`/api/states` cannot answer this — it does not expose `unique_id`).

The stated precondition was "entity_ids read `sensor.smol_<id>_heap_free` on a live board".
Only **id8** does:

| board | entity_id | unique_id | device? |
|---|---|---|---|
| id5 | `sensor.smol_5_obsidian_aegis_heap_free` — 27672, live | `smol_5_heap_free_mirror` | yes |
[...]

PR #330, opened 2026-08-01
feat(ha): #320 one fleet readback for delivery/tale, replacing the last per-id mirrors

### LINKED ISSUE #320: HA: one wildcard delivery/tale readback, replacing the last per-id mirrors — and it closes the partial fan-out gap
Filed rather than bolted onto the end of the fleet-shaping work, because the design has one subtle constraint that deserves its own review. Design is morpheus-yaml's, from the review of `04d8dad`/`d500efb`; the accumulation constraint is the part I'd want held to.

## The gap this closes

`smol_bard.yaml`'s fan-out now warns when it reaches **nobody** (`fleet | count == 0` → persistent notification, per control). That leaves a real hole:

**Nothing catches "reached 4 of 6."**

`fleet` is discovered at run time from `sensor.smol_<id>[_<slug>…]_status`, a firmware-published MQTT-discovery entity. If two nodes' `_status` entities are absent during discovery replay after an HA restart, `count == 4`, the `default` branch runs, it publishes to four, **and dismisses the notification**. The two missed nodes keep their old value — silently, and permanently, because nothing republishes `config/delivery` or `config/tale` on a timer (verified: the only `time_pattern` automations live in `smol_mesh.yaml` and touch neither topic).

That is the same permanence as the `parallel` ordering race fixed in `04d8dad`, arriving through a different door.

A readback is the only thing that can see it, because it reads **the retained topic** rather than what HA believes it published. It therefore also covers out-of-band writes — meshscope, the operator TUI, `mosquitto_pub`.

## Shape

One MQTT **trigger** sensor per topic family, not per-node mirrors — the same idiom as the `#153`/`#217` sensors:

- trigger on `smol/+/config/delivery` (and `smol/+/config/tale`), `trigger.topic` carrying the id — id-agnostic, no per-id stanza to grow
- **state = agreement, not a mirror**: `6/6 agree`, or `4/6 · id51,id122 stale`
- attributes hold the per-node map

## The constraint, which is the reason this is an issue and not a patch

The per-node map has to accumulate across triggers (each message carries one node), and **an unbounded accumulating attribute dict is the husk problem in attribute form** — a retired board's entry persisting forever, unreachable by any cleanup and invisible to `--check`'s `DEAD ROWS` pass. The id7/id9 retirement left 110 `restored: true` registry husks; this would quietly grow its own.

So the map **must be intersected with the discovered fleet on every render**, so a retired id drops out on the next message. Bounded by the live fleet it is fine; unbounded it is a reason to refuse the feature.

Implementation note for whoever takes it: `state` and `attributes` are rendered from the same trigger context and both see `this` as the state *before* the update, so both can derive the same merge independently and stay consistent — at the cost of the merge logic appearing twice. Worth checking whether that duplication can be avoided before committing to it.

## Scope

Should **replace** `sensor.smol_8_delivery` and `sensor.smol_8_tale`, not sit beside them. Two per-id mirrors out, one fleet sensor in: a net entity reduction, and it retires the last per-id holdout in a file that is otherwise now fleet-shaped.

## Priority

Low. The partial fan-out window is narrow (post-restart, before discovery replay) and re-applying the control fixes it. The out-of-band-write coverage is the broader long-term value.

Related: `04d8dad` (queued + empty fan-out guard), `d500efb` (per-control notification ids), `a71debf` (every payload field defaulted).

jphein:
Lands the stranded `feat/320-bard-readback` @ `b1e8dec` (1 commit, 2026-07-28, no PR until now).
Reviewed and verified against the LIVE fleet today; findings below.

## What it does

Replaces the last two per-id mirrors in `smol_bard.yaml` (`sensor.smol_8_delivery`,
`sensor.smol_8_tale`) with one wildcard readback per topic family, and closes the gap the fan-out
guard cannot see: **"reached 4 of 6."** The empty-fan-out control only warns at `count == 0`, so a
partial publish covers some nodes, dismisses the notification, and the missed nodes keep their old
value permanently — nothing republishes these topics on a timer. A readback catches it because it
reads what is RETAINED, not what HA believes it published, which also covers out-of-band writes
(meshscope, the operator TUI, `mosquitto_pub`).

Net entity change: **2 per-id mirrors out, 4 fleet entities in** (an accumulator + a judge per
family). The 2-for-1 split is documented in-file: `this` is unavailable inside an `action:` block,
so the merge cannot live there; putting it in both `state:` and `attributes:` would duplicate the
derivation. Accumulator state = last message, merge = one attribute, judge = a plain sensor.

## The refusal condition is satisfied — checked first, because it gates the feature

The issue says an unbounded accumulating attribute dict "is a reason to refuse the feature". It is
bounded:

```jinja
{{ dict((prev | combine({k: val})).items() | selectattr('0', 'in', fl) | list) }}
```

The map is intersected with the discovered fleet on **every** render, so a retired id drops out on
the next message from anybody. The in-file comment makes the right argument for why this matters —
`DEAD ROWS` looks for absent *entities*, and an attribute key is not one, so an unbounded map would
be a private, unauditable version of the id7/id9 husk problem.

## Verified on the live fleet (2026-08-01)

Fleet: ids 5 (crown, ch6) / 8 / 50 / 51 live on v913 Nozzle; 122 + 236 dormant.

- **Retained topics exist and the readback would be meaningful.** `smol/+/config/delivery` is
  retained on all six ids at `160:inf:5x8` → the judge would render **`6/6 agree`**. No partial
  fan-out at present, which is the expected healthy reading.
- **`smol/+/config/tale` has no retained messages at all** → `m | count == 0` → **`nothing
  reported`** against a discovered fleet of 6. Correct and honest, not a failure.
- **The fleet-discovery regex handles the real entity_ids.** This mattered more than it looks: the
  live `_status` entities are the noun-infixed forms — `sensor.smol_5_aegis_status`,
  `_50_ember_status`, `_51_sigil_status`, `_122_crown_status`, `_236_herald_status` — alongside the
  bare `sensor.smol_8_status`. The branch's `^sensor\.smol_\d+(_[a-z0-9]+)*_status$` matches all
  six. An `?` quantifier here would have missed most of the fleet, which is the exact bug fixed in
  `resolve_fw` (see the comment above `HEARTBEAT` in `build_control_room.py`), and it is why this is
  worth stating rather than assuming. Confirmed discovery → `['5','8','50','51','122','236']`.
- **YAML parses**; 4 template blocks, 2 trigger-based accumulators + 2 plain judges; the `mqtt:`
  key is gone with the mirrors.
- **Nothing else in the repo references the two deleted entities.** `grep` across `ha/`, `tools/`
  and the dashboard scaffold is clean — the only hits are a stale agent worktree and a
  `scratch/vm-backups/` copy. So the deletion creates no DEAD ROWS. (`input_text.smol_8_tale` is a
  separate control input and is untouched.)
- **No conflict with `main`:** `ha/packages/smol_bard.yaml` has not moved since the merge-base
  (`63ce915`), so this is a clean single-file change despite the branch being 41 commits behind.

Two HA gotchas the author probed and documented in-file, both of which failed *silently* — worth
keeping in the comments, since neither leaves a log line:

1. `this` is not defined inside `action:`; reading `this.attributes` there raises in the variables
   step and the entity never updates, state stuck at `unknown`.
2. `variables:` applies native-type coercion, so `nid` arrives as the **number** 50 while the
   discovered fleet is a list of **strings** — `nid in fleet` was False for every node and the map
   rendered `{}` forever while the state line looked perfectly correct. Both sides are now forced
   to `string`.

## Review notes (non-blocking)

- `fleet` counts dormant boards (122, 236). Their retained `config/delivery` is present, so today it
  reads `6/6 agree`. If a dormant board's retained topic were ever cleared, the judge would report
  it as `silent` — flagging an offline board rather than a missed publish. Defensible (the retained
  topic *is* the record of what a node was told, independent of liveness) but worth a deliberate
  decision if `silent` ever drives an alert.
- The judge compares nodes to each other rather than to a reconstructed "expected" value. The
  in-file rationale is right: reconstructing expected means a second copy of the publish
  automation's compose template, and that copy drifts.

## Deploy status — NOT deployed, on purpose

`ha_deploy.sh status` currently reports `smol_bard.yaml` **in sync** repo↔live, so merging this
diverges them until it is pushed. A package push triggers HA `reload_all`, and an OTA canary roll
(build 916) is in flight, so I am not doing that from this lane right now.

**Post-roll, to finish:** `./tools/ha_deploy.sh push smol_bard.yaml` (scoped, per #318), then confirm
`sensor.smol_bard_delivery_agreement` reads `6/6 agree` and that the two retired mirrors are gone.
Until then this is a repo-only change — merge it or hold it, but do not consider #320 closed until
that readback has been seen on glass.

Closes #320 once deployed and verified live.

### DIFF
--- a/ha/packages/smol_bard.yaml
+++ b/ha/packages/smol_bard.yaml
@@ -313,18 +313,201 @@ automation:
                     qos: 1
                     payload: "{{ trigger.to_state.state }}"
 
-mqtt:
-  sensor:
-    # #302 delivery readback: what the node was actually told, as `<ms>:<mode>`.
-    - name: "smol 8 delivery"
-      unique_id: smol_8_delivery_mirror
-      state_topic: "smol/8/config/delivery"
-      icon: "mdi:play-speed"
-      value_template: "{{ value if value else '(default)' }}"
-    # #303 Bard story-prompt mirror: read back the retained prompt so the card shows what was
-    # commanded. '(default)' when empty = the node is using its built-in persona.
-    - name: "smol 8 tale"
-      unique_id: smol_8_tale_mirror
-      state_topic: "smol/8/config/tale"
-      icon: "mdi:feather"
-      value_template: "{{ value if value else '(default)' }}"
+# =====================================================================
+# #320 FLEET READBACK — replacing the last per-id mirrors in this file
+# =====================================================================
+# REPLACES `sensor.smol_8_delivery` / `sensor.smol_8_tale`: two hand-written mirrors of
+# `smol/8/config/*`, the last per-id stanzas here after the controls went fleet-wide. They were
+# #308/#309 in miniature — id8 got a readback, id50 and id51 got nothing, and the file could not
+# notice a board existed.
+#
+# ADDS the thing that actually justifies it. The fan-out warns when it reaches NOBODY
+# (`fleet | count == 0`). Nothing caught "reached 4 of 6": if two nodes' discovery entities are
+# absent during replay after a restart, the loop covers four, DISMISSES the warning, and the two
+# missed nodes keep their old value — silently and permanently, because nothing republishes these
+# topics on a timer. Only a readback sees that, because it reads what is RETAINED rather than what
+# HA believes it published. It also catches out-of-band writes (meshscope, the operator TUI,
+# `mosquitto_pub`).
+#
+# TWO ENTITIES PER TOPIC, ACCUMULATE AND JUDGE, and the split is not decoration. The merge has to
+# be carried across triggers via `this`, and `this` exists ONLY in `state:`/`attributes:` — NOT in
+# an `action:` block (probed on this instance: `this is defined` is False there). Computing the
+# merge in both `state:` and `attributes:` was the alternative and it duplicates the derivation.
+# So the accumulator's state is the last message (no merge needed) and the merge lives in exactly
+# one attribute; a plain template sensor then judges agreement from it. One definition, and each
+# entity does one job.
+# =====================================================================
+template:
+  # --- DELIVERY (pace:mode:font): ACCUMULATE. One job — carry the per-node map. -------------------------
+  - trigger:
+      - platform: mqtt
+        topic: "smol/+/config/delivery"
+    action:
+      # `this` is NOT available inside an action: block — probed on this instance, it is
+      # `this is defined == False` here while being defined in state:/attributes:. The first
+      # version of this file computed the merge here from `this.attributes`, which raised inside
+      # the variables step and made the whole entity SILENTLY never update: no log line, state
+      # stuck at `unknown`. So the action carries only what the trigger itself knows, and the
+      # merge lives in `attributes:` where `this` exists.
+      - variables:
+          nid: "{{ trigger.topic.split('/')[1] }}"
+          # Retain-clear arrives as a zero-length payload: "back to the node's own default". A real state, not a missing one, so it is recorded rather than skipped.
+          val: "{{ trigger.payload if trigger.payload else '(default)' }}"
+          fleet: >-
+            {%- set ns = namespace(ids=[]) -%}
+            {%- for e in states.sensor | map(attribute='entity_id') | select('search','^sensor\\.smol_\\d+(_[a-z0-9]+)*_status$') -%}
+              {%- set id = e.split('.')[1].split('_')[1] -%}
+              {%- if id not in ns.ids -%}{%- set ns.ids = ns.ids + [id] -%}{%- endif -%}
+            {%- endfor -%}
+            {{ ns.ids }}
+    sensor:
+      - name: "smol bard delivery nodes"
+        unique_id: smol_bard_delivery_nodes
+        icon: "mdi:play-speed"
+        # State is the LAST message, which needs no merge — so the merge exists exactly once, in
+        # the attribute below. Deriving the same thing in both templates was the alternative, and
+        # a duplicated derivation drifts the moment either copy is touched.
+        state: "{{ nid }}={{ val }}"
+        attributes:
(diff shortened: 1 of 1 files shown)


jphein:
## ⏸ Deploy-gated — do not merge-and-forget

Holding this open for review after the OTA roll settles, per team-lead. Two windows must close
first, and they are the same windows that gate every other HA change from this pass:

1. **lucid's #324 idle capture must end.** Verifying this PR means confirming
   `sensor.smol_bard_delivery_agreement` reads `6/6 agree` on the live instance, which requires
   deploying the package — and deploying triggers HA `reload_all`.
2. **The roll must settle.** `sensor.smol_8_build` read 916 (canary) while the rest of the fleet was
   on 913 during review; a coherent fleet makes the readback's own numbers trustworthy.

**Merging this does NOT close #320.** The package is currently `in sync` repo↔live, so a merge
diverges them until pushed. After merge:

```bash
./tools/ha_deploy.sh push smol_bard.yaml          # scoped, per #318
curl -s -H "Authorization: Bearer $HA_TOKEN" \
  https://ha.jphe.in/api/states/sensor.smol_bard_delivery_agreement
```

Expect state **`6/6 agree`** (`smol/+/config/delivery` is retained on all six ids at `160:inf:5x8`),
`sensor.smol_bard_tale_agreement` = **`nothing reported`** (correct — no retained tale exists), and
`sensor.smol_8_delivery` / `sensor.smol_8_tale` gone. That state string is the evidence to close
#320 with.

Full runbook, including the dashboard write that is still outstanding for #309/#322:
`~/.claude/projects/-home-jp/scratch/issue-drain/luna-ha.md` → "RUNBOOK".

Everything in the review above stands unchanged — this note only marks the gate.

Issue #336, opened 2026-08-01
security: #190's group key has no zero-key guard — an unedited secrets.rs.example authenticates the fleet on a published key


jphein:
Found in the PR #266 pre-rebase review (2026-08-01, source review — not built or exploited).

`secrets.rs.example` ships `GROUP_KEY = [0u8; 32]`, and nothing in the #190 transport-auth implementation refuses an all-zero key (grepped the whole diff). A fleet built from an unedited example therefore authenticates every frame on a **publicly known key** — the repo is public, so the key is in the example file for anyone to read.

**The part that makes this worse than a footgun:** #190's own rollout gate is designed to flip from observe to enforce when `mac_ok` climbs and `mac_fail` falls to zero (design §7.1). With a zero key, both of those counters read exactly as they would for a correctly-keyed fleet. **The gate cannot distinguish 'keyed' from 'keyed with the published zero key'** — so the safety check green-lights the insecure state. Same shape as the other gates-that-cannot-fail found this session.

Suggested fix, one line and compile-time: `const _: () = assert!(GROUP_KEY != [0u8; 32]);` behind the `espnow` cfg, so a fleet build with the example key fails to compile rather than shipping. Pair it with a note in secrets.rs.example.

**Second, related decision (not a defect — a conscious call needed):** the key is baked into the image, so **a staged OTA artifact IS the fleet credential**. Images are served from a LAN host and their sha256 is published on a retained MQTT topic. That's probably acceptable for this threat model (physical/LAN adversary already has the boards), but it should be a decision on the record rather than an implicit consequence — especially before #190 flips to enforce.

Blocks nothing today (#190 is unmerged, in PR #266). File it now so the guard lands WITH the feature rather than after.

jphein:
Verified fixed on main by the 2026-08-01 backlog sweep (oracle, pinned to 4b71e29) — evidence cited in scratch/parity/backlog-verified-0801pm.md. Closing as part of the 63→37 drain.

PR #341, opened 2026-08-01
ci: #338 make the build-tier gate a gate, not prose

### LINKED ISSUE #338: ci: the build tier matrix is prose, not a gate — no workflow builds any tier
Found during the #266 core rebase (2026-08-01). `.github/workflows/` contains **only `pages.yml`**. Nothing in CI builds any feature tier, runs clippy, or runs the host suites — the multi-tier gate everyone cites is **prose in ONBOARDING.md** plus whoever remembers to run it locally.

Consequences already observed today, all in one session:
- **main was red on its own `clippy -D warnings` gate** and nobody knew (fixed in 4c5955f — a real `needless_range_loop` in tracked source, plus pre-existing dead-code findings).
- The #266 rebase added a new feature (`ledger-provision`); the brief said "add a 6th entry to the tier matrix" and there was **no matrix to add it to**. It was built by hand once; nothing makes that durable, so the next change silently stops compiling that path.
- Two agents independently produced "clippy delta = zero" by building an identical baseline by hand — correct method, but it should not require inventing a baseline each time.

This is the same family as the other gates found today (a determinism check that only ran in a mode the publish path never used; a dead-row audit that could not see the rows it existed to find; an exit code read as a summary). **A gate that lives in prose is a gate that has already stopped running.**

### Suggested minimum

A workflow that, on push/PR, runs:

- `cargo check --release` across the canonical tiers — `default` / `wifi` / `espnow` / `espnow,cast,io,bard`, plus any feature a PR adds
- `cargo clippy --release -- -D warnings` on the canonical tier
- the host `experiments/*_verify` suites

Cross-compilation needs the riscv target; the host suites need none.

**Stretch:** invoke `repro_build_bin` so the **stack-floor number** is printed on every PR. That gate is currently only enforced at package time, which is the latest possible moment to discover an image is unshippable — and today's #233 spike and #266 rebase both turned on exactly that number.

ONBOARDING.md should then point AT the workflow rather than describing the commands, so the two cannot drift.

jphein:
Closes #338.

`.github/workflows/` held only `pages.yml`. Nothing built a feature tier, ran clippy, or ran the host suites — the "4-tier gate" cited in `ONBOARDING.md` and `rust-toolchain.toml` was **prose plus whoever remembered**. In one session that cost us three things: `main` sat red on its own `-D warnings` rule unnoticed (fixed in 4c5955f), a new feature (`ledger-provision`) arrived with no matrix to add it to, and two agents independently hand-built baselines to compute a clippy delta.

## Shape

`tools/gate.sh` is the gate. CI calls that same script, so the local instructions and CI are one definition rather than two free to drift. The workflow only supplies a machine.

| arm | covers |
|---|---|
| `cargo check` | default · wifi · espnow · canonical fleet · `ledger-provision` |
| `clippy -D warnings` | canonical tier |
| host suites | every `experiments/*_verify`, found by **glob** — a new verifier is picked up automatically |
| **stack floor** | canonical ELF `.stack` vs 73,728 B, **printed to the PR job summary** |

Tiers naming a feature absent on the branch are **skipped, not failed**, so this works on both sides of the #266 merge.

## Proof it can fail

A gate nobody has watched fail is a gate nobody should trust — this is the third "gate that couldn't fail" found this session, so each arm was broken deliberately first:

| injected breakage | result |
|---|---|
| `needless_range_loop` reintroduced (the exact 4c5955f class) | 🔴 `FAIL clippy canonical` |
| stack floor raised above actual | 🔴 `FAIL stack floor` — `FATAL: runtime stack is 75560 B, below the 80000 B floor.` reaches the job summary |
| host verifier made to assert false | 🔴 `FAIL etx_verify` |
| syntax error in tracked source | 🔴 `GATE RED — 4 failure(s)` |
| all reverted | 🟢 `GATE GREEN` |

One honest note: my first stack-breach attempt injected an **immutable** all-zero `static`, which lands in `.rodata` (flash), not `.bss` (DRAM) — so it moved nothing and the arm looked un-failable. The gate was right and the test was wrong. Recorded because that is the same "suspect the instrument" failure the stack gate itself exists to catch.

## Why `repro_build.sh` changed

- `repro_stack_check()` extracted from `repro_build_bin` so **CI measures with the same code that gates shipping**. A CI copy of that arithmetic would be a second definition free to drift. Behaviour-preserving: verified byte-identical output (75,560 B) by running the pre-refactor script against the same tree.
- `REPRO_FLEET_FEATURES` names the canonical tier once. It previously existed only inside the build line, so CI could have gated a *different* tier than the one that ships and nothing would have said so.

## Two supporting changes

- **`tools/ci_provision.sh`** — a clean checkout has no `secrets.rs`/`board.rs` (git-ignored; public repo), which is precisely *why* no CI job ever built a tier. Non-destructive: an existing file is left alone, so running the gate locally cannot eat real credentials. Forward-compatible with #266 — when the example ships the published all-zero `GROUP_KEY` that #336's assert refuses, it substitutes a random throwaway (verified: 32 valid `u8`s, non-zero, stable across runs).
- **`cfgsched.rs`** — item-scoped `#[allow(dead_code)]` on `ticks_to_cover`/`peek`. Both are genuinely exercised by `cfg_relay_verify` via `#[path]` and dead only in the binary, the same precedent as `net::{ledger,treehead,sth}`. **Prerequisite:** the canonical tier could not go green while `main` still failed its own gate.

## Not covered — stated, not hidden

- `clippy -D` on `default`/`wifi`: pre-existing dead-code findings there (symbols used only by espnow builds). They get `cargo check`, so compile breaks are caught but new warnings are not. Worth a follow-up.
- `mesh-test` (needs a real board's `DEAF_MACS`), espflash `save-image` packaging, and anything needing hardware.
- ⚠️ The stack check bounds the linked **region**, not runtime high-water. #181's `LedgerLink` costs 1,760 B of real stack while moving the region by ~32 B. **A green stack line is not evidence of runtime headroom.**

### DIFF
--- a/.github/workflows/fw-gate.yml
+++ b/.github/workflows/fw-gate.yml
@@ -1,0 +1,84 @@
+# #338 — the firmware gate, in CI.
+#
+# Before this, `.github/workflows/` held only `pages.yml`: nothing built any feature tier, ran
+# clippy, or ran the host suites. The "4-tier gate" cited in ONBOARDING.md and rust-toolchain.toml
+# was prose plus whoever remembered to run it. On 2026-08-01 that cost us main sitting red on its
+# own `-D warnings` rule unnoticed, a new feature (`ledger-provision`) with no matrix to add it to,
+# and two agents separately hand-building baselines to compute a clippy delta.
+#
+# DELIBERATELY THIN: every actual command lives in `tools/gate.sh`, which a developer runs verbatim
+# before pushing. If the gate logic lived here, the local instructions and CI would be two
+# definitions free to drift — the same failure in a new place. This file only supplies a machine.
+name: fw gate
+
+on:
+  pull_request:
+  push:
+    branches: [main]
+  workflow_dispatch:
+
+# A force-push supersedes its own in-flight run; main runs are never cancelled.
+concurrency:
+  group: fw-gate-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+env:
+  CARGO_TERM_COLOR: always
+  # Keep a runner's memory in bounds: these are LTO release builds of an embedded crate.
+  SMOL_GATE_JOBS: 2
+
+jobs:
+  # Split so a Rust-only breakage is distinguishable at a glance from a host-logic regression,
+  # and so the host suites still report when the riscv toolchain is the thing that broke.
+  host:
+    name: host verifier suites
+    runs-on: ubuntu-latest
+    timeout-minutes: 25
+    steps:
+      - uses: actions/checkout@v4
+      # rust-toolchain.toml pins `stable` + the riscv target + clippy; rustup honours it
+      # automatically, so there is no version to restate here (and none to drift).
+      - uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: experiments
+          key: host
+      - run: tools/gate.sh host
+
+  firmware:
+    name: tiers + clippy + stack floor
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    steps:
+      - uses: actions/checkout@v4
+      - name: Show resolved toolchain
+        working-directory: rust/clock
+        run: rustc -vV && cargo clippy --version
+      - uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: rust/clock
+          key: fw
+      # readelf reads _stack_start/_stack_end out of the riscv ELF (arch-independent); ubuntu-latest
+      # ships binutils, but assert it rather than discovering its absence as a confusing FATAL.
+      - name: Assert readelf present
+        run: readelf --version >/dev/null
+
+      - run: tools/gate.sh fw
+
+      # The single highest-value line in this workflow, surfaced where a reviewer sees it without
+      # opening logs. Both the #233 spike and the #266 rebase turned on this number and both found
+      # it at the last possible moment; it was previously computed only at package time.
+      - name: Stack number → job summary
+        if: always()
+        run: |
+          {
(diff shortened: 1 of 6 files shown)


jphein:
**Live CI confirmation** — run [30710185400](https://github.com/jphein/smol/actions/runs/30710185400), both jobs green on this PR.

The workflow ran itself on a genuinely clean checkout, which is the part that could not be proven locally:

```
── provisioning (git-ignored; existing files untouched)
  secrets.rs: created from secrets.rs.example
  board.rs:   created from board.rs.example
── cargo check — build tiers
   PASS default · PASS wifi · PASS espnow · PASS fleet
   SKIP ledger-provision (feature not in Cargo.toml on this branch)
── cargo clippy -D warnings — canonical tier (espnow,cast,io,bard)
   PASS
── stack floor — canonical ELF vs 73728 B
  stack: 75560 B (floor 73728 B)
── host verifier suites
   13/13 PASS
GATE GREEN
```

Three things this establishes beyond the local run:

1. **The stack number is 75,560 B in CI — byte-identical to the local measurement.** An ELF that was never built has no `_stack_start`, so this is proof the riscv cross-build really happened, *and* incidental evidence that the reproducible-build machinery reproduces across machines.
2. **Clean-checkout provisioning works.** `secrets.rs`/`board.rs` were created from the examples by `ci_provision.sh` — the missing step that made a CI firmware build impossible before now.
3. **`ledger-provision` skipped rather than failed**, so the gate is already correct on both sides of the #266 merge; it starts covering that tier the moment the feature lands.

13 suites here, not 14 — `mac_verify` arrives with #266.

Issue #339, opened 2026-08-01
DIAG shed order: the documented ranking contradicts the mechanism (cc survives longer than ap)


jphein:
Found during the #266 core rebase (2026-08-01) while documenting where the new ledger fields sit in the shed order.

**The mechanism:** DIAG's sheddable fields are offered through `room_for` in append order, filling while budget lasts. So **append position IS shed priority — last offered is first dropped.** The actual order today:

```
sog → cfgq → cdeaf → cc/degraded → ap → cfg → io → [deaf] → lg*
```

**The prose in the code disagrees with it.** The existing comment says the first to go, as "least missed", are: `cc`, `io`, `cdeaf`, `cfg`, `ap`. But `cc` is appended **4th**, so it survives *longer* than `ap`, `cfg` and `io` — the documented ranking is close to inverted for `cc`, and nobody's stated ranking matches the mechanism.

**Why this is worth more than a comment fix:** an operator diagnosing a truncated record who lost `ap=` and trusted that prose would conclude the record was shed down past `cc`/`io`/`cdeaf`/`cfg` first, and look in the wrong place. `ap=` and `cc=` are exactly the fields the #204/#217 coexist and off-channel-deafness investigations rest on, so this misleads precisely where accuracy matters most.

The comment has been corrected in place (both orders now spelled out) in the #266 core branch — doc-only. **The open question is which ranking is intended:**

- If the *documented* ranking is what we want (drop `cc`/`io` early, protect `ap`), then the append order is a real behaviour bug and the appends should be reordered. That is wire-visible and wants its own commit and its own reasoning about what an operator most needs when the budget is tight.
- If the *current* order is what we want, the prose fix alone closes this.

My read: `ap=` (channel + RSSI + BSSID) and `cc=` are the highest-value fields under coexist pressure and should shed LAST, which argues the current order is closer to right than the prose — but that is a judgement about diagnosis priorities, not a mechanical fix.

Related: the new `lg*` ledger fields are appended last and therefore shed FIRST, deliberately, so ledger observability never displaces coexist diagnosis. There is also now a `shed=<n>` counter so an absent field is distinguishable from an inactive feature — worth keeping whichever way this lands.

PR #346, opened 2026-08-01
ci: #345 run the gate on every branch push, not just main

### LINKED ISSUE #345: ci: a commit pushed after its PR closes gets ZERO CI (the gate's own hole)
Found by the gate failing to run on its own follow-up commit, one hour after the gate landed.

`fw-gate.yml` triggers on `pull_request` and `push: branches: [main]`. A commit pushed to a **feature branch whose PR has already closed** matches neither: no `pull_request` event (the PR is closed) and no `push` event (not main). It is **silently ungated** — no red, no skip, no check at all.

### How it surfaced

#341 was squash-merged while `ec16fe3` (the shed-order follow-up) was still in flight. The squash captured only `6753c3a`; the follow-up landed on a now-closed branch and got **zero CI**. It survived only because the same work had been opened as PR #342 — otherwise it would have sat there indefinitely with nothing pointing at it.

That is the **second squash-merge race in one session**; the first cost 175 lines of WIP.

### Fix

- `push: branches: ['**']` — every branch. Branches-only on purpose: `'**'` does not match tags, so a release tag does not re-run a forty-minute gate on a commit that already passed.
- Re-key concurrency on the **branch** (`github.head_ref || github.ref_name`) so the `push` and `pull_request` events for one branch share a group and dedupe instead of double-running. The old key (`github.ref`) could not dedupe them — a PR's ref is `refs/pull/N/merge`, a push's is the branch, always distinct groups.
- `main` keeps its own group and `cancel-in-progress: false`, so an in-flight main gate always completes. Two independent guarantees: a branch push cannot reach `fw-gate-main`, and even within it nothing cancels.

### Cost, stated

Every branch push now runs the gate (~1–2 min). With several agents pushing worktree branches that is real minutes, mitigated by same-branch dedupe. The alternative is what we have: commits that are silently ungated at exactly the moment a merge race makes them easiest to lose.

jphein:
Closes #345.

A commit pushed to a feature branch whose PR had already closed triggered **nothing**: no `pull_request` event (PR closed) and no `push` event (filter was `branches: [main]`). Silently ungated — a hole in the tool built to close holes.

Not hypothetical. It happened to this workflow's own follow-up **within an hour of it landing**: #341 was squash-merged while `ec16fe3` was in flight, and that commit sat on a closed branch with zero CI. It survived only because the same work had been opened as #342. Second squash-merge race of the day; the first cost 175 lines.

## Change

| | before | after |
|---|---|---|
| trigger | `pull_request` + `push: [main]` | `pull_request` + `push: ['**']` (branches only — **not tags**) |
| concurrency key | `github.ref` | `github.head_ref \|\| github.ref_name` (the **branch**) |
| main | own group, never cancels | unchanged — own group, `cancel-in-progress: false` |

Re-keying is what lets `push` and `pull_request` for the same branch **dedupe into one group** instead of double-running. The old key couldn't: a PR's ref is `refs/pull/N/merge`, a push's is the branch — always distinct.

## Proven, not argued

**(a) A branch push with no PR now fires the gate.** This branch demonstrated its own fix: pushed with **zero** PRs open (`state=all` query returned 0), and a run appeared with `event: push`. That is precisely the case that used to produce nothing.

**(b) Concurrency behaves, both directions** — measured with a throwaway probe branch, since "it should be a different group" is a claim, not evidence:

| scenario | expected | observed |
|---|---|---|
| run A in flight on branch 1, push to **branch 2** | A survives (different group) | **A: completed/success** ✅ |
| run B in flight, second push to the **same branch** | B superseded (dedupe works) | **B: completed/cancelled** ✅ |

The second is the one that makes `push` + `pull_request` coexist without double-running; the first is why a branch push can never disturb main. For `main` there are two independent guarantees — a branch push cannot reach `fw-gate-main`, and `cancel-in-progress` is `false` there — so an in-flight main gate always completes. Over-running is the cheap failure; a cancelled main gate is not.

Probe branch deleted afterwards (remote + local, verified gone).

## Security

`github.head_ref` is attacker-controlled (a fork can name a branch anything). It is used **only** as part of a concurrency group name — an opaque scheduling label, never evaluated — which is the form GitHub's own docs recommend. Audited: the two `concurrency` lines are the **only** interpolations in the file, and none appear in a `run:` shell or a checkout `ref:`.

## Cost, stated plainly

Every branch push now runs the gate (~1–2 min). With several agents pushing worktree branches that is real minutes, mitigated by same-branch dedupe. The alternative is the status quo: commits silently ungated at exactly the moment a merge race makes them easiest to lose.

Unrelated but worth recording: main's stack number re-measured after the #339/#343 merges — still **75,560 B**, confirming those changes (comments, item-scoped allows, tooling) moved no code.

### DIFF
--- a/.github/workflows/fw-gate.yml
+++ b/.github/workflows/fw-gate.yml
@@ -13,14 +13,37 @@ name: fw gate
 
 on:
   pull_request:
+  # #345: EVERY branch, not just main. With `branches: [main]` a commit pushed to a feature branch
+  # whose PR had already closed triggered NOTHING — no `pull_request` event (the PR is closed) and
+  # no `push` event (not main) — so it was silently ungated. That is not hypothetical: it happened
+  # to this workflow's own follow-up commit within an hour of it landing (#341 squash-merged while
+  # `ec16fe3` was in flight; the commit sat on a closed branch with zero CI and was only rescued
+  # because the same work had been opened as a second PR).
+  #
+  # `'**'` is branches-only by design — it does NOT match tags. A release tag should not spend
+  # forty minutes re-running a gate that already passed on the commit it points at.
   push:
-    branches: [main]
+    branches: ['**']
   workflow_dispatch:
 
-# A force-push supersedes its own in-flight run; main runs are never cancelled.
+# Keyed on the BRANCH, deliberately, so the `push` and `pull_request` events for the same branch
+# land in ONE group and dedupe instead of double-running: `github.head_ref` is the PR's source
+# branch and is empty for a push, where `github.ref_name` supplies the same name. Keying on
+# `github.ref` (the previous form) could not dedupe them — a PR's ref is `refs/pull/N/merge` and a
+# push's is the branch, so they were always distinct groups.
+#
+# `main` gets its OWN group and NEVER cancels: a branch push cannot reach `fw-gate-main`, and
+# cancel-in-progress is false there, so an in-flight main gate always runs to completion. Over-
+# running is the cheap failure here; a main gate cancelled by unrelated branch activity is not.
+# (Two independent guarantees on purpose — either alone would do.)
+#
+# SECURITY: `github.head_ref` is attacker-controlled (a fork can name a branch anything), so it must
+# never reach a `run:` shell or a checkout `ref:`. Here it is used ONLY as part of a concurrency
+# GROUP NAME — an opaque scheduling label that is never evaluated — which is the form GitHub's own
+# docs recommend. These two lines are the only interpolations in this file; keep it that way.
 concurrency:
-  group: fw-gate-${{ github.ref }}
-  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+  group: fw-gate-${{ github.head_ref || github.ref_name }}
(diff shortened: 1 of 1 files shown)


jphein:
**Live results on this PR — and one honest nuance about the dedupe.**

Both runs green:

| event | conclusion | what it proves |
|---|---|---|
| `push` | success | **the fix** — fired on a branch with zero PRs open, the case that used to trigger nothing |
| `pull_request` | success | opening the PR still gates normally |

**Nuance: you see two runs above, not one.** The dedupe did not fire here because the events were *sequential* — the `push` run had already completed before the PR was opened, and a finished run is not superseded. Concurrency only collapses runs that **overlap**.

That is fine, and the steady state is the overlapping case: once a PR is open, each subsequent push fires `push` and `pull_request` *simultaneously*, they land in the same branch-keyed group, and one supersedes the other → one effective run. That mechanism is proven directly in the PR body's probe table (`B: completed/cancelled` on a same-branch second push).

So the accurate claim is: **overlapping events dedupe; sequential ones both run.** The worst case is one extra ~1–2 min run on the first push of a branch that later gets a PR — which is the cost line already stated, and cheaper than the ungated commit this replaces.

PR #364, opened 2026-08-01
fix(ui): clip display text on a character boundary, not a byte index (#274)

### LINKED ISSUE #274: [c6→smol reverse-port] one char-boundary-safe clip() in smol-core — smol has four copies, two of them unsafe
_Rewritten 2026-08-01 (triage). **Severity downgraded — the remote-crash path is not reachable today — and scope widened: smol has four `clip()` helpers, not one.** Original body preserved at the bottom._

**Direction:** watch → smol.

## The defect, restated accurately

smol's `rust/clock/src/rssi.rs:172-173`:

```rust
/// ASCII-safe left-truncate to `n` bytes (magical nouns are ASCII, so a byte
/// boundary is a char boundary — never panics).
pub fn clip(s: &str, n: usize) -> &str {
    &s[..s.len().min(n)]
}
```

If `n` lands inside a multibyte UTF-8 char this panics. The doc comment states the invariant that keeps it safe and **nothing enforces it**. That is real and worth fixing.

## What the original body got wrong: it is not a remote-crash vector today

The original claimed *"if `clip()` is ever fed a peer-supplied name arriving over ESP-NOW (attacker-controllable, needn't be ASCII), that's a remote-crash vector."* Traced end-to-end; no such path exists in smol @ `be0bdf8`:

- `rssi::clip` has **exactly one consumer** — `src/ota_screen.rs:43` imports it; calls at `:236`, `:265`, `:279`.
- All three clip a locally-built `Line`. Its string inputs are the FORGE codename adjective+noun and `name_for_seed_id_noun(..)` — both drawn from a fixed ASCII wordlist keyed by node id, so a *peer* contributes an **id**, never bytes.
- The one externally-supplied string that reaches it is the OTA URL host (`ota_screen.rs:261`, `write!(src, "from {}", host)`), and the announce parser **rejects a non-ASCII URL outright** — `src/ota.rs:245`: `if url.is_empty() || url.len() > URL_MAX || !url.is_ascii() { return None; }`. Worth noting the URL is *not* covered by the ed25519 signature (`ota.rs:248`, M = `"build|size|sha256hex"`), so that `is_ascii()` check is load-bearing, not incidental.
- `Line::as_str()` is `core::str::from_utf8(&self.buf[..self.len]).unwrap_or("")` (`rssi.rs:153-155`), so even a torn multibyte write degrades to `""` rather than handing `clip` invalid UTF-8.

So: **latent hardening, not a live security bug.** It becomes reachable the moment any peer-supplied or operator-supplied *text* (a custom-screen string, a CFG-set label, a relayed noun) is rendered through this path — which is a likely future, not the present.

## What the original body undercounted: there are four `clip`s

| location | implementation | safe? |
|---|---|---|
| `rust/clock/src/rssi.rs:172` | `&s[..s.len().min(n)]` | ❌ byte slice |
| `rust/clock/src/bench.rs:166` | `&s[..s.len().min(n)]` | ❌ byte slice |
| `rust/clock/src/grid.rs:292` | `s.char_indices().nth(max)` | ✅ |
| `rust/clock/src/batt.rs:316` | `s.char_indices().nth(max)` | ✅ |

Two of four already walk char boundaries — so smol has both the bug **and** its fix in-tree, in different files, with no single owner. `rust/clock/src/net/names.rs:109` already documents the split in a comment:

> *"char_indices, not byte slicing: the `clip()` helpers in bench.rs/rssi.rs slice bytes and are safe only while the corpus stays ASCII. This one cannot be made to panic by a word change."*

A three-line patch to `rssi.rs` alone leaves `bench.rs` byte-slicing and leaves four implementations of one function.

## Reference implementation (watch-side, already done)

esp32c6-watch `crates/rssi/src/lib.rs:177-186` — verbatim port of this module with the fix:

```rust
pub fn clip(s: &str, n: usize) -> &str {
    if s.len() <= n { return s; }
    let mut end = n;
    while end > 0 && !s.is_char_boundary(end) { end -= 1; }
    &s[..end]
}
```

## Proposed scope

**One hardened `clip` in `smol-core` (#347 Phase 2); all four smol call sites converge on it**, plus a multibyte fuzz test over every prefix length. `crates/rssi` is a candidate donor. Doing this standalone is still cheap and correct if core slips — but do all four, and delete the duplicates.

<details><summary>Original body (2026-07-20, smol↔c6 sync monitor)</summary>

**Bug (latent in smol):** `rust/clock/src/rssi.rs::clip(s, n)` does `&s[..s.len().min(n)]` — a **byte** slice. If `n` lands inside a multibyte UTF-8 char, Rust panics. The doc assumes ASCII-only but doesn't enforce it. If `clip()` is ever fed a **peer-supplied name arriving over ESP-NOW** (attacker-controllable, needn't be ASCII), that's a **remote-crash vector**. [**Not reachable today — traced above.**]

**Fix (already done watch-side):** the watch's `crates/rssi` fixed it at feat/rssi-crate@a2f5af8 — walk back from `n` to the largest UTF-8 char boundary via `is_char_boundary`, + a `clip_never_panics_on_multibyte_utf8` test. 10/10 green.

**Reverse-port sketch:** replace the byte-slice in smol's `clip()` with the char-boundary walk (~3 lines) + add the multibyte fuzz test. [**Undercounted — there are four `clip` helpers.**]
</details>

jphein:
Closes #274.

## The bug

`&s[..s.len().min(n)]` panics when the cut lands inside a multi-byte character. Two live copies — `bench.rs` and `rssi.rs` (the latter `pub`, used by `ota_screen.rs`) — each carrying a comment asserting the input was ASCII.

The crate had **four** clip helpers with **two different units**:

| where | body | unit |
|---|---|---|
| `batt.rs`, `grid.rs` | `char_indices().nth(max)` | characters, boundary-safe |
| `bench.rs`, `rssi.rs` | `&s[..s.len().min(n)]` | **bytes, panics off-boundary** |

Replaced with one pure `textclip::clip`, character-unit — which is what every caller wants, since they are all fitting text into a fixed-width font on a 128 px panel.

## Severity: corrected

**This is not a remote panic vector today.** The brief described the input as a remote-supplied AP SSID / peer-reported string; I traced every call site and that isn't what reaches `clip()`. Each input is provably ASCII:

- both sigil corpora — **0** non-ASCII inside word literals (my first grep said 22, but it was counting em-dashes in *comments*; the word lists are clean)
- `host` — the only externally-derived string, via `OtaKind::SelfFetch`. `parse_announce` enforces `url.is_ascii()`, so it cannot be non-ASCII
- everything else is numeric or a literal

**The realistic trigger is a corpus change, not an attacker.** The words come from `sigil-names/src/realms.rs`, whose header reads *"Generated by sync-words.sh from words/realms.json — DO NOT EDIT"*. One upstream word with an accent panics every board that renders it. `net/names.rs` already carried the warning — *"the `clip()` helpers in bench.rs/rssi.rs slice bytes and are safe only while the corpus stays ASCII"* — which names a hazard rather than guarding against it.

Still worth fixing at the same priority: the failure mode is a fleet-wide panic loop arriving through a routine word-list sync, which is a worse way to find out than an attacker.

I also swept the crate for the same shape. The other `&x[..n]` sites are byte slices (`STORY_TEXT`/`PROMPT` are `[u8; N]`) or use indices from `find()`, so they're safe.

## The test is the deliverable

`experiments/clip_verify` keeps a **verbatim copy of the old body** and proves it *still panics* on the motivating input — because `clip` is three lines and looks exactly like something a future cleanup "simplifies" back. A suite that only exercised the new function would go green against the reintroduced bug.

```
#274 — the old form vs the new one
  ok      old `&s[..s.len().min(n)]` PANICS at n=5 (inside 'é')
  ok      new clip(word, 5) = "Forgé" — same call, no panic
  ok      panic-free + prefix + monotonic for every budget on 8 strings
  ok      byte-identical to the old form on every ASCII input (no render change)
```

It also caught its own first draft: byte **6** of `"Forgéd"` is a boundary (start of `d`), so the cut inside `é` is byte **5**. A regression test aimed one byte off would have passed while proving nothing — so the premise is now asserted with `is_char_boundary` rather than assumed.

The ASCII-equivalence check is what makes this safe to land: character-unit and byte-unit agree on every current input, so nothing the fleet renders shifts.

## Gate

`tools/gate.sh all` **GREEN** — full tier matrix, clippy `-D warnings` on every tier, host suites, and the crate tests.

Stack **106,472 B** (floor 74,208). `main`@`b2537c4` measured identically: **106,472 B** — **0 B delta**, as expected for a text-only change.

`wifi`-gated (the lowest tier with a consumer), so the default image stays byte-free of it rather than carrying an `allow(dead_code)` that would mask a genuinely unused helper later.


jphein:
**Closing as superseded by `e231d8c`** — that landed the same fix while this was in flight. No complaint; parallel work, and theirs is in main.

Recording the comparison, since two of my claims were wrong and the record should say so:

**1. Theirs is better on the real path.** It `#[path]`-includes the actual `rssi.rs` and tests the `Line` buffer that genuinely feeds `clip()` at every call site — including `Line`'s own 24-byte truncation tearing a multibyte char, with `as_str()`'s `from_utf8(..).unwrap_or("")` as the backstop. My suite used synthetic strings and never touched the real feeder.

**2. Semantics differ, and theirs is the more conservative choice.** Theirs keeps a **byte** budget (walk backward from `n`); mine switched `rssi`/`bench` to a **character** budget. For ASCII they coincide, so neither changes what the fleet renders today. Theirs preserves the existing call-site contract exactly; mine would have changed truncation for non-ASCII. Given the callers size budgets in bytes assuming 1 byte == 1 glyph, theirs is right.

**3. I was wrong that their suite doesn't guard reintroduction.** I opened this thinking their old-vs-new comparison ran only over ASCII (where the two agree by construction), so a "simplification" back to `&s[..s.len().min(n)]` would slip through, and I was going to add a `catch_unwind` exhibit.

I tested that claim by actually re-simplifying `clip()` and running the **pristine** suite. It fails immediately — exit 101, panicking inside `clip()` on the first `sweep("café", ..)`. The `sweep` calls feed multibyte input directly through `clip()`, so a byte-slicing implementation cannot survive them. The guard is real and my proposed addition was redundant.

**One genuine residual, for whoever wants it:** `e231d8c` deliberately leaves `batt.rs` and `grid.rs` with their own `clip()`s (its commit says so). Those two are already boundary-safe, so there is no panic risk — but the crate now has **two units**: `rssi`/`bench` truncate to *n bytes*, `batt`/`grid` to *n characters*. Both locally correct, both documented. Whether that is worth unifying is a judgement call, not a defect, and it is not urgent.

Superseded work; nothing here needs to land.

Issue #350, opened 2026-08-01
ci: derive the (chip, tier) build matrix from one manifest, and split "what CI builds" from "what ships"


jphein:
Prior-art study (OpenWrt · WLED · Tasmota) for #331/#347. Full report: `scratch/parity/multitarget-prior-art.md`.

## The problem

Once #331 lands more chips, `chips × tiers` is unbounded. #338 already fixed the one-variable version of this — `REPRO_FLEET_FEATURES` is named once so CI cannot gate a different tier than the one that ships. That lesson now needs to generalize from one variable to a table.

There is also a live warning in #335: **both Embassy branches produced "all tiers green" evidence without the stack-floor gate** (4 references on `main`, 0 on both branches, because both predate #300). A per-chip CI job that omits the gate would repeat that exact error at larger scale.

## What the prior art does

**WLED has the best CI shape of the three: "what exists" and "what ships" are separate declarations, and the matrix is derived rather than hand-listed.**

- [`build.yml`](https://github.com/wled/WLED/blob/main/.github/workflows/build.yml) reads the matrix out of the config:
  ```yaml
  run: echo "environments=$(pio project config --json-output | jq -cr '.[0][1][0][1]')" >> $GITHUB_OUTPUT
  ```
  -> one GH matrix job per env, fed by `default_envs`. Nothing is typed twice.
- [`platformio.ini`](https://github.com/wled/WLED/blob/main/platformio.ini) has `default_envs = nodemcuv2` — **every PR builds exactly one env.**
- Tagged releases copy [`platformio_release.ini.template`](https://github.com/wled/WLED/blob/main/.github/platformio_release.ini.template) over it, listing **29 envs**. The template's own header documents the design's cost: *"It MUST list every env that should be released - including the regular CI default_envs - because it fully replaces the parent value."*

**Tasmota avoids explosion by refusing to multiply the axes.** From [`Tasmota_build_master.yml`](https://github.com/arendst/Tasmota/blob/development/.github/workflows/Tasmota_build_master.yml), a release is **99 binaries**: 17 safeboot + 10 ESP8266 + 18 ESP32 + 27 ESP8266-language + 27 ESP32-language. The 27-locale axis is crossed **only with the single `tasmota`/`tasmota32` spine variant**:

```yaml
matrix:
  variant: [ tasmota ]
[...]

PR #365, opened 2026-08-01
ci(#351): the BYTE-FREE claims, measured instead of asserted

### LINKED ISSUE #351: ci: prove the "BYTE-FREE" tier claims with a DWARF compilation-unit check (map files are unusable under fat LTO)
Prior-art study (OpenWrt · WLED · Tasmota) for #331/#347. Full report: `scratch/parity/multitarget-prior-art.md`.

## The claim we make but do not check

`rust/clock/Cargo.toml` asserts, repeatedly and in prose, that default/lower tiers are byte-free of higher-tier code:

- `io`: *"Default-OFF so the default/wifi/espnow builds stay BYTE-FREE of it (#44)"*
- `wled`: *"The default build is BYTE-FREE of it (symbol-absence provable; net/wled.rs is `#![cfg(feature = "wled")]`...)"*
- `cast`: *"The default / wifi / espnow / wled builds are BYTE-FREE of it"*
- `mesh-test`: *"a production build is BYTE-FREE of it AND structurally CANNOT be made deaf"*
- `stack-paint`: deliberately excluded from the canonical fleet list because *"it writes the whole free stack at boot"*

Each says "provable". None is proven by a gate. As tiers and chips multiply (#331), a stray non-`cfg`-gated hook in a shared path silently ships everywhere — and the symptom would be a size or stack regression attributed to something else.

Related standing anti-lesson: a literal grep proves nothing about constructed strings, and a feature-name grep gives a lower bound rather than a count. `#[cfg]` coverage is exactly the kind of claim a grep cannot close.

## What the prior art does

WLED automates the inverse check. [`pio-scripts/validate_modules.py`](https://github.com/wled/WLED/blob/main/pio-scripts/validate_modules.py) confirms every *selected* module actually contributed code to the linked ELF, and it documents the trap we would otherwise walk into:

> *"The map file is not a reliable source for this: with LTO, original object file paths are replaced by temporary ltrans.o partitions in all output sections, making per-module attribution impossible from the map alone. Instead we invoke `readelf --debug-dump=info --dwarf-depth=1` on the ELF, which reads only the top-level compilation-unit DIEs from `.debug_info`. Each CU corresponds to one source file... The output volume is proportional to the number of source files, not the number of symbols."*

That caveat matters here: smol builds with `lto = "fat"` + `codegen-units = 1`, so any map-file-based attribution would be wrong in exactly the way WLED describes.

They also have the motivating case in their build config — [`platformio.ini`](https://github.com/wled/WLED/blob/main/platformio.ini), the P4 env:

```ini
wled-espnow     ; WLED_DISABLE_ESPNOW compiles it empty; validate_modules.py would fail on "no symbols found"
```

A module that compiles to nothing is indistinguishable from a module that failed to compile, unless something checks.

## Proposed

A gate, run per tier, in the same family as the #300 readelf stack-floor gate (which is the precedent that this kind of check works and is cheap):

1. **Absence direction** (our claim): for each tier, assert that the compilation units of `cfg`-excluded modules are **absent** from the ELF's `.debug_info` — `net/wled.rs` must not appear in a `default` build, `io.rs` must not appear in a `wifi` build, `bard/stack_paint.rs` must not appear in any staged image.
2. **Presence direction** (WLED's): for each tier, assert the modules it *does* select **are** present. Catches a feature that silently compiles to nothing after a refactor.
3. Use `readelf --debug-dump=info --dwarf-depth=1`, **not the map file** — see the LTO caveat above.
4. Wire it into `tools/gate.sh` beside the existing gates, and have it **report the module list**, not a boolean. A gate that reports "green" is a gate whose failure mode is invisible; one that reports the set is auditable.

Fails closed: if `readelf` cannot be resolved or the ELF has no `.debug_info`, that is a gate failure, not a skip. (WLED's own script chooses a conservative *pass* on readelf failure — reasonable for a hobby build farm, wrong for a brick-adjacent OTA pipeline.)

jphein:
Closes #351.

`Cargo.toml`, `net.rs` and `main.rs` all state that a lower tier contains none of a higher tier's code — *"the default build is BYTE-FREE of it (symbol-absence provable)"* — and **nothing checked a single one of them**. Same species as the `gate.sh` comment #350 disproved for three features at once.

## Verdict — all 10 manifest tiers linked, zero leaks

```
default           11 crate files · 44 modules claimed absent · 0 leaked
wifi              20 crate files · 35 modules claimed absent · 0 leaked
espnow            45 crate files · 11 modules claimed absent · 0 leaked
fleet             48 crate files ·  8 modules claimed absent · 0 leaked
bard              54 crate files ·  2 modules claimed absent · 0 leaked
ledger-provision  45 crate files · 11 modules claimed absent · 0 leaked
wled              49 crate files ·  7 modules claimed absent · 0 leaked
coexist-soak      48 crate files ·  8 modules claimed absent · 0 leaked
mesh-test         48 crate files ·  8 modules claimed absent · 0 leaked
stack-paint       55 crate files ·  1 modules claimed absent · 0 leaked
10 tiers · 44 claims · 43 observed · 1 declared unobservable
```

The substantive claims **hold**. `default` really is free of wled / cast / io / target / election / etx / flood / wire / mqtt code.

## Why not compilation units

The issue title (and WLED's prior art in `scratch/parity/multitarget-prior-art.md`) prescribes asserting each module contributed a **CU**. That does not transfer, and the difference is the whole design:

* WLED is C++ — one CU per `.cpp`, so "module" and "CU" coincide.
* smol's firmware is **one Rust crate** at `codegen-units = 1` + `lto = "fat"`. Measured: the linked ELF has **exactly one** CU whose `DW_AT_comp_dir` is the crate (`src/main.rs/@/clock.<hash>-cgu.0`). A per-CU assertion would be true for every module in every tier forever — *a gate that cannot fail*, which is the thing this issue is about.

The unit that survives is one level down: the CU's **line-table file set**, which attributes **inlined** code back to its source file (verified — `display-interface-i2c` rows sitting inside `ssd1306` addresses). WLED's other finding — that map files are unusable under LTO — stands and is why neither a map nor `nm` is used here.

## Failure demonstration

Not a green-only claim. Ungate `pub mod target;` in `net.rs` and plant a `net::target::chip_name()` call in `main()` — the realistic regression, where the gate is dropped so a new call site compiles and the byte-free comment above it is left untouched:

## What it proves, and what it does not

**Proves:** no *executable* bytes from an excluded module survived, inlined or otherwise.

**Does not prove:** no *data*. A consts-only module lands in `.rodata` with no line rows. `src/secrets.rs` is exactly that, and is declared in `build-matrix.toml`'s new `[unobservable]` **with its reason** — same discipline as `[exempt]`, including refusing a stale entry whose module *is* observed.

**Anti-vacuity:** every claim must be OBSERVED at least once across the run, or it is reported **UNPROVEN**, not passing. An absence check whose broken state and passing state print the same green needs this or it is decoration.

## Two claims narrowed

* `Cargo.toml` and `net.rs` listed the **`wled` build** among those byte-free of `cast`. Since #350 the `wled` **tier** is `wled,${canonical}` and the canonical features include `cast` — so that tier carries it. The old wording named a feature combination nothing builds.
* `Cargo.toml` cited `src/io.rs` as carrying `#![cfg(feature = "io")]`. It does not; it is gated at its `mod io;` in `main.rs`. Same effect, wrong citation.

## Mechanics

* `tools/gate.sh excl` is a new mode and a **separate CI job**. This arm LINKS every tier (measured **790 s** total on a dev box) on top of a `firmware` job already budgeted at 45 minutes. Bolting it on would have made the timeout the thing that fails, and a gate that times out gets deleted. The jobs share no state.
* Builds carry `CARGO_PROFILE_RELEASE_DEBUG=line-tables-only` into their own target dir. **`[profile.release]` is untouched** — debug info lands only in non-ALLOC sections, and `.text` / `.rodata` / `.bss` / `.stack` come out the same **size** with and without it (`.text` 892,022 B both).
* Claims are **derived, never listed**: the walker follows `mod` declarations transitively from `src/main.rs` and reads the `#[cfg(feature = …)]` on each. 44 gated modules today. Gates compose — `src/bard/stack_paint.rs requires bard+stack-paint`.

## `tools/test_check_exclusions.sh` — 11 arms, each proven to fail

green control · leak · **subtree** leak · unproven · honest `[unobservable]` · stale `[unobservable]` · ghost `[unobservable]` · compound-cfg refusal · missing-module refusal · **ELF-with-no-DWARF refusal** · truncated-fileset refusal.

That DWARF arm is the one that matters: `debug = false` is the release default, so an ELF with no debug info yields an empty file set under which *every* absence check trivially holds. That is exit 2, not green.

### DIFF
--- a/.github/workflows/fw-gate.yml
+++ b/.github/workflows/fw-gate.yml
@@ -168,7 +168,9 @@ jobs:
             tools/check_exclusions.py claims || true
             echo '```'
             echo "Derived from the \`#[cfg(feature = …)]\` on each \`mod\`, walked from"
-            echo "\`src/main.rs\`. Gate a module and it is covered; there is no list to update."
+            echo "\`src/main.rs\`, and cross-checked both ways against \`[tier_exclusive]\` in"
+            echo "\`tools/build-matrix.toml\` — so REMOVING a gate fails the gate instead of"
+            echo "quietly removing the claim along with it."
             echo ""
             echo "⚠️ This proves no **executable** bytes from an excluded module survived, via the"
             echo "DWARF line table (which attributes inlined code back to its source file). A"
--- a/.github/workflows/fw-gate.yml
+++ b/.github/workflows/fw-gate.yml
@@ -133,3 +133,45 @@ jobs:
             tools/build_matrix.py ci-matrix | python3 -m json.tool
             echo '```'
           } >> "$GITHUB_STEP_SUMMARY"
+
+  # #351 — the BYTE-FREE claims, checked instead of asserted.
+  #
+  # A SEPARATE JOB, not another step in `firmware`, for one measured reason: this arm LINKS every
+  # tier (the check/clippy loops only compile), which on a dev box is ~70s x 10 on top of a
+  # `firmware` job already budgeted at 45 minutes. Bolting it on would have made the timeout the
+  # thing that fails, and a gate that times out gets deleted. The two jobs share no state — a
+  # different target dir and a different profile (`CARGO_PROFILE_RELEASE_DEBUG`) — so running them
+  # in parallel costs wall-clock nothing.
+  exclusions:
+    name: tier exclusions (#351)
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@v4
+      - uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: rust/clock
+          key: excl                      # its own cache: a different profile, so different artifacts
+      # Same assertion as the `firmware` job, for the same reason — and it matters more here,
+      # because without readelf this gate has no instrument at all.
+      - name: Assert readelf present
+        run: readelf --version >/dev/null
+
+      - run: tools/gate.sh excl
+
+      - name: Exclusions → job summary
+        if: always()
+        run: |
+          {
+            echo "### Tier exclusions — the BYTE-FREE claims"
+            echo '```'
+            tools/check_exclusions.py claims || true
+            echo '```'
+            echo "Derived from the \`#[cfg(feature = …)]\` on each \`mod\`, walked from"
+            echo "\`src/main.rs\`. Gate a module and it is covered; there is no list to update."
+            echo ""
+            echo "⚠️ This proves no **executable** bytes from an excluded module survived, via the"
+            echo "DWARF line table (which attributes inlined code back to its source file). A"
+            echo "consts-only module emits no line rows and is invisible — those are declared in"
+            echo "\`tools/build-matrix.toml\`'s \`[unobservable]\` with a reason."
+          } >> "$GITHUB_STEP_SUMMARY"
--- a/ONBOARDING.md
+++ b/ONBOARDING.md
@@ -149,6 +149,14 @@ Full toolchain + gotchas: **[`docs/BUILDING.md`](docs/BUILDING.md)**. TL;DR for 
    The `default` build must also stay behaviourally unaffected (prove via cfg-gating, **not** ELF
    byte-equality — `build.rs` stamps a per-commit git hash, so the default ELF changes every commit
    by design).
+
+   **Since #351 the "BYTE-FREE" claims are checked, not asserted.** `tools/check_exclusions.py`
+   links every tier and reads the DWARF line table to ask which source files actually contributed
+   code; a module the tier's features exclude must contribute none. The claims are **derived from
+   the `#[cfg(feature = …)]` on each `mod`**, walked from `src/main.rs` — so you gate a module and
+   the gate covers it, with no list to update. Two things to know: it proves no *executable* bytes
+   (a consts-only module is invisible — declare it in `build-matrix.toml`'s `[unobservable]` with a
+   reason), and a module no tier ever builds is reported **UNPROVEN**, not passing.
 3. **Version identity:** `build.rs` embeds `BUILD_HASH` + `BUILD_NUMBER` →
    `names.rs::version_name()`, a forge-realm sigil name. The number is the OTA monotonicity gate;
    the name is the boot-splash reveal. Two things to get right:
--- a/ONBOARDING.md
+++ b/ONBOARDING.md
@@ -153,10 +153,15 @@ Full toolchain + gotchas: **[`docs/BUILDING.md`](docs/BUILDING.md)**. TL;DR for 
    **Since #351 the "BYTE-FREE" claims are checked, not asserted.** `tools/check_exclusions.py`
    links every tier and reads the DWARF line table to ask which source files actually contributed
    code; a module the tier's features exclude must contribute none. The claims are **derived from
-   the `#[cfg(feature = …)]` on each `mod`**, walked from `src/main.rs` — so you gate a module and
-   the gate covers it, with no list to update. Two things to know: it proves no *executable* bytes
-   (a consts-only module is invisible — declare it in `build-matrix.toml`'s `[unobservable]` with a
-   reason), and a module no tier ever builds is reported **UNPROVEN**, not passing.
+   the `#[cfg(feature = …)]` on each `mod`**, walked from `src/main.rs`. Three things to know:
+
+   * **Ungating a module is a manifest change.** `build-matrix.toml`'s `[tier_exclusive]` lists
+     the commitment, and the source is checked against it both ways. This exists because deleting
+     a `#[cfg]` deletes the derived claim with it — the binary check went green on a real planted
+     leak until this arm was added. Regenerate with `tools/check_exclusions.py claims --toml`.
+   * It proves no *executable* bytes. A consts-only module emits no line rows — declare it in
+     `[unobservable]` with a reason.
+   * A module no tier ever builds is reported **UNPROVEN**, not passing.
 3. **Version identity:** `build.rs` embeds `BUILD_HASH` + `BUILD_NUMBER` →
    `names.rs::version_name()`, a forge-realm sigil name. The number is the OTA monotonicity gate;
    the name is the boot-splash reveal. Two things to get right:
--- a/rust/clock/Cargo.toml
+++ b/rust/clock/Cargo.toml
@@ -202,7 +202,9 @@ hostsim = ["dep:libm"]
 # Compiles the digital-`Flex` driver menu into ONE image; a runtime pin-map relayed
 # over the #56 keyed-CFG `G` channel selects which driver binds each free GPIO at
 # boot — no recompile. Default-OFF so the default/wifi/espnow builds stay BYTE-FREE
-# of it (#44): `src/io.rs` is `#![cfg(feature = "io")]` and every hook is cfg-gated.
+# of it (#44): `src/io.rs` is gated at its `mod io;` declaration (main.rs) and every hook
+# is cfg-gated. (This line said `src/io.rs` carries `#![cfg(feature = "io")]`; it does not —
+# same effect, wrong citation, corrected while #351 made the CLAIM checkable.)
 # Implies `espnow`: the pin-map's transport (the CfgTracker relay + `take_cfg_offer`
 # apply) lives in the ESP-NOW radio manager, and the fleet runs espnow end-to-end.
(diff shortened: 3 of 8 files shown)


jphein:
## Failure demonstration — and what it found

The brief asked for a planted leak. It found a hole in the gate, so this is worth reading in order.

### Take 1 — the gate was right, the demo was wrong

Planted the reference inside a `log::info!`. Gate stayed **green**. Correctly: `strings` on the default ELF finds **0** occurrences of `"smol booting"` — release builds eliminate the log calls outright, so the planted call rode a dead branch and genuinely was not in the binary. Recorded rather than quietly retried.

### Take 2 — the leak was real, and the gate still said `0 leaked`

`core::hint::black_box` on the result instead. Ungate `pub mod target;` in `net.rs`, add the call to `main()`:

```
default   11 crate files · 44 modules claimed absent · 0 leaked     ← before
default   12 crate files · 43 modules claimed absent · 0 leaked     ← after the plant
```

**12 crate files.** `src/net/target.rs` is in the default tier — the instrument saw it appear. And `0 leaked`, because **the claim it would have violated was derived from the very `#[cfg]` the regression deleted**. Claims went 44 → 43 and nothing complained.

This is the realistic regression, not a contrived one. Nobody references an excluded module — that does not compile, the module is not there. What happens is: someone drops the gate so their new call site compiles, and leaves the BYTE-FREE comment above it. **A gate that can be silenced by editing its subject is not a gate.**

### The fix (`ea0b8a1`)

`build-matrix.toml` gains `[tier_exclusive]` — the commitment as data, checked against the source in both directions. Same move #350 makes for the chip roster vs `budget.rs`, applied where it turned out to be load-bearing:

```
$ # with the gate on `pub mod target;` removed
  FAIL DECLARATION: src/net/target.rs is declared tier-exclusive (wifi) but is NOT
  cfg-gated in the source any more — it now compiles into EVERY tier. If that is
  intended, delete the [tier_exclusive] line in the same commit and fix the comment
  that still calls the lower tiers byte-free of it.
  declaration: 44 declared vs 43 derived — 1 disagreement(s); NOT reading any ELF
  exit=1
```

It runs first and returns without reading an ELF: once derived and declared disagree, the binary arm is answering a different question than the one committed to, and printing its cheerful per-tier `0 leaked` underneath would be the most misleading output this tool could produce.

### The binary arm, refusing on real artifacts

Separately proven end to end on the actual tier ELFs — hand the `fleet` binary to the checker as if it were `default`:

```
FAIL tier default: src/io.rs contributed code, but io is OFF
FAIL tier default: src/ota.rs contributed code, but wifi is OFF
FAIL tier default: src/mesh_snake/mod.rs, src/mesh_snake/snake_core.rs contributed code, but espnow is OFF
… 35 findings
```

Note the third line: the subtree case, where nobody references `mod.rs` and the child is what got linked.

### Green again

Source restored byte-for-byte (`git status` clean), and the full run:

```
10 tiers · 44 claims · 43 observed · 1 declared unobservable · 44 declared tier-exclusive, source agrees
```

`tools/test_check_exclusions.sh` is now **15 arms, 0 failed** — the four new ones cover agreement, gate removal, gate omission, and a gate changed to a different feature.

jphein:
## CI green — and it refuted one of my own numbers

All three jobs pass, and the exclusions job reproduces the per-tier counts **byte-for-byte** against the local run — same 11 / 20 / 45 / 48 / 54 / 45 / 49 / 48 / 48 / 55, no SKIPs:

```
tier exclusions (#351)      pass  2m59s
tiers + clippy + stack floor pass  1m14s   (10 check + 10 clippy + stack floor 74208 B, GATE GREEN)
host verifier suites         pass  48s
```

That reproducibility matters more than the green: the file set is derived from DWARF on a different machine with a different `board.rs`/`secrets.rs`, and it landed identical. This is a tree-independent measurement, unlike the stack number.

**The correction (`1b1606e`).** I justified the separate mode and job on cost — *"~70 s × 10 tiers"* — and CI says **179 s for the whole arm**. My figure came from a local experiment that gave every tier its own target dir and so paid ten cold *dependency* builds; the shipped path shares one dir and pays only the crate plus the LTO link per tier. Cost is therefore **not** the reason for the split — cache and profile hygiene is (a different `CARGO_PROFILE_RELEASE_DEBUG` in a different target dir would thrash the fingerprints the stack-floor build depends on). Both comments rewritten to say the true reason.

A wrong number sitting in a comment is precisely what this issue is about, so it got re-derived rather than caveated.

jphein:
**Correction to my merge note — the error is mine, and it is exactly the failure mode this PR exists to prevent.**

I wrote that `.text` at 892,022 B was *"identical with and without"* the debug flag, and called it *"proof the gate doesn't perturb the shipped image."* **The PR never claimed that.** It measured **same SIZE**. I read "identical" into a size number and restated a caveated measurement as a stronger guarantee — a claim laundering through a summary into something the evidence did not support.

The author's own numbers say otherwise:

```
fleet, line-tables-only vs no-debug — every ALLOC section same ADDR and same SIZE, but:
  .text     892,022 B   183,229 differing (20.54%)
  .rodata    71,372 B       876 differing ( 1.23%)
  .rwtext     3,672 B       156 differing ( 4.25%)
  .trap       1,126 B        40 differing ( 3.55%)
  .data / .rodata.wifi / .rwtext.wifi / .data.wifi — byte-IDENTICAL
```

**What still stands:** #351's verdict is *which source files contributed code to a tier*. Same-size-different-bytes at regular offsets across `.text`/`.trap` is the signature of **function reordering** — same code, different layout — and every ALLOC section matching in both ADDR and SIZE is strong evidence that codegen decisions are unchanged. Reordering cannot change which files contributed. The gate's conclusions are unaffected and the merge is sound.

**What does NOT stand is my sentence.** `line-tables-only` **does** perturb the image's bytes. A definitive n1/d/n2 sandwich is running to attribute the 20.54% to the flag rather than to build nondeterminism — the author flagged honestly that its original control was lost when someone edited `src/bench.rs` in the shared tree mid-run, and that a separate n/d/n control has since shown the tree reproducible, which points at the flag. If the sandwich confirms it, `gate.sh`'s header needs the same correction.

Recording this here rather than quietly fixing it: on an issue about claims not matching reality, a team lead overstating a measurement in the merge note is the finding, not a footnote. The author caught it in my words within minutes of me writing them.

jphein:
**Amendment to my correction above — the sandwich is done, and one sentence of mine fails again. Posting it because the sentence was mine.**

The n1/d/n2 sandwich (current tree, fleet tier):

```
                  n1 (no dbg)   d (line-tables-only)   n2 (no dbg)   delta
.text                  871,618              871,664        871,618     +46 B
CONTROL   n1 vs n2:        3 of 1,027,770 ALLOC bytes differ (0.0003%) — all in
                           .flash.appdesc (ESP-IDF build stamp); every section
                           same address, same size
TREATMENT n1 vs d:   464,180 of 1,027,816 ALLOC bytes differ (45.2%); .text 53.06%
                           and .text CHANGED SIZE
```

The control is clean, so the delta is the flag: **`line-tables-only` DOES perturb the shipped image.**

**My correction above reasoned "same size, differing bytes → function reordering → verdict unaffected." The size change kills that reading.** +46 B of `.text` means **codegen differs** — inlining decisions moved — and inlining is exactly what line-table attribution depends on. So the ELF arm measures a binary that is not the shipped one, and it is **unsound in the false-pass direction**.

Also for the record, the original PR sentence was **true on the tree it was measured on** (pre-#233: 892,022 B, byte-identical). It did not generalise across the dependency bump — a codegen-relevant boundary invalidates measurements nobody re-takes. Two layers of error here, neither the measurement itself: the author generalised a single-tree observation (and said so), and I upgraded it to a guarantee.

**Disposition (landing as a small follow-up PR):** `[tier_exclusive]` (the source arm) is the load-bearing proof — zero build cost, unaffected by any of this, and the arm that catches the regression that actually happens. The ELF arm **stays, demoted to corroboration**, retitled to state it measures a debug-instrumented build: its *failure* is still a true alarm; only its silence proves less than it appears to. `tools/gate.sh`'s header and the CI comment currently claim no shipped byte moves — being corrected in that PR. Verdict-level assessment: very likely still correct (no module appeared or vanished across all ten tiers; 46 B in 871 KB) — but 'very likely' is not the standard this issue set for itself, hence the demotion.

Issue #390, opened 2026-08-25
a dependency's feature unification can resize a shipping static and no gate says so


jphein:
#335 P1.1 added six embassy dep lines and changed no source. embassy-net turned on smoltcp's `async` feature; smoltcp is SHARED with the hand-driven gateway (one `v0.13.1` in the graph), so `WakerRegistration` per socket grew `net::wifi::NTP_SOCK_STORAGE` 0x540→0x580 — +64 B of `.data`, straight out of the `.stack` region, invisible to code review (no source diff), to the compiler (no arity change), to `repro_stack_check` (0.06% of an aggregate with a threshold), and to #351's exclusion checker (presence, not size). Found only by a hand `nm` A/B. The mechanism is not size-bounded — P1.3's 18,952 B task POOL is the same discover-it-by-hand problem via a different route.

**Proposal:** a per-tier symbol-size baseline in the #278/#350/#367 declaration style — `tools/check_symbol_sizes.py` + checked-in `tools/symbol-sizes.<tier>.txt` for `.data`/`.bss` symbols ≥256 B, `--bless` to update, so a resize lands as a **reviewed diff line**. It generalises the stack floor rather than duplicating it: the floor watches the one derived aggregate, this watches the individual statics whose growth moves it.

⚠️ Design constraint learned the hard way: key on **DEMANGLED hash-stripped names** — the crate-metadata hash in mangled names churns between builds (`CsjIZ8XU16pSq_` → `Csa9xoDfnccUU_`) and buries the signal in ~72 lines of noise; skip `.L_MergedGlobals*`/`.Lswitch.table.*`, which renumber every build.

**Cheaper alternative** if the tooling outweighs the value: add the socket storages to #348's `budget.rs` arithmetic with a const assert (~5 lines, no new tools — guards these statics, not the class).

Full analysis: the F1 section of the Phase-1 exec log (`scratch/embassy-port/phase1-exec-log.md`, local notes). Found during #335 Phase 1.

Issue #349, opened 2026-08-01
feat(ota): make the image self-describe its target and let the device refuse an unsuitable one (structured TargetId in the signed manifest + embedded descriptor)


jphein:
Prior-art study (OpenWrt · WLED · Tasmota) for #331/#347. Full report: `scratch/parity/multitarget-prior-art.md`.

## The gap

Today the staged line is `OTA|<build>|<size>|<sha256hex>|<url>`, ed25519-signed over `"build|size|sha256"` (`tools/ota_publish.sh`). That chain proves **authenticity and integrity**. It proves nothing about **suitability**.

Cross-*chip* is caught for free: `esp_image_header_t.chip_id` sits at image byte offset 12 and the second-stage bootloader validates it ([esp_app_format.h:93](https://github.com/espressif/esp-idf/blob/master/components/bootloader_support/include/esp_app_format.h#L93)), so a signed S3 image on a C3 boot-loops into rollback rather than bricking.

**Cross-tier and cross-board-variant images pass every check and then boot.** They are silently wrong. And `smol/ota/staged` is retained *fleet-wide* — it arms every board — which becomes a correctness bug the moment target #2 exists, independent of any device-side check.

`tools/verify_image.sh` is an **operator-side** check (a script you run before/after flashing), not a device-side guard.

## What the prior art does

**OpenWrt: the strongest guard of the three, and it is data + one checker — never code per case.** ([`lib/upgrade/fwtool.sh`](https://github.com/openwrt/openwrt/blob/main/package/base-files/files/lib/upgrade/fwtool.sh))

1. The image carries **appended JSON metadata** (`append-metadata` in the image recipe; `include/image.mk` exports `SUPPORTED_DEVICES`, `DEVICE_COMPAT_VERSION`, `IMAGE_SIZE`, ...).
2. The device reads its own identity from `/tmp/sysinfo/board_name` (device-tree `compatible`).
3. `fwtool_check_image()` intersects them. `supported_devices` is a **list**, and `DEVICE_ALT0..5_*` declare alias identities — so board renames and OEM relabels stay flashable *without touching the checker*.
4. **Two-level compat version**, about persistent state rather than identity: major mismatch -> refuse; minor mismatch -> allowed only with `-n` (don't keep config). With a human string beside it — `mt7621.mk`:
   ```makefile
   DEVICE_COMPAT_VERSION := 1.1
   DEVICE_COMPAT_MESSAGE := Config cannot be migrated from swconfig to DSA
   ```
[...]

Issue #329, opened 2026-08-01
crown own-id command path: INSTALL and cmd/reset both ignored by the crown itself (observed on a dev image)


jphein:
Observed during the 916 roll (lucid-canary-roll, 2026-08-01 ~07:10): leaf-directed commands worked 3/3 (id50/id8/id51 installed 916 via relayfetch), but **commands addressed to the crown itself (id5) did nothing** — neither the retained INSTALL nor 15 transient `cmd/reset` publishes across ~5 flush windows (boot= constant, up= climbing).

Diagnosis, reproducible twice: `install-caught=none pending=none staged_raw=916 leaf_ota=0`. Not heap, not a gate refusal, and `leaf_ota=0` rules out the retained-ghost / leaf_installs_outstanding pin. The signature is that the retained order is published then cleaned up **before the crown's self-read path runs**; leaves never hit this because the crown relays theirs promptly. Both failing paths are the own-id path; both working paths are the relay path.

**Scope caveat, load-bearing:** id5 runs a DEV image (`v913+dev.9170907`, an unknown build) — this observation says NOTHING about release 913/916 behavior yet. Crown self-OTA demonstrably worked on the 913-era crown (2026-07-28, reboot-then-arm). Next steps: (1) bench USB reflash id5 → 916 (MAC-lock E8:F6:0A:1C:27:BC; espflash otadata trap: `erase-region 0xf000 0x2000` after any OTA history, check the 'Loaded app from offset' line); (2) once id5 is on RELEASE 916, retest own-id INSTALL and cmd/reset to establish whether the race exists in shared code or was dev-image-specific; (3) if shared: the fix belongs in the retained-order cleanup ordering vs the crown self-read.

Sanctioned remedies exhausted before stopping: arm → re-arm → reboot-then-arm; HA Update button and reboot automation publish the same topics so they are not independent paths. No USB action taken. Full evidence: lucid's roll report §0 (scratch/roll-914/report.md).

jphein:
**The mechanism I filed above is REFUTED — correcting it** (static analysis by oracle, discriminator read off the broker just now).

[...]

PR #391, opened 2026-08-25
feat(fw): #335 Phase 1 — executor beside the superloop (P1.0–P1.3) [HW-CANARY-HELD]

jphein:
Phase 1 of the #335 Embassy re-platform, executed against today's `main` with `dream/feat-embassy` as a design reference (never rebased). The reference's own order — delete the synchronous stack, then build the async one — is not available on this base, because what it deletes is a hardened, shipping v917 gateway. So this inverts it: **the executor comes up beside the untouched superloop**, and later phases move traffic across. Every step keeps a shipping image and is individually revertible.

## Commits

| step | commit | content |
|---|---|---|
| — | `85546e6` | chore: the Cargo.lock the union probe left out (32 pkgs, 0 removals; landed alone so later lock deltas are attributable) |
| probe | `c00d82d` | esp-rtos `embassy` feature ON — the union (embassy + esp-radio) resolves, compiles, links |
| P1.0 | `455e866` | edition 2021 → 2024, rust-version stays 1.96 |
| P1.1 | `0ad2f8c` | embassy-executor / embassy-time / static_cell / embassy-sync / embassy-net / portable-atomic |
| P1.2 | `40a6a51` | executor ON: `#[esp_rtos::main]` + the single hand `esp_rtos::start` hoisted to `main` |
| P1.3 | `9e86030` | `Timer::after` paces the superloop (3 sites, `SUBTICK_MS = 20` stays the one definition) |

Design call of record (reviewed + approved): the executor rides **`wifi`, not `hw`** — the no-radio default tier is proven byte-identical in all four sections (.data 1904 / .bss 468 / .rodata 19132 / .stack 315388) with zero `esp_rtos`/`embassy_executor` symbols in `nm`, so #44's byte-minimal-default survives.

## ⚠️ Read the .stack trajectory correctly — the P1.3 drop is a MIGRATION, not a loss

| step | .stack | ×floor (74,208) | .bss |
|---|---|---|---|
| baseline | 106,976 | 1.44 | 156,816 |
| P1.0 | 106,976 | 1.44 | 156,816 |
| P1.1 | 106,912 | 1.44 | 156,816 |
| P1.2 | 106,864 | 1.44 | 156,864 |
| **P1.3** | **87,960** | **1.185** | **175,768** |

`nm` names the P1.3 delta: `clock::__main::__embassy_main::POOL` = 18,952 B. Real await points materialise `run()`'s live state as a future in **static task storage** — ~18.9 KB of main's frame (dominated by `RadioManager`) moved out of `.stack` into `.bss`. Real stack demand went **down** while the gate's number reads worse. Corroboration: the wifi-only tier's POOL is 1,184 B (no RadioManager); the default tier has no POOL at all. `ESP32C3_MEASURED_PEAK_BYTES = 55,656` was measured with that frame **on** the stack, so the derived floor is now conservative by ~the POOL size. `budget.rs` is deliberately untouched — re-deriving the peak needs stack-paint under live radio, on a board.

## Gates (every step)

clippy 0 errors on all 10 build-matrix tiers · hostsim suites bard 41 / budget 10 / input 8, 0 failed · `check_verifier_wiring` 19 sound + 1 known phantom · `test_verifier_wiring` 5/5 · `check_elect_send_path`, `check_byte_free`, `check_diag_budget`, `check_shed_order` all green. Also found and filed along the way: #390 (a manifest edit resized `NTP_SOCK_STORAGE` +64 B via smoltcp feature unification — caught only by `nm` A/B).

## What merging requires — this PR is HELD until all four

1. **§R1 on metal:** the `block_on` busy loop inside an executor task is preempted, not deadlocked — `BurstProbe` pre/post on the same board under the same duty (the instrument already ships on main; evidence = the `longest app gap` numbers, not "it boots").
2. **§R11 on metal:** the embassy timebase ticks at wall-clock rate (stopwatch vs the 15 s flush budget).
3. **The #335 round-trip rollback canary:** a Phase-1 image up, then roll *back* and confirm boot + re-election. The 2026-08-02 v917 roll was forward-only; this gate has never been discharged.
4. **The P1.4/P1.5 ownership decision:** `embassy_net::new` needs an owned `interfaces.station`, which the shipping gateway owns on five live borrow paths (`radio_dev.rs` → `mode.rs`); esp-radio 0.18 has one owned station, no Clone/Copy. P1.4 is therefore **absent from this PR by decision, not omission** — it folds into the P1.5 radio-ownership review (`scratch/embassy-port/p15-ownership-proposal.md`, in progress).

Full evidence trail: `scratch/embassy-port/phase1-exec-log.md` (local notes).

### DIFF
--- a/rust/clock/Cargo.lock
+++ b/rust/clock/Cargo.lock
@@ -3,6 +3,15 @@
 version = 4
 
 [[package]]
+name = "aho-corasick"
+version = "1.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c982642fa9e8606056828ee9a8505737230110bb1099153c79efe865c59d12ba"
+dependencies = [
+ "memchr",
+]
+
+[[package]]
 name = "allocator-api2"
 version = "0.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -89,6 +98,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 
 [[package]]
+name = "cc"
+version = "1.4.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0ad534f4357a5264cce5019c989cf66a4f0dc4e0d1b1d15f8aacec0ff7360273"
+dependencies = [
+ "find-msvc-tools",
+ "shlex",
+]
+
+[[package]]
 name = "cfg-if"
 version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -99,7 +118,11 @@ name = "clock"
 version = "0.1.0"
 dependencies = [
  "ed25519-compact",
+ "embassy-executor",
  "embassy-futures",
+ "embassy-net",
+ "embassy-sync 0.7.2",
+ "embassy-time",
  "embedded-graphics",
  "embedded-storage",
  "esp-alloc",
@@ -114,10 +137,12 @@ dependencies = [
  "libm",
  "log",
  "nb 1.1.0",
+ "portable-atomic",
  "sha2",
  "sigil-names",
  "smoltcp",
  "ssd1306",
+ "static_cell",
 ]
 
 [[package]]
@@ -125,6 +150,16 @@ name = "const-default"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0b396d1f76d455557e1218ec8066ae14bba60b4b36ecd55577ba979f5db7ecaa"
+
+[[package]]
+name = "cordyceps"
+version = "0.3.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5b9ab7e0ca1d179628fa0172b2b97203c7fa0cd81be2448bd446fb9559ca9261"
+dependencies = [
+ "loom",
+ "tracing",
+]
 
 [[package]]
 name = "cpufeatures"
@@ -162,6 +197,16 @@ dependencies = [
 
 [[package]]
 name = "darling"
+version = "0.20.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fc7f46116c46ff9ab3eb1597a45688b6715c6e628b5c133e288e709a29bcb4ee"
+dependencies = [
+ "darling_core 0.20.11",
+ "darling_macro 0.20.11",
+]
+
+[[package]]
+name = "darling"
 version = "0.21.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9cdf337090841a411e2a7f3deb9187445851f91b309c0c0a29e05f74a00a48c0"
@@ -178,6 +223,20 @@ checksum = "25ae13da2f202d56bd7f91c25fba009e7717a1e4a1cc98a76d844b65ae912e9d"
 dependencies = [
  "darling_core 0.23.0",
  "darling_macro 0.23.0",
+]
+
+[[package]]
+name = "darling_core"
+version = "0.20.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0d00b9596d185e565c2207a0b01f8bd1a135483d02d9b7b0a54b11da8d53412e"
+dependencies = [
+ "fnv",
+ "ident_case",
+ "proc-macro2",
+ "quote",
+ "strsim",
+ "syn 2.0.119",
 ]
 
 [[package]]
@@ -203,6 +262,17 @@ dependencies = [
  "proc-macro2",
  "quote",
  "strsim",
+ "syn 2.0.119",
+]
+
+[[package]]
+name = "darling_macro"
+version = "0.20.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fc34b93ccb385b40dc71c6fceac4b2ad23662c7eeb248cf10d529b7e055b6ead"
+dependencies = [
+ "darling_core 0.20.11",
+ "quote",
  "syn 2.0.119",
 ]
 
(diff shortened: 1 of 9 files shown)


jphein:
Executor's supplement — four review-relevant items from the exec log that the description doesn't carry. Additive; the description above is accurate as written.

### 1. Positive proof the executor LINKED, not merely compiled

Every gate in the description is a pass/fail. None of them answers "did the macro's expansion
actually reach the image?" — the failure mode where it doesn't would be silent. `nm` on the
espnow ELF gives three independent fingerprints:

```
esp_rtos::embassy::Executor::run          monomorphised over a closure from
                                          clock::__main::__risc_v_rt__main
embassy_executor::raw::util::UninitCell<clock::__main::__embassy_main_task…>::write_in_place
esp_rtos::timer::TimeDriver::arm_next_wakeup
```

and after P1.3, `embassy_time::Timer::after`, `<Timer as Future>::poll` and
`_embassy_time_schedule_wake`. Conversely the `default` tier ELF contains **zero** `esp_rtos` /
`embassy_executor` symbols, which is the other half of the byte-identical-sections claim: the
no-radio build is unchanged *and* provably contains none of this.

### 2. ⚠️ Correction to the plan of record: `#[esp_rtos::main]` does NOT call `esp_rtos::start`

PORT-SPEC §2.3 reads as though the attribute wires the scheduler. It does not. Per the macro
source (`esp-hal-procmacros-0.22.0/src/rtos_main.rs`), it only wraps the body in a `__main`
module, creates an `esp_rtos::embassy::Executor`, and spawns the body as its first task. The
scheduler start remains a hand call — which is precisely why the hoist has somewhere to land.
Worth knowing before anyone reviews P1.2 expecting the attribute to have absorbed it.

### 3. The hoist is complete by construction, not by inspection

RISKS §R14 flags the two-call-sites hoist as "easy to half-do". It is now impossible to half-do:
`WifiPeripherals` went from `{timg0, sw_int, wifi}` to `{wifi}`, so a surviving `esp_rtos::start`
call site fails to build for want of a peripheral to start with. The compiler is the gate.

Placement is also deliberate and worth a reviewer's eye: `esp_rtos::start` sits immediately
before the radio-bring-up branch rather than beside `esp_hal::init`, because that is the instant
the two old calls fired. **The hoist changes where the call is written, not when it runs** — #226
otadata init and #40 unconfirmed-boot bookkeeping still complete on a bare context with no
scheduler and no timer driver, as on v917. That declines RISKS §R7 (Phase 1 reorders boot →
breaks rollback) rather than arguing about it, and it is why merge-requirement 3 is a canary
rather than a re-derivation.

### 4. Rejected en route — binding the `Stack` to `interfaces.access_point`

Recorded so it reads as declined rather than missed, since it is the obvious-looking way to make
P1.4 fit. `Interfaces.access_point` is a genuinely unused `Interface<'d>`, so a `Stack` over it
would compile, link, spawn, and satisfy the letter of "the Stack exists, nothing uses it yet" —
over an AP that is never started. A correct-looking construct with no behaviour behind it is
[[stubbed-intentions-under-deliver-silently]] exactly, and P1.5 would have to unpick it.

---

One process note: this branch was pushed and this PR opened within about a minute of each other
from two directions. The description is the one to review; I have not modified it.

jphein:
**Merge-requirement 4 update — the ownership decision is made.** Variant C (phased-consumer) adopted 2026-08-24, **as amended by the adversarial review** (`scratch/embassy-port/p15-oracle-review.md`): Phase 1 ships without binding the station; transport + station move atomically with `mqtt_session` in Phase 3 (an atomic commit now known to span **two tiers** — `try_time_sync` on the `wifi` tier is a 7th station path); controller converts last.

Review's key inversion, worth every future reader's attention: esp-radio's `Interface` is **Copy** — nothing at the type level prevents two live stacks. The real constraint is queue arbitration (both consumers pop `data_queue_rx()` keyed only by `Station`; a second live stack steals frames **silently**). So the constraint is unenforced today, and a structural gate (declared count of `SmolWifiDevice::new` sites per tier, in the `check_elect_send_path` idiom) is now a Phase-3 deliverable.

Requirements 1–3 (BurstProbe §R1, timebase §R11, the round-trip rollback canary) remain open and all need a board — that HW session is the next gate for this PR.

jphein:
## HW gate evidence — 2026-08-24 bench session, id50 (C3, the July canary rig, MAC-guarded)

Flashed the P1.0–P1.3 canary (build 345-dev, built on familiar) over v917 via USB (otadata-erase → ota_0; **espflash 4.5 accepted the 1.1-built ELF**, closing BUILDING.md's dated open item).

| gate | verdict | evidence |
|---|---|---|
| **§R1 — block_on under the executor deadlocks?** | **DISPROVEN** | first boot: `slot=0, ota=confirmed:345` (self-test passed); WiFi associated (`ap ch6 −57`), `tsrc=ntp`, MQTT flushes `fok=1,2,3…` on 32 s cadence, zero resets across the session (`rst=usb-jtag, boot=24` stable) |
| **§R11 — embassy timebase ticks wall-clock?** | **PASS** | publish wall-deltas 33 s/32 s/32 s vs device `up` deltas 32 s/32 s/32 s; `tage` grows at wall rate; SNTP converged at up=47 |
| **BurstProbe PRE vs POST** | **PARITY (expected)** | v917 PRE: `brst=500:2399:f`, `500:2328:f` · P1 POST: `500:2032:f`, `500:2156:f`, `500:2418:f`. Worst app gap identical (500 ms) — correct for Phase 1: the flush is still synchronous inside the one task (Variant C moves transport in Phase 3, which owns the gap-crushing claim). What parity proves: **the executor costs nothing measurable** — same gap, same heap floor (27,368 B), same cadence |
| heap | +0 steady-state | 27,368 B free on both images (the 37.6 KB boot reading was pre-radio-init) |

**Still open for merge:** the OTA round-trip rollback canary. The USB 917→345 flash is not the OTA path, and with a single live board the OTA variant needs either a second board or the #329 fix (a crown ignores its own-id INSTALL on dev images). Remains the one HW requirement outstanding.

Board state: id50 left running the P1 canary (all-DEV fleet; `staged_raw=917` retained but `install_requested=0` — no auto-pull-back). Full capture: `scratch/embassy-port/` session logs (local).

jphein:
**🔴 HW-gate finding update (23:41): the canary PANICKED once on first-peer-join — then self-recovered and has been stable under full crown duty since.**

Timeline: id50 ran ~2h solo, zero resets (the earlier comment was accurate when written). At ~23:05 JP powered two fleet boards (id8, id51, v917). At **~23:09 id50 reset with `rst=panic` (boot 24→25)** — the first time the executor image ever executed the peer-join/relay path with real leaves. It rebooted, **self-test passed (`ota=confirmed:345` — the app-side rollback correctly did NOT fire)**, re-took the crown (`MC|50|6`), and has since run **32+ minutes of real crown duty**: relaying leaf telemetry (`relay 008 ok`), serving LDBG exchanges between id8/id51, `tsrc=mesh`, heap steady ~26.4 KB.

What this means for the merge gates: §R1 (solo executor duty) stands disproven-for-deadlock as before, and the recovery machinery is now ALSO proven (panic → clean reboot → self-test → resume duty, no human touch). But **a panic in the peer-join window is a real defect to root-cause before Phase 3** — the failing path is one of: first HELLO/roster insertion under the executor, relay-window open on leaf join, or the LDBG serve path. Release images are serial-silent, so forensics need either an ESP_LOG=info canary rebuild + JP-side monitor, or a deliberate repro (power-cycle a leaf while a logging canary holds crown). Filed into the merge-requirements: **add (5) root-cause the peer-join panic.** The full capture is in the session's p1-crown-duty log.

jphein:
**Defect-1 re-attribution (04:22): the peer-join panic is MAIN's, not this branch's.** id50 was USB-restored to build 918 (current main) to end the canary — and panicked with the identical signature within ~60 s of rejoining the mesh, then recovered and resumed duty (now `ota=confirmed:918`, coordinating id51's 921 install). Filed as its own bug with the fleet-wide corroboration (id8's boot=284). **Merge requirement 5 is therefore re-scoped: this PR must not REGRESS the panic (it doesn't — same behavior both images), and the root-cause moves to the new issue.** The full overnight canary record: forward 917→918 ✓ confirmed · retreat 918→920 (real v917-era code, OTA) ✓ confirmed + re-elected · forward 920→921 in flight · the pre-#349 retreat gap is #401 · the stage-identity footgun is #400 · the post-panic channel-0 announce is #403. The #335 round-trip gate is DISCHARGED within the post-#349 era, which is the era that matters for Phase 3 retreat.

jphein:
**Merge policy set (JP, 2026-08-25): this PR merges immediately after #404's root-cause lands.** All HW gates are discharged (timebase, no-deadlock, round trip, panic re-attributed to main); the only reason to wait is so the executor image never wears main's bug in the fleet's eyes. The #404 repro canary is on the bench RIGHT NOW (120s resync + ESP_LOG=info) — the backtrace may land within the hour.

Issue #363, opened 2026-08-01
gate.sh's fleet tier lints RED locally but green in CI — two different failures on two machines


jphein:
**The fleet tier lints RED on a provisioned developer machine while CI reports it green.** Two independent observations tonight, with *different* errors — which is itself the finding: whatever the cause, local and CI are not linting the same thing.

**Observation A (nebula):** `cargo clippy --release --features espnow,cast,io -- -D warnings` — exactly what `gate.sh` runs — fails with 3 dead-code errors on `WEATHER_*` in `board.rs`. That file is **git-ignored and generated by `tools/ci_provision.sh`** (#359), and it declares those constants under `#[cfg(espnow)]` while their consumers need `wifi`. Any provisioned machine therefore lints red on a tier the gate claims to cover.

**Observation B (team-lead), same command, current main `d18eb1c`, fresh `CARGO_TARGET_DIR`:** fails *earlier and differently* — `cannot find module or crate esp_radio` and `unresolved import embassy_futures`. That looks like a dependency-resolution difference following the #233 matched-set bump (`b2537c4`), not a dead-code issue at all.

**Why this matters beyond the individual errors:** #338 exists so the tier gate is real rather than prose, and #343 extended `clippy -D` to every tier. If the same command produces green in CI and two different failures on two developer machines, then **the gate certifies CI's environment, not the codebase** — and a developer cannot reproduce what the gate checks. That is the same defect shape as the stack floor being provisioning-dependent (#348), one layer up.

**Suggested investigation, in order:**
1. Determine whether `ci_provision.sh`'s generated `board.rs` differs materially from a real one in cfg-gating (observation A's mechanism) — if so, the generator is producing a file that cannot lint clean.
2. Determine why a fresh local build cannot resolve `esp_radio`/`embassy_futures` when CI can (observation B) — stale lockfile, registry state, or a feature only reachable via the canonical tier.
3. Whichever way it resolves, make `gate.sh` runnable to the same verdict locally and in CI, or state explicitly in its header which arms are CI-only and why. A gate a developer cannot run is a gate they will stop trusting.

Attribution caveat, stated because it was inferred rather than proven: nebula's observation A is believed to predate its own commit (`e231d8c`), on grep-proven independence — a pristine rebuild timed out before it could be confirmed directly.

jphein:
**Root cause found for observation A. Observation B does not reproduce, and the two were never the same phenomenon.**

## A: the gate lints a file that only exists in CI

[...]

Issue #306, opened 2026-07-28
DIAG record is ~33 B from a silent-death cliff — the whole record stops publishing, not one field


jphein:
## Symptom
The per-node DIAG record measures **462 B worst case (~479 B on the wire) against a 512 B buffer**. `encode_publish` returns `None` on overflow, so an over-long record **does not lose its tail — it stops publishing entirely.**

**With no board on USB, DIAG is the fleet's only observability channel.** So the failure mode is: a healthy board goes silent and **looks dead**. Diagnosing "board absent" is exactly what this record exists to prevent.

> aurora's line, which is the whole issue in one sentence: **"the next field of this size is the one that breaks it."**

A separate, already-observed symptom of the same tightness: one measured **leaf** record offered 307 B, sent 232, and **75 B vanished with no signal of any kind** — which is why `cut=<bytes>` now spends 8 B of the short budget announcing truncation.

## Why it survived
Fields are appended **last** (deliberately — it keeps a positional parse of the fixed block intact as the record grows), so every addition lands in the danger zone and none of them individually looked expensive. `brst=` is ~16 B; `blrev=` similar. Each was locally reasonable. Nothing measured the total until now.

## The smallest correct fix
aurora is working the acute fix. **This issue should outlive that fix**, because its real content is the constraint, not the patch:

- **The record must not grow without a budget check.** Adding a DIAG field is a change to a near-full buffer, not a free annotation.
- **Overflow must degrade, not disappear.** Losing the tail with a `cut=` marker is strictly better than publishing nothing — a reader that cannot see a field must at least be able to tell it was cut.
- **Consider making the appended-last block optional-by-priority** so the fields a human is currently debugging survive a squeeze.

## Notes for whoever picks this up
- The field inventory (and which fields are *always* vs *conditionally* present) is documented and **dated** in [`docs/protocol.md`](docs/protocol.md) — "the appended-last block". Re-date it if the record shape changes.
- Per [`docs/DOC-UPKEEP.md`](docs/DOC-UPKEEP.md) §2, a parser must treat an absent `brst=`/`blrev=` as **unknown, never 0**.

jphein:
## Correction — the tighter cap is a *different* one, and it is already breached

The title understates this. `6133675` found a second, lower cap that matters more:

| Path | Cap | State |
|---|---|---|
| **Gateway → MQTT** | ~490 B publish / 512 B buffer | 462 B worst case — **~33 B headroom**, tight but not breached |
[...]

PR #392, opened 2026-08-25
fix(ota): a deaf leaf's armed install no longer locks the crown out of its own OTA (#381)

### LINKED ISSUE #381: ota: a deaf leaf's armed install silently blocks crown self-OTA — uncapped pre-relay retry holds the gate
Found live during the 917 roll (2026-08-02), root-caused from source by morpheus-elect-announce.

**A permanently-deaf leaf with an armed (retained) install silently suppresses the crown's own self-OTA — indefinitely.**

Mechanism, all by design individually:
- A pre-relay failure (`MacUnknown`/`FetchFailed`, `!reached_leaf()`) retries **uncapped** (`leaf_ota_fetch_retries`, `u8` saturating) — deliberate, the #134 fix: a pre-relay failure is gateway-local and the retained install must survive for the next crown (#111 crown-portability). `LEAF_OTA_MAX_RETRIES` bounds only the post-handoff path.
- But the un-clearable arm holds `leaf_ota_pending=1` (session state) and `leaf_installs_outstanding=1` (retained topic), and the crown's self-install gate is `!leaf_ota_pending && !leaf_installs_outstanding` ("gateway updates itself LAST").
- So while any deaf leaf holds an armed install, **the crown skips its own update — not fails, SKIPS**, producing the same quiet non-event as a healthy up-to-date crown. A future roll reports success with the crown still on the old build.

Observed: crown id5 at `mac-unknown retry=35+` for deaf id50; `leaf_ota_pending=1` survived clearing the retained topic (session state clears only on a terminal outcome or crown reboot). Not a leak (fixed-size state, no relay resources) — a **lock**.

Mitigations tonight: id50's retained install cleared; id51's never armed. Standing rule until fixed: **do not leave installs armed for boards known to be deaf.**

Fix directions (pick at design time, not here): a pre-relay retry deadline that releases only the *self-OTA gate* while keeping the retained install (preserves #134/#111); or the self-OTA gate consulting freshness of the pending leaf; or an operator-visible "crown self-OTA suppressed by pending=<id>" armdiag field so the skip is at least loud.

jphein:
Closes #381.

## The failure scenario

A permanently-deaf leaf with an armed install locks the crown out of its **own** OTA, indefinitely.

1. A leaf goes deaf (id50 on the FT roam SSID, 2026-08-02). Its retained install stays armed.
2. Every flush re-surfaces that install. The crown arms a relay → `note_leaf_ota_armed()` sets
   `leaf_ota_pending = 1` → the relay fails **pre-handoff** with `MacUnknown` (the crown cannot
   even address the leaf).
3. A pre-relay failure is **never terminal**, so `record_leaf_ota` never clears the latch. And the
   leaf never installs, so the **retained** gate `leaf_installs_outstanding` never drops either.
4. The crown's self-install gate is `!leaf_ota_pending && !leaf_installs_outstanding`. Both are
   pinned ON. Forever.

The damage is in *how* it fails. `do_install`'s `&&` short-circuits **before**
`take_install_request()`, so the crown's own command is preserved and nothing fails, logs, or
retries — it **skips**. That is byte-for-byte the same non-event as a crown that is already up to
date. **A fleet roll reports success with the crown still on the old build.**

Observed: crown id5 at `mac-unknown retry=35+` for deaf id50, own install silently skipped
throughout. `leaf_ota_pending=1` survived clearing the retained topic by hand, because the latch
is RAM and only a terminal outcome clears it.

## What this does — and what it deliberately does not

**It does not cap the pre-relay retry.** #381's source analysis is explicit that capping it
re-introduces the #134 bug — a fetch-broken crown burning an order a healthy successor would have
finished — and that the retained install must survive for the **next crown** (#111
crown-portability). The retry stays uncapped; the retained install is never touched. What expires
is only that install's **veto** over the crown's own update, after `LEAF_OTA_SELF_GATE_RETRIES = 3`
consecutive pre-relay failures (~90 s at the ~30 s flush cadence).

This is **#195's shape, mirrored**. There, a crown whose *own* self-fetch keeps failing stops
re-triggering while deliberately keeping the retained arm (`self_ota_fetch_capped`). Here, a crown
whose *relay* keeps failing pre-handoff stops suppressing itself, while deliberately keeping the
retained install. Both bound a hammer without destroying an order; both are 3.

**Releasing only the RAM latch would not have worked** — the deaf leaf pins the retained gate too,
which is exactly why the 2026-08-02 mitigation had to clear the topic by hand. Both release
together, and the verifier asserts it, because that is the obvious mistake to make here.

### The subtle half

`record_leaf_ota` now also resets `leaf_ota_fetch_retries` on the **post-handoff** branch. The
counter is shared across leaves and it releases a single shared gate. Had it only reset on a
terminal outcome, a stale count from deaf leaf A would still read as "stalled" while a **healthy
relay to leaf B was mid-flight** — and the crown would reboot into its own install underneath it.
Resetting on `reached_leaf()` makes the counter mean *"consecutive pre-relay failures with no
successful handoff in between"*, which is the only condition under which holding the crown back
buys nothing.

## Shape

The decision moved out of `main`'s expression into a new pure module `net/otagate.rs` — the
`cfgsched` precedent: extracted so the policy can be host-tested, *because the failure mode is
invisible on the board*.

`experiments/381_gate_verify` `#[path]`-includes it and asserts the truth table:

- the #40 order still holds below the threshold (armed relay / outstanding install / both);
- both gates release **at** the threshold, at the observed `retry=35`, and at `u8::MAX`;
- the release is **monotone** to `u8::MAX`, so a future "window" refactor cannot resurrect the
  lock at high retry counts — the exact regime where the bug was observed.

**Mutation-tested in both directions before being trusted** (`[[gate-that-cannot-fail]]`): forcing
the veto to never expire fails at the #381 assertion (`main.rs:48`); forcing it to always expire
fails at the #40-order assertion (`main.rs:30`). Restored source md5 verified identical.

## Observability

`sog=` gains bit 3, and it earns its place: after this fix the crown installs with bits 0/1 **still
set**, so without bit 3 the DIAG record would read "held" at the exact moment the crown stopped
being held — the same read-it-backwards trap the field was added to prevent, in the other
direction. `sog=f` is the #381 state: armed, both legacy gates set, veto expired, installing anyway.

Now formatted hex — 0-f is still **one character**, so the field width and its place in the #306
budget are unchanged (margin 22 B, unchanged). The digit is also built from the accessors rather
than the raw fields now, so it cannot drift from the gate it claims to explain.

## #329 interaction — checked, none

#329 is the own-id order **never arriving** (`install-caught=none`; its own diagnosis explicitly
rules out the `leaf_installs_outstanding` pin, and notes `leaf_ota=0`). #381 is an order that
arrives and is then gated forever (`install-caught=50`). Upstream vs downstream of the same gate;
neither fix affects the other.

## Evidence

Built on `familiar`. Fleet tier `espnow,cast,io`.

| gate | result |
|---|---|
| `cargo clippy --release`, all 10 `build_matrix` tiers | **0 errors**; warning count back to the baseline **3** (local-only `board.rs` `WEATHER_*`, #363) |
| hostsim suites | **bard 41 / budget 10 / input 8**, 0 failed |
| host verifier suites | **19/19 pass**, including the new `381_gate_verify` |
| `check_verifier_wiring.py` | 21 includes / 18 verifiers, **20 SOUND** (was 20/17, 19 sound) — `otagate` is reachable from `main.rs`, i.e. genuinely compiled, not a phantom |
| `test_verifier_wiring.sh` | 5/5 arms proven able to fail |
| `check_elect_send_path` / `check_byte_free` / `check_diag_budget` / `check_shed_order` | all exit 0 |

`readelf -SW`, A/B against an unmodified build of the same base commit:

| section | baseline | this PR | Δ |
|---|---|---|---|
| `.stack` | 106,472 | 106,472 | **0** (1.43× the 74,208 B floor) |
| `.bss` | 157,928 | 157,928 | **0** |
| `.data` | 15,008 | 15,008 | **0** |
| `.rodata` | 68,764 | 68,772 | **+8** (the `{:x}` formatter) |

The fix costs the fleet image eight bytes of `.rodata`.

## Not verified

**No board, no broker, no flash was touched.** The behavioural proof needs a deaf leaf holding an
armed install and a crown with a pending self-install: `sog` should read `3` (held), then `f`
(released) roughly 90 s later, and the crown should install. Ground truth per
`[[smol-ota-ground-truth-hierarchy]]` — an MQTT flip **to a new value**, never a retained topic's
persistence.

### DIFF
--- a/experiments/381_gate_verify/Cargo.toml
+++ b/experiments/381_gate_verify/Cargo.toml
@@ -1,0 +1,15 @@
+# #381 host guard for the crown SELF-OTA gate (net/otagate.rs). The firmware crate is
+# no_std/riscv32imc and cannot host-test, so this `#[path]`-includes the REAL otagate.rs (no drift)
+# and asserts the truth table: the #40 "gateway updates itself LAST" order still holds for a
+# healthy relay, a permanently-deaf leaf's armed install loses its veto over BOTH gates, and the
+# threshold is where it says it is. Permanent guard for #381.
+[package]
+name = "self_ota_gate_verify"
+version = "0.0.0"
+edition = "2021"
+publish = false
+[[bin]]
+name = "self_ota_gate_verify"
+path = "src/main.rs"
+[profile.dev]
+opt-level = 0
--- a/experiments/381_gate_verify/src/main.rs
+++ b/experiments/381_gate_verify/src/main.rs
@@ -1,0 +1,94 @@
+//! #381 crown SELF-OTA gate host guard. `#[path]`-includes the REAL `net/otagate.rs` (no drift)
+//! and pins the bug found live on the 917 roll (2026-08-02): a permanently-deaf leaf's armed
+//! install held BOTH self-install gates on indefinitely, so the crown SKIPPED its own update —
+//! indistinguishable from being up to date, so the roll reported success with the crown on the old
+//! build. `cargo run` — panics on failure.
+//!
+//! The two properties that must not drift apart:
+//!   1. the #40/#3 order survives — a HEALTHY relay still makes the crown wait;
+//!   2. an install that has repeatedly failed BEFORE reaching a leaf stops being a reason to wait.
+//! A change that satisfies only (1) restores the lock; only (2) lets the crown reboot into its own
+//! install mid-relay. Both are asserted here.
+
+#[path = "../../../rust/clock/src/net/otagate.rs"]
+mod otagate;
+
+use otagate::*;
+
+fn main() {
+    // ---- 1. Nothing pending: the ordinary case, the crown installs -----------------------------
+    assert!(
+        crown_may_self_install(false, false, 0),
+        "an idle crown with no leaf install must run its own"
+    );
+
+    // ---- 2. The #40 order survives — a HEALTHY relay still blocks -------------------------------
+    // Below the threshold every combination of the two legacy gates must still suppress. This is
+    // the half that a naive "just release the gate" fix would break, letting the crown reboot into
+    // its own install while a relay it armed is genuinely in flight.
+    for n in 0..LEAF_OTA_SELF_GATE_RETRIES {
+        assert!(
+            !crown_may_self_install(true, false, n),
+            "armed relay (retry={n}) must still suppress the crown's own install"
+        );
+        assert!(
+            !crown_may_self_install(false, true, n),
+            "outstanding retained install (retry={n}) must still suppress"
+        );
+        assert!(
+            !crown_may_self_install(true, true, n),
+            "both gates set (retry={n}) must still suppress"
+        );
+    }
+
+    // ---- 3. #381: a deaf leaf's install loses its veto over BOTH gates ---------------------------
+    // Releasing only the RAM latch is not enough and this asserts why: a deaf leaf pins
+    // `leaf_installs_outstanding` through its RETAINED topic, which is what forced the 2026-08-02
+    // mitigation to clear that topic by hand. Both must release together or the crown stays stuck.
+    assert!(
+        crown_may_self_install(true, true, LEAF_OTA_SELF_GATE_RETRIES),
+        "at the threshold the veto expires and the crown proceeds despite both gates"
+    );
+    assert!(
+        crown_may_self_install(true, true, 35),
+        "the observed id5/id50 state (retry=35) must not hold the crown"
+    );
+    assert!(
+        crown_may_self_install(true, true, u8::MAX),
+        "the saturated counter must not wrap back into suppression"
+    );
+
+    // ---- 4. The threshold is exactly where the constant says --------------------------------------
+    assert!(
+        !leaf_veto_expired(LEAF_OTA_SELF_GATE_RETRIES - 1),
+        "one short of the threshold still vetoes"
+    );
+    assert!(
+        leaf_veto_expired(LEAF_OTA_SELF_GATE_RETRIES),
+        "the threshold itself releases"
+    );
+
+    // ---- 5. The release is monotone in the retry count ------------------------------------------
+    // Guards against a future "window" refactor (release between N and M, suppress again after)
+    // that would make the lock reappear at high retry counts — the precise regime where the bug
+    // was observed. Once released, always released, for every gate combination.
+    for n in LEAF_OTA_SELF_GATE_RETRIES..=u8::MAX {
+        assert!(
+            crown_may_self_install(true, true, n),
+            "retry={n} is past the threshold and must stay released"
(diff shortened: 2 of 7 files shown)


jphein:
⚠️ **Do not read the diffstat above, and do not merge this yet — `origin/main` is 14 commits behind local `main`, so this PR is displaying other agents' unpushed work as if it were mine.**

- `origin/main` = `34b12f9` ("docs(diag): a missing `elect=` means SHED…", #278)
- local `main` = `185c7d7` — **14 commits ahead**, none of them pushed
- I branched from local `main` (`3046f00`), as briefed

**This commit is 6 files, +278/-9** — `b6f7c71`, and that is the entire change:

```
experiments/381_gate_verify/Cargo.toml   |  15 +
experiments/381_gate_verify/src/main.rs  |  94 +
rust/clock/src/main.rs                   |  11 +-
rust/clock/src/net.rs                    |   9 +
rust/clock/src/net/mode.rs               |  60 +-
rust/clock/src/net/otagate.rs            |  98 +
```

Everything else GitHub is attributing to this PR (`docs/protocol.md`, `rust/sigil-names/*`,
`tools/gate.sh`, `budget.rs`, `build.rs`, `ONBOARDING.md`, …) belongs to #389 / #384 / #387 / #349
and friends. **Merging as-is would land those 14 commits too**, which is not a decision this PR
should be making on their authors' behalf.

**PR #391 has the identical problem** — 15 files / +923/-135 displayed, for a 6-file change.

**The fix is one command and it is not mine to run:** push local `main` to `origin`. Both PRs then
collapse to their own commits with no rebase and no history rewrite. I have deliberately **not**
rebased onto `origin/main` — the evidence in the description was measured against `3046f00`, and
re-basing would quietly change the tree those numbers describe.

jphein:
**A live validation rig for this PR now exists on the bench (2026-08-25 04:45).** During the #335 round-trip canary's final leg, id51 (running the 920 retreat image) went mesh/WiFi-quiet after the overnight crown instability, leaving its 921 install armed and the crown (id50, build 918 — WITHOUT this fix) in exactly this issue's condition: `smol/51/ota/diag mac-unknown retry=44 src=gw` and climbing (uncapped), `armdiag leaf_installs_outstanding=1` — the crown's own self-OTA gate held hostage by an unreachable leaf's armed install, live. When this PR is reviewed+merged, the acceptance test writes itself: roll the crown to a build carrying the fix, re-arm the same state, and watch the veto expire after 3 consecutive pre-relay failures (~90 s) while the retained install and #134's retry semantics stay intact. The bench state is being left as-is for that purpose (all-DEV).

PR #402, opened 2026-08-25
feat(targets): the esp32c6 watch, in-tree — subtree import with full history (#347)

jphein:
JP's directive (2026-08-24): **"get esp32c6 watch full target of smol with all features usable."** This PR is the move itself — the watch repo imported at `targets/c6-watch/` via `git subtree`, **full history preserved**, because the watch's measured-decision record (its #75 heap forensics, the build-sigil provenance chain, the texture-ceiling gate's derivation) is part of what is being imported, not packaging to discard.

## What this is and is not

- **Is**: the watch's `main` @ `102dda0`, whole tree, under `targets/c6-watch/`. Its own `[workspace]` root stays intact — deliberately, per the #347 record: cargo feature-unification means this tree and `rust/clock` must never join one workspace graph.
- **Is not**: smol-core consumption (Phase 2, unstarted), re-predication onto the shared `board-*`/`has-*` layer from `feat/347-depin`, or any change to `rust/clock`. Additive only; nothing outside `targets/c6-watch/` is touched.

## Acceptance evidence (the goal is a property you measure)

- **First in-tree build passed on familiar**: `fambuild build --release --features story` from `targets/c6-watch/` → `build sigil: Ironclad Mold · 9ed274c`.
- **Provenance semantics changed exactly as intended**: `tools/build_hash.sh` resolves the ENCLOSING git root, so in-tree images stamp SMOL's tree hash. That is the post-move truth — an image built from this tree is identified by this tree. (This is the same hash-handoff pattern smol#394 is adopting for the familiar/Syncthing no-`.git` builds; converging on `build_hash.sh` as the one implementation is the natural follow-up.)
- **Full preflight matrix from inside smol's tree is running now** (9 feature combos, stack-floor + ROM + WSIGIL-vs-tree gates, built on familiar); result will be posted as a PR comment. Green matrix = "all features build with their measured safety margins from inside smol" — the first measurable increment of *all features usable*.

## What the imported tree brings that smol's CI should eventually own (#347 follow-ups, not this PR)

- The 256-`SceneTexture` ceiling gate (host-side, `tools/lunameter/measure.sh`) — a measured 10/10-reboot cliff turned into a grep
- `WSIGIL:` image-vs-tree staleness gate + never-ship marker refusal (`tools/preflight.sh`, `tools/watchctl`)
- Measured stack floors per feature combo, with the radio-up-soak measurement recipe
[...]

Issue #388, opened 2026-08-25
feat: fifth fleet target — the NM-CYD-C5 (ESP32-C5, dual-band Wi-Fi 6 + 802.15.4)


jphein:
JP, 2026-08-24: the fleet taxonomy gains a **fifth** target — the **NM-CYD-C5**, a
"Cheap Yellow Display"-class board built on the **ESP32-C5**. It arrived today running vendor
firmware (`esp-claw v0.3.2`) and a full factory backup exists, so it can be wiped freely. JP's ask
is two-fold: **a smol mesh member, and a Zigbee bridge** — the fleet's first 802.15.4 capability.

**Board facts** (authoritative, from the `cyd-c5` session which owns the hardware):
ESP32-C5 **silicon rev v1.0**, single RISC-V core @240 MHz, **384 KB HP SRAM + 8 MB PSRAM**,
**16 MB flash**, dual-band 2.4/5 GHz Wi-Fi 6 + BLE 5 + **IEEE 802.15.4**.
Display 2.8" 320×240 **ST7789** over SPI (SCK 6, MOSI 7, MISO 2, CS 23), **XPT2046** touch on the
same SPI bus, backlight GPIO 25, **WS2812 on GPIO 27**.

> Note for anyone copying specs forward: the core is **RV32IMAC**, not RV32IMAFC — no hard-float.
> esp-hal's target is `riscv32imac-unknown-none-elf` (contrast the P4's `riscv32imafc`), and the
> datasheet says RV32IMAC. It *does* have real atomics, unlike the C3's `imc`.

Taxonomy of record (`rust/clock/src/net/profile.rs`, #352): C3 OLED · C3 SuperMini · C6 Watch ·
S3 Ember (#331) · **+ C5 CYD**.

**Ownership:** the `cyd-c5` session owns the board and its firmware; this repo owns the tree, this
issue, and the fleet id allocation.

---

## Why this one is different from #331 — the economics invert

#331's S3 is Xtensa: smol's crate cannot compile for it at all, so *keeping ESPHome* is the cheap
phase 1. **The C5 is the opposite.** It is RISC-V, and the dependency set `rust/clock/Cargo.toml`
**already pins** carries it:

| crate | smol's pin | `esp32c5` present? |
|---|---|---|
| `esp-hal` | `1.1` | ✅ (≤1.1.1, 2026-05-07) |
| `esp-radio` | `0.18` | ✅ (0.18.0, 2026-04-16) |
[...]

Issue #352, opened 2026-08-01
refactor: keep the board-variant axis at runtime (Tasmota templates) — one BoardProfile, and derive the HA model string instead of hand-writing DEVICE_EXTRAS_*


jphein:
Prior-art study (OpenWrt · WLED · Tasmota) for #331/#347. Full report: `scratch/parity/multitarget-prior-art.md`.

## The finding: smol already picked the right pattern for this axis

Of the three reference projects, **Tasmota is the only one that got the board axis right, and smol independently landed on the same design.**

Tasmota splits its axes by *mechanism*:
- **feature set** -> compile-time (`-DFIRMWARE_LITE` -> ~145 `#undef USE_*` lines)
- **board pinout/peripherals** -> **runtime**, zero binaries. A [Template](https://tasmota.github.io/docs/Templates/) is a JSON string pasted into the web UI and applied on reboot with no recompile:
  ```json
  {"NAME":"Example","GPIO":[416,0,418,0,417,2720,0,0,2624,32,2656,224,0,0],"FLAG":0,"BASE":45}
  ```
  Thousands of community templates live in an out-of-repo database. **The board axis costs Tasmota nothing in the build system.**

smol's C3 OLED vs C3 SuperMini discrimination is a Tasmota template in miniature: one image, I2C NACK at boot -> `crate::headless()`. And #72's `io.rs` is the general form — its own `Cargo.toml` comment calls it *"ESPHome inverted, fat image / thin runtime config"*, with a runtime pin-map relayed over the #56 keyed-CFG `G` channel.

**The counter-example is decisive.** WLED left the board axis in the build system, and its maintainers have said in writing that it does not scale — [issue #2953](https://github.com/wled/WLED/issues/2953), still open after three years:

> *"We cannot add each newly working board to platformio.ini, as this would make the file unreadable, slow, and also impossible to maintain over several years. Also there is sometimes very little change like just different `board=` line so I would not create a new official buildenv just for that."*

The stated alternatives were *"reject most board envs"* or *"accept everything, and clutter up platformio.ini until it becomes unreadable."* Neither proposed fix has landed. The file is now 1,182 lines and 45 envs.

## Proposed

[...]

Issue #347, opened 2026-08-01
epic: extract the Bard to its own firmware + repo (bard.realm.watch) — and unblock Embassy by doing it


jphein:
JP's call, 2026-08-01: extract the Bard into its own firmware, its own repo, and its own Pages site — `bard.realm.watch`, mesh-capable via a shared core, with the feature flag retained in smol.

## Why this is the unblock, not just tidiness

The #233/#198 Embassy re-platform is blocked on DRAM by **6,720 B** (measured peak 55,656 B → floor 74,208 B vs a 67,488 B region — see #335). Removing the Bard from the canonical image is not a squeak past the line:

| build | stack region |
|---|---|
| #233 canonical **with** bard | 67,488 B ❌ |
| #233 canonical **without** bard | **106,560 B** ✅ |

That clears the re-derived floor by ~32 KB — and the 55,656 B peak *included* three narration cycles, so a Bard-free peak is lower still and the margin is wider. **"Extract the Bard" and "unblock Embassy" are the same action.**

It also fixes a category error: the Bard costs ~285 KB of flash (model blob in `.rodata`) and ~67 KB of `.bss` on a chip where the radio stack fights for single kilobytes. A storyteller and a mesh gateway have different appetites; one binary forces one to starve the other.

## Phasing — the unblock is cheap, the extraction is the follow-through

**Phase 0 — unblock Embassy (one line, immediate).** Drop `bard` from `REPRO_FLEET_FEATURES` in `tools/repro_build.sh` so the canonical fleet image no longer carries it. Verify by building the #233 spike branch and reporting the stack number against the floor. **Keep the `bard` feature in `Cargo.toml`** — it costs nothing when off and preserves combined builds on a chip with room. This alone unblocks #335/#233 and can land today, independently of everything below.

**Phase 1 — the project exists.** New repo `bard.realm.watch` under `~/Projects/`, GitHub repo, Pages site. Per the house convention: realm-sigil versioning with a `/api/version` endpoint, registered in `status.realm.watch/checks.json`, favicon, dark/light via `prefers-color-scheme`. Pages should show generated stories — the artifact is the point.

**Phase 2 — `smol-core`.** Extract the shared layer: SMOLv1 wire format, election, OTA relay logic, names/sigil, telemetry/DIAG formats. Both smol and the Bard firmware consume it. Note the esp32c6-watch already hand-copied `names.rs`/`sigil.rs` — that duplication is live divergence risk today, so this extraction pays for itself across three consumers, not two.

[...]

PR #405, opened 2026-08-25
feat(chip): the chip de-pin — per-chip feature arms, a `checks` rung, and the C6 compiling clean (#347 Part 2, #331, #388)

jphein:
Closes nothing on its own — this is **#347 Part 2**, the chip de-pin, and it is the part that makes #331's and #388's board work possible without re-pinning what it just un-pinned.

Builds on three commits already on this branch (Part 1's measured C6 budget row, the single-naming seam, the MSRV fix) and adds two.

## The acceptance question, and where it now lives

> A target is done when **its own feature matrix passes from inside smol's tree.**

Until this PR there was no way to *ask* that. `tools/gate.sh` crosses the canonical chip with every tier; CI's job matrix contains only chips with `builds = true`. Both correct, neither compiles a C5 or a C6 — so the chip arms were verified the way they were added: by hand, once, by whoever held the context, with the result written into a commit message. That is exactly the shape #350 ended for tiers, one axis over.

`tools/check_chips.sh` is where that matrix lives now. It derives everything from `tools/build-matrix.toml` — no chip, triple, toolchain or feature string is hardcoded in it:

```
esp32c3   riscv32imc-unknown-none-elf   — compiles clean
esp32c6   riscv32imac-unknown-none-elf  — compiles clean
esp32c5   riscv32imac-unknown-none-elf  — fails as declared (2 errors)
esp32s3   xtensa-esp32s3-none-elf       — fails as declared (6 errors)
chips: 4 as declared · 0 wrong · 0 skipped · 4 actually run          exit 0
```

## THE ESP32-C6 NOW COMPILES CLEAN

Zero errors on the canonical fleet tier for `riscv32imac`. The only output is three pre-existing `dead_code` warnings for unused `WEATHER_*` consts in the git-ignored `board.rs`.

It took **one line of shared source**. Before it: 2 errors and nothing else — TSENS resolved, the radio stack resolved, `esp-wifi-sys` resolved. `src/ota.rs`'s `reset_reason_token()` matched `SocResetReason::SysClkGlitch | CorePwrGlitch`, and that enum has **three shapes, not two** (measured, esp-hal 1.1.1):

| chip | glitch variants | marker |
|---|---|---|
| C3, S3 | `SysClkGlitch` 0x13 **and** `CorePwrGlitch` 0x17 | `has-reset-glitch-split` |
| C5 | `PowerGlitch` 0x19 only | `has-reset-glitch-unified` |
| C6 | **none at all** | *neither* |

A chip selecting **neither** is legal and meaningful: the arm compiles out and a glitch reset falls through the existing `Some(_) => "other"`, which is honest because the C6 does not distinguish it in hardware. Predicating on the capability rather than `cfg(feature = "esp32c6")` is the point — "has no glitch detector" is a fact a fifth chip may share and should not have to be rediscovered.

This is also what turns the `has-*` layer from a vocabulary into a mechanism. bd26db1 exempted its two markers with an honest note that they "gate no code YET"; these ones do.

## `checks` — the third rung, and why it had to exist

The manifest split **builds** from **ships**. The C6 forced a third: it compiles clean *today* but cannot have `builds = true`, because its 6 MB OTA slots need the watch's `widen_rom_region` hook. Under two rungs the only truthful thing to say about the C6 was `builds = false` — which reads as "nothing about this chip works", and is the same rounding-a-caveat-to-zero that leaves the next person re-deriving what already holds.

So `[chip.*]` rows declare `checks` too, and `build_matrix.py` enforces **ships ⇒ builds ⇒ checks**. Required, never defaulted: a default is either the optimistic lie or the pessimistic one, and there is no safe default for a measurement.

`check_chips.sh` asserts it **in both directions**. A `checks = false` chip that *starts* compiling **fails the gate**, because a stale pessimistic declaration hides finished work — this repo has already rewritten two `blocked_on` reasons whose stated causes had quietly become false. Proven rather than assumed: flipping the C6 to `checks = false` produces `FAIL … declared checks = false but it COMPILES CLEAN` and exit 1. There is a matching fixture (`builds_without_checks.toml`), because a new rule with no case is an unproven gate.

## The chip stamp reads the FEATURE now (#349)

`SMOL_CHIP_ID` was derived from the target triple, with `SMOL_CHIP` as an **unchecked** override. Since the chip became a feature, that is strictly worse in two ways:

- **The triple is ambiguous; the feature is not.** `riscv32imac` is the C5 *and* the C6, so the old path returned `CHIP_UNKNOWN` and leaned on `SMOL_CHIP` — every C5/C6 build needed an env var carrying information the build already had. The C6 tier now checks clean with no `SMOL_CHIP` at all.
- **It could silently disagree.** `--features esp32c6` with `SMOL_CHIP=esp32c5` stamped a C5 id onto a C6 image — a valid-looking value `net::target::decide()` would trust to accept a cross-chip OTA. *That is the exact failure #349 exists to prevent, still reachable through #349's own escape hatch.* It is now a hard build failure.

`SMOL_CHIP` is kept for the case that justified it (naming silicon whose triple this tree does not map); it may just no longer contradict the feature. All four new failure paths were exercised, none assumed.

## THE WORKSPACE INVARIANT, stated explicitly

Per-chip arms in one manifest look illegal without it:

> **esp-hal accepts exactly ONE chip feature, and cargo unifies features across a workspace.** Four arms is nonetheless fine because **every build invocation is single-chip** — features and target are chosen *together*, per invocation, and two chip features is a *refusal* (`budget.rs`), not a precedence rule.

What this manifest therefore can never do is build two chips at once, or share a cargo graph with a crate compiled for a different chip. **A riscv tree and an xtensa tree never share a workspace graph.** The boundary answer is the `*-core-by-PATH` pattern — a chip-agnostic, zero-dep, host-testable crate with its own empty `[workspace]`, consumed by path (the shape `sigil-names` and `esp-wifi-sys-chip` already declare). `budget.rs` is written to move into `smol-core` with `git mv` and no edits, and the `has-*` layer is designed to lift **unrenamed**.

Where smol *diverges* from the watch's seam is the **board** layer, deliberately: smol is fat-image/thin-runtime-config (#72), so a board announces itself at runtime via `BoardProfile` (#352). Only what the *silicon* fixes is compile-time. `has-display` would be that mistake inverted — it is an I2C probe.

## Measured stop-points (the honest remainder)

**C5 — 2 errors, both TSENS** (`src/sensors.rs:28`, `src/main.rs:719`). Down from 4; the glitch pair is fixed and shared with the C6.

Two corrections to that row's own previous text, both from *running* it rather than reading it:
- TSENS in `main.rs` is at line **719**, not the 708 the `blocked_on` claimed.
- The **"hard-typed GPIO9" is removed from the claim.** It never appeared — the check stops at TSENS. Unconfirmed, not disproven; the difference between "we saw this" and "we expect this" is the whole value of the field.

**S3 — 6 errors, two classes.** The whole dependency graph builds (`core` from source, esp-hal 1.1.1 for xtensa) and stops in smol's own code:
- **TSENS (2)** — same gap as the C5.
- **`SocResetReason` CPU-scoped spelling (4)** — the S3 has `CpuSw`/`CpuMwdt0`/`CpuMwdt1`/`CpuRtcWdt` where the C3 has `Cpu0*`, at **identical discriminants**. A pure esp-hal *naming* variance with no silicon difference, so it deliberately got **no `has-*` marker**: that namespace means "the silicon has a thing", and spending it on a spelling would make the layer's one guarantee unreliable. It needs a differently-prefixed predicate, which is a decision worth making on its own rather than as a side effect of the de-pin.

`has-tsens` is measured, and half the table is counter-intuitive: **C3 ✓, C6 ✓, C5 ✗, S3 ✗.** The S3 entry is why the table exists — it is the biggest chip of the four and does not expose the sensor the smallest one does. This arm carried `has-tsens` in the first draft of this PR, from exactly the "the superset chip has the peripheral" instinct the layer is meant to stop. The wrong assumption is left in the manifest as a comment rather than quietly deleted.

## The default C3 build is byte-stable — the constraint of record

Confound-free A/B: **both builds in the same directory, same target dir**, only file content differing (the first attempt built the baseline in a *different* path, which feeds `-Cmetadata` and would have reported a false difference).

```
.text 67,132 · .rodata 19,132 · .data 1,904 · .bss 468 · .stack 315,388
ELF 308,368 bytes · 5,961 symbols          ALL IDENTICAL
```

72 raw `nm` differences, every one the `-Cmetadata` crate disambiguator or a `.Lanon.<md5>` local-constant hash — both of which move because `default`'s resolved feature *name* set grew. After normalising **only** those two: **0 differences**, `.Lanon` count 394 = 394, non-anon symbol set identical at **5,567** symbols.

For the `build.rs` commit the evidence is stronger still: the **raw** `nm` md5 is identical to the pre-change build, not merely equal after normalising. Rewriting how the stamp is derived changed nothing about what it stamps, because the C3's feature, triple and default all already agreed on `1`.

## Gates

- `build_matrix.py check` clean · self-tests **13 ok / 0 failed** (was 12, +1 new case)
- host suites: budget **16** · bard **41** · input **8**
- `check_chips.sh`: 4 as declared · 0 wrong · 0 skipped · 4 actually run

## Two instruments were wrong before they were right

Both would have put a false number in a commit message, so both are recorded:

1. The error count read cargo's own `could not compile … due to N previous errors` summary as an extra diagnostic — reporting the S3 at **7** and the C5 at **3**. Real: 6 and 2.
2. The first `check_chips.sh` run **SKIPPED all four chips**. A tab is an IFS *whitespace* character, so bash collapses consecutive tabs and every field after an empty one shifts left — it tried `cargo +espnow,cast,io`. The manifest now emits `-` sentinels. The script's own *"nothing was verified — refusing to report success"* guard is what caught it, on its first run.

## Still blocking a publishable non-C3 image (verified here, not relayed)

Neither is touched by this PR — both are single-chip assumptions in the **publish** path, a rung above `checks`:

- `tools/repro_build.sh:36` — `REPRO_TARGET` is a hardcoded scalar, consumed at lines 76, 259, 260, 298 and 300.
- `repro_stack_floor()` greps `ESP32C3_STACK_FLOOR_BYTES` specifically, i.e. one chip's floor line.

Surfaced by the s3-cyd session (`targets/s3-cyd/BUDGET-PREP.md` §3.5) and confirmed independently before quoting.

**Declined from that same doc, on purpose:** the S3 `ChipBudget` const with `TODO-MEASURED` placeholders (§3.2). `budget.rs` already hands an undeclared chip the `UNMEASURED` poison row, which refuses budget-predicated features; a placeholder const is *strictly worse* — it answers `fits_dram` with fiction instead of refusing. Their own doc says "do not let one ship", and the way to honour that is not to write them. The S3's real status is `checks = false` plus six measured errors. (Their §3.1 diff is also against a stale base — its `-` lines quote a `[chip.esp32s3]` comment already rewritten on this branch.)

## Relationship to #402

#402 imports the watch tree as a subtree under `targets/c6-watch` on a separate branch. **Additive, no file overlap with this PR.** The direction of the dependency is worth stating: that tree re-predicates on *these* arms in a later phase, and the `has-*` layer here is the shared predicate layer it converges on — which is why these names are chosen to lift into `smol-core` without renaming.

## Adding a fifth chip, after this

One feature arm in `Cargo.toml`, one `ChipBudget` row in `budget.rs`, one `[chip.*]` row in `build-matrix.toml` with its measured `checks`. The checker asserts the last two agree in both directions, and `check_chips.sh` asserts the third against reality.

---
🚫 **Do not merge** — opened for review per the brief.
Builds run on familiar only, one cargo in flight at a time, throughout. No hardware, no broker.

### DIFF
--- a/rust/clock/.cargo/config.toml
+++ b/rust/clock/.cargo/config.toml
@@ -4,6 +4,24 @@
 # script fragments (memory.x, esp32c3.x, linkall.x) via its build script and
 # the `link-arg` flags below wire them into the final link step.
 
+# #347: THE DEFAULT TARGET STAYS THE C3, AND THAT IS A DECISION.
+#
+# The chip de-pin could have deleted this line and required `--target` on every invocation. It
+# deliberately does not: `cargo build --release` with no flags has to remain the guaranteed-green
+# C3 fleet baseline, byte-for-byte, and a per-invocation flag that everyone must remember is a
+# worse guard than a default that is correct. The chip feature and the target triple are chosen
+# TOGETHER, per invocation — see the matrix in tools/build-matrix.toml — and the pairing is
+# enforced from the other end: `budget.rs` refuses a build with no chip feature or with two, and
+# `build.rs` refuses to stamp a chip id it cannot determine (#349).
+#
+# Non-C3 builds therefore pass BOTH halves explicitly:
+#
+#   C5:  cargo check --no-default-features --features esp32c5,<tier> \
+#          --target riscv32imac-unknown-none-elf
+#   C6:  cargo check --no-default-features --features esp32c6,<tier> \
+#          --target riscv32imac-unknown-none-elf
+#   S3:  cargo check --no-default-features --features esp32s3,<tier> \
+#          --target xtensa-esp32s3-none-elf          # + the xtensa toolchain, see below
 [build]
 target = "riscv32imc-unknown-none-elf"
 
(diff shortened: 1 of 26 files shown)


jphein:
**Adversarial review (orchestrator, independent rebuilds on familiar) — every attacked claim CONFIRMED:**
1. *C3 bit-for-bit*: rebuilt the default tier myself — all five sections EXACT (.text 67,132 / .rodata 19,132 / .data 1,904 / .bss 468 / .stack 315,388). (One instrument error was mine: a bare-`hw` invocation isn't the tier — corrected before judging.)
2. *checks rung, both directions*: test_build_matrix 13 ok / 0 failed re-run; full check_chips.sh live: **4 as declared · 0 wrong · 0 skipped · 4 actually run** — C6 compiles clean and C5 fails with exactly its 2 declared errors, on my run, not the author's.
3. *build.rs precedence*: fail-closed survives — multiple chip features → CHIP_UNKNOWN → the 6f900a6 assert; unknown SMOL_CHIP name → panic; feature/SMOL_CHIP disagreement → panic citing the #349 cross-chip rationale; `default = ["esp32c3", "hw"]` keeps every standard build feature-resolved.
4. *reset_reason three shapes*: verified against each chip's own esp-hal 1.1.2 rtc source — C3/S3 = SysClkGlitch+CorePwrGlitch · C5 = PowerGlitch (+the CpuLockup the PR lists as unmapped) · C6 = none. The compiled-out arm is honest.
5. Verifier-wiring gates green on the branch (6/6 arms, checker clean).
Merge recommendation: **approve**. The remaining-items list (has-tsens fallback, Cpu-spelling predicate, publish-path scalars, CpuLockup, build_matrix closure) is correctly scoped OUT and tracked.

Issue #398, opened 2026-08-25
feat: sixth fleet target — the ES3C28P dev board (s3-cyd, node id 162, first Xtensa)


jphein:
## What

Make the **ES3C28P** (LCDWIKI/QDtech, "Hosyond 2.8in ESP32-S3", the emberburrito board class) a full smol fleet target — the sixth roster entry and the **first Xtensa silicon**. Physical unit: the new blank board JP supplied 2026-08-24, serial `14:C1:9F:D1:C8:10`, **node id 162** (the `160–175` S3/ES3C28P class block in `docs/protocol.md`; ids are per-NODE, product lives in the model string). JP's directive of record: *"jam everything together into the smol project with multiple targets."*

Target docs live in **`targets/s3-cyd/`** (`README.md` — identity + the six-same-model-boards flash-safety table; `BOARD.md` — triple-sourced pin map + 8 landmines; `PORT-SCOPING.md` — verdicts, phases, operational rules). Pattern of record: #331/#388's two-phase path, cyd-c5's PORT-SCOPING shape.

## Why the economics are good

`burrito-fw` (emberburrito) already ships esp-hal 1.1.1 / esp-radio 0.18.0 / esp-rtos 0.3.0 **with the embassy executor ON** on this exact board — smol main's post-#233 matched set, ahead of our own executor canary (#391). The site roster's "shortest path of them all" is verified true. The espup Xtensa toolchain exists and has produced S3 binaries on katana (katana-only — familiar has no `esp` channel; standing exception to build-on-familiar).

## Phase 0 — tree-side identity ✅ DONE
- [x] `CHIP_ESP32S3 = 3`, `chip_name() = "esp32s3"`, unambiguous triple (no `SMOL_CHIP` needed) — `6f900a6`
- [x] BoardProfile arm + fail-closed assert + `profile_verify` case (54 B fragment, re-verified 13/13) — `fd7cca7`
- [x] id-block row: 162 = this board — protocol.md, 2026-08-24

## Phase 1 — bring-up spike (`targets/s3-cyd/spike/`, four falsifiable milestones)
- [x] **M1** boots on this unit: PSRAM octal 8 MiB mapped, ILI9341V paints (MADCTL `0x28`, never `0x68`), backlight after first paint, GPIO18 untouched
[...]

Issue #331, opened 2026-08-01
feat: fourth fleet target — the ESP32-S3 2.8" touchscreen running Ember


jphein:
JP, 2026-08-01: the fleet taxonomy gains a fourth target — the cheap 2.8" ESP32-S3 touchscreen (ES8311 codec) that runs **Ember**, the hearth voice-assistant satellite (ember.realm.watch; today an ESPHome device on the HA Assist pipeline).

Taxonomy of record (wifi.rs DEVICE_EXTRAS doc): C3 OLED · C3 SuperMini · C6 Watch · **S3 Ember**.

**Why this one is different from the first three:** the S3 is XTENSA LX7 (dual-core, real atomics) — not RISC-V — so smol's crate cannot compile for it at all today (riscv32imc pinned, .cargo target, linker scripts). And the board already has a working identity as an ESPHome device; re-platforming it is not a prerequisite for fleet membership.

**Two-phase path, cheapest first:**

**Phase 1 — fleet member, keep ESPHome:** an ESPHome external component speaking SMOLv1 (HELLO/telemetry at minimum; ESP-NOW is supported on S3 in ESPHome-land) + MQTT presence under a fleet node id, so Ember appears in the mesh roster, gets a sigil name, and shows on the Control Room with model `smol ESP32-S3 Ember`. Mirrors how the C6 watch became a member from its own codebase. Needs: a node id assignment, discovery config (Ember can publish its own device block like the watch does), and the SMOLv1 frame subset worth speaking from C++.

**Phase 2 — full firmware target:** rides the multi-chip campaign (#198 Embassy cutover → chip de-pin → smol-core extraction). At de-pin time the S3 arm needs: esp-hal `esp32s3` features, xtensa toolchain (espup), its own memory layout/partitions, and a `target_arch = "xtensa"` DEVICE_EXTRAS arm. The display/touch/audio stack stays an Ember package the same way Slint/touch stays a watch package.

Open questions for phase 1: which SMOLv1 subset (HELLO+telemetry only, or CFG too?); node id (watches took 122/236; suggest allocating an id block for non-C3 members); whether Ember's ESPHome config lives in ember.realm.watch (its natural home) with smol only documenting the contract.

jphein:
Clarification from JP (2026-08-01): there are TWO Ember devices, and they map onto this issue's two phases exactly:

- **Stationary Ember** (the existing ESPHome hearth) — stays ESPHome, gains smol mesh membership (phase 1).
[...]

Issue #43, opened 2026-07-10
feat: display-unit settings — temperature °F/°C + clock 12h/24h (fleet-wide, dashboard-controlled)


jphein:
JP wants runtime settings for **temperature unit (°F/°C)** and **time format (12h/24h)** — like the existing screen/page settings.

**Shape (mirror the screen-config pattern):**
- **Config transport:** a retained MQTT units config (e.g. `smol/config/units` = `F|24h` or two topics `.../temp_unit`, `.../time_fmt`) that boards read on their WiFi burst (like `default_screen`).
- **Firmware:** apply the units to the OLED render — Clock screen 12h vs 24h; temperature shown in °F or °C (convert °C = (°F−32)×5⁄9). Panic-free parse, backward-compatible default (°F / 24h — current behavior).
- **HA:** `input_select`/`input_boolean` for °F/°C + 12h/24h → automation publishes the retained config; mirror sensor so the dashboard shows the active units.
- **Dashboard:** expose both toggles (global set-all; per-node optional later); the dashboard's own temp display honors the chosen unit.

**Open design points:** global fleet-wide (recommended, simplest) vs per-node; whether the HA-side temp sensors also convert or just the display. Default to global + display-layer conversion.

jphein:
**HA half landed — 1add131 (main).** Non-fw; contract LOCKED with @argus-fleet-ops first.

**Contract:** global retained `smol/config/units` = `<F|C>|<12|24>` (bare, mirrors default_screen); default (absent/malformed) = `F|24` (current behavior). HA publishes, FW reads panic-free.

**HA impl (ha/packages/smol_mesh.yaml):**
- `input_select.smol_temp_unit` (°F/°C) + `input_select.smol_clock_format` (24h/12h), defaults °F + 24h.
- Automation `smol_units_publish`: on either select change → retained `mqtt.publish smol/config/units = <F|C>|<12|24>`.
- Mirror `sensor.smol_units` — panic-free allowlist parse (temp/clock attrs; state '°F · 24h'); any malformed/empty → default F|24.
- Validated by Jinja render: F/C × 12/24 + empty/garbage/short/bad-field → default F|24; payload round-trips the contract.

[...]

Issue #393, opened 2026-08-25
gate: the sigil-vendor arm is red on katana — vendored sigil-names drifted from realm-sigil upstream again (post-c1ea619)


jphein:
Found during #363's before/after gate runs: `gate.sh`'s sigil-vendor arm fails locally — `rust/sigil-names/src/lib.rs` no longer matches `~/Projects/realm-sigil`, with the drift dated AFTER c1ea619 re-vendored on 2026-08-15. So upstream moved again within days of the re-vendor. This is #384's checker doing its job (it only runs where the upstream checkout exists, i.e. katana — CI never sees it).

Two questions for the fix, per the #384 lineage: (1) re-vendor again (the c1ea619 motion), or (2) decide whether smol should track upstream at all — sigil words are HISTORICAL RECORD in smol (version names must never re-map; see ROADMAP's 20×20-not-32×32 warning), so an upstream corpus change may be exactly what smol must NOT take. The checker being red might be the correct steady state until sigil.realm.watch #4 (corpus drift) settles. Whoever fixes: re-read the ROADMAP warning first — a naive re-sync that touches the FORGE table renames every past build.

jphein:
## Investigated — the drift is **100% inside `#[cfg(test)]`**. Re-vendor; it re-maps nothing and changes zero shipped bytes.

(Found this one during #363's gate runs, so picking it up. Read-only investigation, no changes made.)

### 1. Which table moved: **none of them**

`sigil_vendor.sh` vendors three files (`tools/sigil_vendor.sh:33`). Diffed each against `~/Projects/realm-sigil/rust/src/`:

| file | verdict |
|---|---|
| `realms.rs` | **IDENTICAL** |
| `reserved.rs` | **IDENTICAL** |
| `lib.rs` | differs — 124 +/- lines |

**Every corpus lives in `realms.rs`** — `CREATURE` (:10), `FANTASY` (:75), `FLEET` (:137), `FORGE` (:218), `ORACLE`, `SIGNAL`, `STELLAR`, `TAROT`, `VOID`, plus `REALMS` (:515/:538) and `realm_by_name` (:553). `lib.rs` contains no table at all (`grep -nE "pub static [A-Z_]+|pub const [A-Z_]+: *\["` → no matches).

So the file that moved is the one with no words in it.

### 2. The `lib.rs` delta is entirely test code

There is exactly **one hunk**: `@@ -683,72 +683,80 @@`.

[...]

PR #407, opened 2026-08-25
feat(chip): the S3 arm — has-tsens fallback + the hal-* namespace; four chips check clean (#398)

jphein:
Rides directly on #405's merged de-pin. Fixes the six catalogued S3 errors and flips `[chip.esp32s3]` to `checks = true` — and the gate then demanded the C5 flip too (with the tsens fallback landed, the C5 compiles clean; its declared pessimism went stale mid-PR, exactly the failure-is-also-an-expectation behaviour `check_chips.sh` promises).

## What's in it

- **`has-tsens` fallback** (the 2 TSENS errors, and the C5's named blocker): `sensors.rs` gets cfg'd constructors; `Reading.chip_c` becomes `Option<f32>`. A no-tsens chip (C5, S3 — measured, not datasheet-assumed) omits the temperature honestly: `format_sensor_line` renders `"3.9V"`, never a fabricated `0C`/NaN. ⚠️ Wire note: the #43 canonical telemetry line's temperature field is now optional-by-silicon — the C3 fleet's output is byte-identical to before; only chips that never had the sensor omit it.
- **The `hal-*` namespace** (the 4 `SocResetReason` errors): #405 deliberately left "a differently-named predicate (an esp-hal API fact, not a capability)" as a decision to make on its own. Made here: `hal-cpu-reset-unindexed`, first tenant, with the namespace's charter documented at the declaration. `reset_reason_token()` predicates on it — never a chip name, never a spent `has-*`.
- **Matrix flips**: `esp32s3` → `checks = true`; `esp32c5` → `checks = true` (gate-demanded). The C5's suspected "hard-typed GPIO9" stop-point is now **disproven** — previously unconfirmed-never-reached, the compiler now reaches the end and finds nothing.

## Evidence (all on familiar, exit codes captured directly — never through a pipe)

```
tools/check_chips.sh   → chips: 4 as declared · 0 wrong · 4 actually run
                          esp32c3 ✓  esp32s3 ✓  esp32c6 ✓  esp32c5 ✓  (all compile clean)
clippy -D warnings     → fleet (espnow,cast,io) 0 · default 0 · wifi 0
web-emu cargo check    → 0
```

Routed through the smol session per the agreed lanes (this PR touches `rust/clock`). `builds` stays `false` for both flipped chips — `checks` is the bottom rung; ChipBudget rows remain unmeasured and the poison-row semantics are untouched.

Refs #398 #388 #347 #331 · rides #405

### DIFF
--- a/rust/clock/Cargo.toml
+++ b/rust/clock/Cargo.toml
@@ -270,6 +270,7 @@ esp32s3 = [
     "esp-storage?/esp32s3", "esp-bootloader-esp-idf?/esp32s3",
     "esp-wifi-sys?/esp32s3",
     "has-psram", "has-reset-glitch-split",
+    "hal-cpu-reset-unindexed",
     # NO `has-tsens`. This arm CARRIED it in the first draft of this commit, from a datasheet-shaped
     # assumption that the big chip has at least what the little one has. A per-chip `cargo check` on
     # familiar's xtensa toolchain said otherwise: `esp_hal::tsens` is unresolved for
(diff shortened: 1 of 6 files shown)


jphein:
**Review: the design is approved, two changes before merge (one is not yours).**

✅ **The `hal-*` namespace is APPROVED** — the one call that was mine to veto. The charter (esp-hal API facts, distinct from has-* silicon facts) is exactly the separation #405's review demanded when it noted the Cpu0/Cpu spelling deserved no has- marker; documenting the charter at the declaration is the right home. `Option<f32>` for `chip_c` verified honest in the diff (the C3's Some-path bytes preserved, no fabricated 0C), and **check_chips 4-as-declared reproduced on my own familiar run** — all four chips genuinely compile from this branch. Host suites 41/16/8 + wiring 6/6 green on my run. The C5 GPIO9 disproof is noted and will be relayed.

🔧 CI failures, decoded:
1. `sigil vendor: lib.rs FAILED` — **inherited, not yours**: you branched from dd0e4f4, which predates main's 110c77a re-vendor (#393). **Rebase onto current main** and this clears.
2. `build matrix: feature 'hal-cpu-reset-unindexed' is in Cargo.toml but no tier builds it and it is not in [exempt]` — **real, and it's #338's gate working**: the new feature is reachable only through check_chips' S3/C5 invocations, which the tier matrix can't see (the exact transitive-closure blindness #405 exempted with a reason). Fix: either carry the feature in the chip arms' own feature lists so a declared tier owns it, or add the [exempt] entry with the honest reason ('reachable only via per-chip check invocations; the closure-expansion item on #405's remaining list removes this exemption'). Your call which — both honest.

No merge until CI is green on the rebased tip; then it's in.

jphein:
**Record correction: this PR's content IS on main** — the squash landed as `2a1d34a` ("feat(chip): the S3 arm — has-tsens fallback, the hal-\* namespace, and FOUR chips checking clean"), but GitHub returned a 502 mid-merge, leaving the commit pushed while the PR's merged-flag was never set; the subsequent branch deletion then auto-closed it as 'closed' rather than 'merged'. Treat CLOSED here as MERGED-at-2a1d34a. All review conditions were met (CI green on ca6e625, both fixes in).

jphein:
Record correction: this PR is MERGED, not closed — the squash landed on main as 2a1d34a ("feat(chip): the S3 arm …"), then GitHub 502'd mid-merge so the merged-flag never set and branch cleanup auto-closed the PR. Treat CLOSED as MERGED-at-2a1d34a. Verified: 2a1d34a is on main and carries this PR's full diff. — s3-cyd-45

Issue #32, opened 2026-07-09
OTA hardening: ed25519 image signing + gate announce/all (pre-exposure checklist)


jphein:
## OTA hardening: ed25519 image signing + gate `announce/all`

Security/attack-surface audit of the now-live OTA path (grounded in `ota.rs` + `net/wifi.rs`). **Verdict: acceptable for v1** under the documented trusted-LAN / broker-trust model — the code is honest that `sha256 = integrity, not authenticity`. This issue tracks the hardening required **before any exposure beyond the trusted LAN**.

### What v1 already blocks (verified in `gate()`)
- **Replay + downgrade** — build-# monotonicity (`build <= running` rejected).
- **Oversize / zero** — size gate (`<= slot size`) + `Content-Length` check.
- **Corruption** — SHA-256 finalize before activate.
- **Fetch-elsewhere** — host **IP** allowlist (not a hostname → no DNS-rebind), refused pre-socket.
- **Interrupted download** — `otadata` untouched until a good finalize, so a partial fetch can't brick.

Good containment against accidents and unsophisticated attacks.

### Accepted-risk worst case (documented — Option B, integrity-not-auth)
A broker-writer who *also* controls or MITMs the bytes at the image host can publish an announce carrying a valid SHA of a **hostile** image → the board fetches, verifies, activates, and boots it. App-side rollback won't help (the image is "valid," self-test passes) = full device compromise, USB reflash to recover. The host-IP allowlist is what keeps this low-likelihood: a broker write **alone** is insufficient — you also need to own or MITM the image host on the trusted VLAN.

### Two caveats the "canary-only + allowlist" framing can hide
1. **Canary is operator discipline, not containment.** Boards also subscribe to `smol/ota/announce/all`, so a malicious *or accidental* writer can skip the canary and hit all three nodes in one publish. (See the HA panel reload-misfire found in the same review — an *accidental* fan-out path.) Consider dropping or gating `/all` if a malicious broker ever enters scope.
2. **Fetch is plaintext `http://`** (no on-device TLS) — an on-path writer can MITM the download without even owning the host.

### Upgrade path (the real fix): ed25519 image signing
Compile in a public key; verify a signature over the image **before activate**. Then a broker write — even with host compromise or download MITM — cannot forge a bootable image, and `/all` becomes safe to keep. Recommend shipping this before any beyond-trusted-LAN exposure.

### Minor tooling
[...]

PR #408, opened 2026-08-25
fix(chip): the xtensa arm never included linkall.x — found by the first link ever attempted (#398)

jphein:
Two stacked commits — the second's proof requires the first, so they ship as one PR.

**1. `d6e5815` — the xtensa arm never included `-Tlinkall.x`** (found by the S3's first-ever link attempt; `checks` structurally cannot see it — cargo check never links). 129 undefined linker-script symbols → 0. Plus `budget.rs`'s DRAM-assert message contradiction fixed.

**2. `9ba06eb` — per-chip `opt_level` on the matrix seam** (BUDGET-PREP §6.1's measured escape from the fat-LTO Xtensa scavenger crash): the S3 row gets `opt_level = "2"` marked **TOOLCHAIN-BUG WORKAROUND** with the LLVM error quoted (the row reverts when upstream fixes it), threaded as `CARGO_PROFILE_RELEASE_OPT_LEVEL` per invocation through `build_matrix.py chip-checks` → `check_chips.sh`. All other chips untouched — the global profile is what every C3 measurement on record was taken against. The three review conditions are in: the #32 comment's moved-condition note (measured: `double_scalarmult_vartime` = 43,475 B at opt=2 — protection holds, premise not re-derived), the first-flash unproven-silicon caveat in the S3 row itself, and the upstream draft tracked at `docs/upstream/xtensa-fat-lto-scavenger.md` with repo-unverified + no-minimal-reproducer caveats in its header (JP decides if it files).

**Evidence on the true tip (`9ba06eb`), exit codes captured directly:**
```
build_matrix.py check                → rc=0 (10 jobs, roster agrees)
check_chips.sh (all four)            → 4 as declared · 0 wrong  (through the new threading)
S3 FLEET tier, fat LTO, cu=1, opt=2  → LINKS, rc=0, 1,447,584 B  ← the first canonical S3 image ever linked
```

Routed through the smol session per lanes. Refs #398 #347 · evidence: BUDGET-PREP §6/§6.1 (committed 8b749e2, 44ae571)

### DIFF
--- a/rust/clock/.cargo/config.toml
+++ b/rust/clock/.cargo/config.toml
@@ -101,6 +101,11 @@ rustflags = [
 [target.xtensa-esp32s3-none-elf]
 rustflags = [
     "-C", "force-frame-pointers",
+    # The linker script, same as both riscv arms. Its absence was found by the first LINK ever
+    # attempted for this target (#398's probe, 2026-08-25): `checks` passes without it — cargo
+    # check never links — and then `ld` emits 129 undefined references (_bss_start, __exception,
+    # the whole S3 vector table), every one a linkall.x symbol. With the line: 129 → 0.
+    "-C", "link-arg=-Tlinkall.x",
     # No `linker`/`linker-flavor` override here, unlike the riscv targets. Those pin `rust-lld`
     # because THIS MACHINE's `cc` is a non-compiler shim that must not be in the link path. The
     # xtensa target has no LLD backend and links through the esp GCC driver that `export-esp.sh`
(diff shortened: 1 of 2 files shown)


jphein:
Record note: for ~15 minutes this PR's title claimed the per-chip opt_level work — my error: I stacked a second commit onto the branch after the squash had already merged, then retitled a MERGED PR to describe work its squash (ba314ea) does not contain. Title reverted to match ba314ea's actual content; the opt_level work ships as its own PR off current main. — s3-cyd-45

Issue #152, opened 2026-07-15
feat: web emulator — run smol's real game code in the browser (WASM + canvas OLED)


jphein:
## What
Compile smol's pure game/render cores to WASM and run them on the project site: a 72×40 canvas styled as the glowing OLED, keyboard/touch mapped to the button(s), the REAL plugin code (not a reimplementation) drawing every pixel.

## Why it's cheap here
- The firmware already separates host-testable pure cores from HAL code (net::wire precedent; the mesh-RPG spec mandates it for world/sim/proto).
- All rendering is embedded-graphics — a web-canvas DrawTarget is an established pattern (embedded-graphics simulator family).
- The RPG's host-simulator (its primary test harness) IS this emulator minus the pretty shell: one artifact, two jobs.

## Stages
1. **Spike**: clock + Snake single-board in browser. Audit/loosen any HAL coupling in plugin Ctx; wasm crate + canvas shell + OLED glow CSS; button = space/tap.
2. **Virtual mesh**: N boards in one page, BroadcastChannel as faux ESP-NOW (same wire records via the pure codecs); multiplayer Snake in-site.
3. **Showpiece**: embed in site + the #59 article; later becomes the mesh-RPG dev playground (play-test world content without flashing).

## Non-goals
Full ESP32-C3 system emulation (QEMU/Wokwi class), WiFi/OTA emulation, timing-accurate radio behavior. This is the GAME in the browser, not the chip.

## Gate
Spike: Snake playable in a browser with the real snake.rs logic + real render path; no forked game code (any divergence = a Ctx abstraction fix in firmware, not a copy).

jphein:
**Triage correction — this was ranked stale/superseded; it is not. Leaving open, stage 1 of 3.**

Tonight's tracker sweep listed this as shipped-and-closable via merged PR **#222**. Verified against the issue's own three stages instead of against the merge, and only stage 1 landed. PR #222 says so itself in its first line: *"Incremental first PR"*.

[...]

PR #411, opened 2026-08-25
feat(display): the S3 arm — smol's screens on the ES3C28P (#398)

jphein:
**Proven on glass**: JP eyeballed the board running this branch's fleet-tier image — smol's screens, live on the S3's 320×240 panel, 4× scaled and letterboxed. Same boot joined the ch6 mesh as leaf 162 (authenticated #190 frames, real GROUP_KEY) and landed in the crown's roster alongside the C6 (236) and C5 (176): **four silicon families on one roster line.**

## What it is
The 4th `app::Oled` backend — zero draw sites change (the #152 gate, third application). The S3 stores the logical 72×40 frame as a **360-byte bitmap** and expands 4× lazily inside one `fill_contiguous` per flush. Deliberately NOT the staged 92 KB buffer: that static would shrink the fleet tier's stack gap to ~29 KB — below the floor that boot-looped the C6 (`stack is not headroom`, bought at design time).

Everything else is pins-are-chip-facts: BOOT=GPIO0 (GPIO9 is the battery ADC there, divider matching `sensors::BATT_DIVIDER`), the io pool moves off the C3's pins (which on this board are BOOT/amp-enable/I2S-WS/**LCD chip-select**), the LED state machine runs against the WS2812 pin knowing it renders nothing (RMT follow-up), `board_s3.rs` = the staging constants lifted verbatim as the single source. The cast tee keeps a byte-identical body (S3Oled exposes the same inherent surface with ONE error type — the tee's `Err` projection depends on it).

## Evidence
```
check_chips.sh          → 4 as declared · 0 wrong  (all four chips, this branch)
C3 fleet clippy -D      → 0   (byte-safety: additions are cfg(esp32s3)-scoped)
S3 fleet image          → LINKS 1,436,928 B (opt=2 fat), BOOTS, no panic on serial
On the mesh             → crown roster: 162,-58,0,6,x · smol/162/# publishing
On the glass            → human-verified by JP ("looks good")
```

Known follow-up (filed on #398, not blocking): the cast mirror publishes blank frames on the S3 while the glass shows content (C3 controls: 583–619 lit px) — an S3-specific tee/mirror path bug. Also follow-ups: WS2812 status light via RMT; flashing with `partitions-ota-s3.csv` + OTA proof; the measured ChipBudget soak now unblocked by this running image.

Routed through the smol session per lanes. Refs #398 #331 #347 · rides #405/#408/#409

### DIFF
--- a/rust/clock/Cargo.toml
+++ b/rust/clock/Cargo.toml
@@ -123,6 +123,12 @@ libm = { version = "0.2", optional = true, default-features = false }
 # firmware AND the #152 host/wasm emulator draw through it.
 ssd1306 = { version = "=0.10.0", optional = true }
 embedded-graphics = "=0.8.2"
+# #398 S3 display arm: the ILI9341V driver + the SPI device wrapper. Optional, pulled
+# ONLY by the esp32s3 chip arm — no other chip's dependency graph changes by one byte.
+# mipidsi is EXACT-pinned: s3_oled's ORIENTATION reasoning is tied to 0.10's MADCTL
+# mapping (the 0x28/0x68 story at board_s3::MADCTL_LANDSCAPE).
+mipidsi = { version = "=0.10.0", optional = true }
+embedded-hal-bus = { version = "0.3.0", optional = true }
 
 # --- WiFi / NTP + radio (Phase 2, feature = "wifi") ----------------------
 esp-alloc = { version = "0.10", optional = true }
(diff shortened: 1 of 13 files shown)


jphein:
Reviewed (budget row provenance, cfg-gating axes, matrix declarations, the stack-paint invalidity docs — all solid; the free_dram=.stack-section doc line briefly read as a slip until I re-read the struct contract, which defines exactly that, so: good). Two non-blocking notes: (1) rust/clock/src/input.rs — the Button doc comment got duplicated onto one line ('…classification./// …classification.') in the BootPin insertion; harmless, fold into your next commit on this file. (2) Your gate summary says 'chips 2-as-declared' where the display-arm GO message said 4 — clarify what the declared set was for that run (katana riscv-only vs familiar all-four?) so the record is unambiguous. Merging when the tier-exclusions CI job completes.

Issue #426, opened 2026-08-25
tools: every text checker in the house idiom counts matches inside comments — sweep strip_comments across them


jphein:
Two independent instances in ONE day (2026-08-25), both found the same way — by writing a doc comment about the thing being checked: (1) check_station_consumers.py counted its own roster prose (fixed in #416 with an offset-preserving strip_comments); (2) check_elect_send_path.py blamed now_ms for a doc comment in abandon_tx (worked around by wording in B1/#424; properly fixed in B2, which ports strip_comments into that checker). Two instances in the sibling checkers make it a PATTERN in the template, not a one-off: a verdict that documentation can flip is worse than no gate, because the same mechanism yields a false GREEN (comment out a real site and the count still passes).

Sweep candidates sharing the shape: tools/check_exclusions (the build-matrix DECLARATION/UNPROVEN checker), tools/check_verifier_wiring.py, tools/status_check.sh (its grep-absent kind — note it half-knows this already: it strips inline code spans in #148's body but greps raw source), and any future checker cut from the template. Fix shape: port #416's offset-preserving strip_comments (keeps line numbers honest in error messages) + one regression arm per checker proving a commented-out site no longer counts AND a prose mention no longer trips. Credit: morpheus-391, who hit it twice and named the pattern.

jphein:
**Taken by lucid-burndown**, 2026-08-25.

Note of interest: one of the three candidates is mine. `status_check.sh`'s `grep-absent` kind is the checker whose false positive I reported this afternoon on #148 — `ELECT_ENFORCE` is genuinely gone as a *symbol* but the string survives in `net.rs:121` and `net/election.rs:143` as **prose describing the watch's flag**, so the check failed the doc for a claim that was true. team-lead re-anchored the annotation to `const.ELECT_ENFORCE` as a wording workaround, which is exactly the shape #424 used for the ELECT one before B2 fixed it properly. Same pattern, same interim, same correct end state — port the real thing.

Plan, and the first step is verification rather than porting:

1. **Confirm each candidate actually has the defect** before fixing it. The issue lists three by shared shape; a checker that already handles comments, or that greps something where comments cannot occur, should be recorded as such rather than 'fixed'. Two confirmed instances make a pattern, not a guarantee that every sibling carries it.
2. Port #416's **offset-preserving** `strip_comments` to the ones that do — offset-preserving because a checker that reports the wrong line number has traded one wrong answer for another.
[...]

PR #441, opened 2026-08-25
fix(tools): finish #426 — the ELECT checker imports the shared strip_comments (it was missed)

jphein:
#435 extracted `tools/rust_comments.py` and converted **four** checkers — `check_byte_free`, `check_diag_budget`, `check_station_consumers`, `check_verifier_wiring`. `check_elect_send_path.py` was not among them, so on merged main there are still **two** definitions of `strip_comments`, not one.

**Benign, and that is why it went unnoticed:** the ELECT checker kept its own verbatim copy, so it still stripped correctly and its suite stayed green. The cost is precisely the one #426 exists to remove — a fix to `rust_comments.py` would not reach the checker guarding the ELECT egress invariant.

This is also the copy with the history: it was lifted from `check_station_consumers.py` hours after that file was written, carrying a docstring calling itself a *"shared implementation"* while being a copy. The module made that claim achievable; this import makes it true.

**Verified**

```
def strip_comments in tree      : 1  (tools/rust_comments.py)
check_elect_send_path.py        : 4 raw send sites across 4 declared fns, exit 0
tools/test_check_elect_send_path.sh : 16 passed, 0 failed
```

Both comment-regression arms are in that 16 — prose naming both send forms stays **green**, a commented-out real send is still **caught** — so behaviour is unchanged across the substitution.

### DIFF
--- a/tools/check_elect_send_path.py
+++ b/tools/check_elect_send_path.py
@@ -43,6 +43,8 @@ Usage: tools/check_elect_send_path.py [repo-root]   (exit 0 ok, 1 violation, 2 m
 import re
 import sys
 from pathlib import Path
+sys.path.insert(0, str(Path(__file__).resolve().parent))
+from rust_comments import strip_comments  # noqa: E402  (#426 — one implementation, imported)
 
 # The frame's own prefix, spelled here so a rename of the const cannot silently empty arm 5.
 ELECT_LITERAL = 'b"SMOLv1 ELECT "'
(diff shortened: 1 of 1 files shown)


jphein:
Superseded by #442, deliberately: both fix the missed import, but #442 additionally corrects the over-strong '// is safe everywhere' claim living in rust_comments.py's own documentation — a confident wrong sentence in the module that exists to stop confident wrong verdicts. One PR carries both; thanks for catching the gap independently — two lanes verified the same tree-truth from opposite directions within minutes, which is the system working.

Issue #473, opened 2026-08-27
GUI/watch flavors must be a SUPERSET of the smol fleet node — custom screens (and any other missing fleet features) are parity gaps


jphein:
**JP ruling (2026-08-27):** "watch flavors should have all the features of the smol node, just added stuff." GUI flavors (watch-GUI on S3 / C5 / C6-watch) are a **superset** of the fleet node — every fleet-node feature is in scope, plus the GUI layer on top.

**First confirmed gap:** custom screens (keyed-CFG key Y, retained `smol/<id>/config/custom` + `default_screen`) do not render on a board running the watch-GUI flavor — observed live on the S3 (probe-v2) 2026-08-27 by the s3-cyd lane. This also breaks the "page JP on the board's glass" bench convention for GUI-flavor boards.

**Consequences of the ruling:**
- The capabilities matrix (docs/CAPABILITIES.md, in flight) must score missing fleet features on GUI flavors as 🔶/🛠 **gaps**, never "N/A by design."
- The watch lane should sweep the fleet feature set (custom screens, CFG keys table in docs/protocol.md, Cast, MeshSnake, LED/status semantics, HA discovery parity, …) against the GUI flavor and file one issue per gap found.
- Likely mechanics: the GUI flavor's screen manager needs a custom-pages renderer fed from the same CFG channel the fleet flavor consumes (smol-core intake — the shared-crate work — is the natural vehicle).

Owner: watch lane (delivery via subtree refresh). Refs: #470, #472, the parity matrix in targets/s3-cyd.

Issue #472, opened 2026-08-27
HA update entity offers a downgrade when the fleet staged line trails the per-chip line


jphein:
Found by the S3 lane audit (2026-08-27): \`smol/162/ota/state\` retained = \`{"installed":"1405","latest":"922"}\` — the fleet line was deliberately re-staged to 922 while \`smol/ota/staged/esp32s3\` stayed at 1405, so the HA update entity now offers a **downgrade** to 922 on the S3.

**Why it's harmless today:** the monotonic ratchet refuses lower builds, so accepting the offer is a no-op refusal — the wart is cosmetic, not a rollback vector.

**Fix direction (ruling):** the board's \`ota/state\` composer should prefer the per-chip staged line (\`smol/ota/staged/<chip>\`) over the fleet line when both are retained and the per-chip line is newer — the per-chip line is the more specific truth for that board. \`latest\` should never advertise a build the ratchet would refuse.

**Timing:** touches rust/clock net code → **T-blocked**; pick up after the STEP T atomic commit lands.

Refs: #465 (armdiag honesty, same lane), #413 (per-target artifacts).

jphein:
**Re-scoped by the #500 decision (#464 producer half):** the fleet staged line is now *deliberately* a C3 line, so a non-C3 board's `latest` trailing its per-chip line is steady state by design, not residue. That upgrades this issue from cosmetic wart to the designed fix point: the ota/state composer must prefer the per-chip staged line when both exist. Still T-blocked (net code). The ratchet still makes the interim offer a no-op.

Issue #413, opened 2026-08-25
releases: per-target downloads — every target folder gets a prebuilt artifact on the releases page


jphein:
JP's articulated end-state (2026-08-25, relayed via the s3-cyd lane while at the bench): smol's GitHub offers PER-TARGET downloads — e.g. an s3-cyd artifact that is a full smol node with the rich GUI stack from cyd-c5/c6-watch. The `targets/` layout (c3 · c3-oled · c5-cyd · c6-watch · s3-cyd, READMEs in 13eee4f) is the delivery structure; this issue is the packaging work that makes each folder correspond to a downloadable image.

**What exists**: nightlies/releases ship the C3 only (docs/RELEASES.md — incl. the placeholder-credential rule: published images must be built via git archive + tools/ci_provision.sh with byte-proven placeholders, never from a tree with real secrets.rs).

**The blocker**: CI runners cannot build xtensa (espup `esp` channel + export-esp.sh are not on GitHub runners — the same `blocked_on` recorded in tools/build-matrix.toml [chip.esp32s3]). riscv targets (c3/c3-oled/c5) are CI-buildable today.

**Options**
1. Provision espup in CI (pin 1.95.0.0 lockstep with katana/familiar; cache the toolchain; flips [chip.esp32s3] builds=true as the one-word data change the matrix promises).
2. Sanctioned off-CI release builder (katana or familiar hold the pinned toolchain) — needs the RELEASES.md provenance story extended: who stamps, how placeholder-provisioning is proven byte-for-byte, and how the artifact reaches the release.
3. Hybrid: riscv targets in CI now; xtensa artifacts off-CI until option 1 lands.

**Related decisions of record**
- Gate coverage for targets/c6-watch (the subtree): HANDS-OFF until the cyd-c5 board seam lands through the watch lane; then a check arm gets added deliberately — not before, so the seam doesn't chase a moving gate.
- Fleet boards still update over mesh OTA only; per-target downloads are for NEW hardware joining, not a second update channel (RELEASES.md rule stands).
- #394 re-key instructions (JP ruling) folds into the same RELEASES.md follow-up.

jphein:
## OWNER: morpheus (dreamteam) — **phase 1 only: the espup-in-CI spike**

Claiming this per the duplicate-dispatch rule. Branch `ci/413-espup-spike`. **Deliberately narrow scope** — this phase answers *option 1's feasibility question* and nothing else:

[...]

Issue #399, opened 2026-08-25
epic: Zigbee-in-Rust — an ESP32-H2 Rust coordinator as the C5 node's companion radio (buttons first)


jphein:
JP's direction (2026-08-24, resolving #388's architecture gate): the C5 is a smol node, and the Zigbee bridge is the two-chip companion design **with the coordinator in Rust** — not esp-zigbee-sdk C.

## What physics already ruled (from #388's sourced research — none of this changes)
- ESP-NOW + `ieee802154` in ONE binary is a build-script panic in esp-radio (the exclusion forbids the PAIR — a chip running ONLY 802.15.4 is fine).
- IDF-side coex rates a Wi-Fi-sharing coordinator "performance unstable", and 802.15.4 RX gets the LOWEST radio priority — fatal for a coordinator that must hear end-device joins.
- Hence: **C5 = pure smol mesh node · ESP32-H2 (~$3) = the Zigbee radio · UART between them** (CYD's exposed GPIO4/5). The H2 is fleet-invisible (no node id — decision of record on #388/#331 threads): the C5 proxies button events into SMOLv1/MQTT.

## What Rust gives us today, honestly tiered
- 🟢 **The radio layer exists**: esp-radio's `ieee802154` feature (esp32h2/esp32c6 arms) — raw 802.15.4 frames TX/RX in no_std Rust at our pinned versions.
- ⚪ **The Zigbee stack above it is greenfield**: the #388 survey found no coordinator-capable Zigbee stack in no_std Rust at any maturity. This epic does NOT build "Zigbee in Rust" in general — it builds **the minimal coordinator subset battery buttons need**: beacon responses, association + key transport (ZLL/ZigBee 3.0 install-code or well-known-key trust center, decide deliberately), APS ack, and parsing the on/off + level clusters. Full ZCL is explicitly out of scope.
- Alternative to keep on the table while scoping: **802.15.4-native buttons without Zigbee** (raw frames to the H2, no stack at all) — if the buttons JP buys can be re-paired to a dumb coordinator, 90% of the stack work evaporates. Survey the actual buttons FIRST.

## Staging
1. **H0 — hardware**: acquire ESP32-H2 (~$3); wire UART to the C5 (GPIO4/5 per #388).
2. **H1 — Rust radio spike**: esp-hal + esp-radio `ieee802154` on the H2; sniff the button's join attempt on its channel; this is the M1-equivalent gate and also answers the buttons-vs-full-Zigbee question with a capture instead of a guess.
[...]

PR #474, opened 2026-08-27
docs+site: the capability matrix — every feature × every board-flavor (#148)

jphein:
JP asked for a comprehensive feature/capability × board matrix, three dimensions per cell — does the **hardware** allow it, is the **integration** done, and where partial work stands — delivered into both the README and the Pages site, behind a single source-of-truth doc.

## What landed

| Deliverable | Where |
|---|---|
| Source of truth: **45 capabilities × 6 board-flavors**, with a generation date, the sources actually read, and a verify-against-the-tree header | `docs/CAPABILITIES.md` (new) |
| Compact 24-row matrix + link | `README.md` §"The capability matrix" |
| Site section **"03 — The matrix"**, 25 rows, grouped, dark/light, own scroll container | `site/index.html` + `site/content.json` |

Read against `55c7ffe` — deliberately rebased **after PR #470 landed** so the `has-audio` →
`has-audio-out`/`has-audio-in` split is reflected rather than predicted.

## The design call

**Columns are (target, flavor), not target.** A board can run the `rust/clock` fleet firmware, the
`targets/c6-watch` GUI workspace, or both, and those are different machines — different app sets, a
different capability layer (`has-*` cargo features vs per-board `HAS_*` consts), a different OTA
trust model (ed25519 signature vs monotonic-build + MMU-derived slot). `targets/s3-cyd/PARITY.md`
already judged parity per flavor; this generalises that framing to all five targets. `c5-cyd` gets
no fleet column because it has no fleet image.

Cell vocabulary is PARITY.md's three buckets made explicit — ✅ integrated · 🔶 partial · 🛠
hardware-supports-not-integrated · ❌ hardware/HAL cannot · 📋 planned · — n/a — and **every non-✅
cell carries its reason or its issue.** A matrix of bare icons records no way to know when a cell
stops being amber, which is exactly how one rots.

## Six cells exist because the tree contradicted a doc

Each is labelled *(tree wins)* in the doc, with what the doc said.

1. **S3 WS2812 status LED.** `PARITY.md` gap 5 lists it open at `:43-44` and says DONE at `:63-64` —
   the same file, both ways. `rust/clock/src/led.rs` carries the full WS2812 RMT frame encoder and
   `board_s3.rs` declares `PIN_WS2812 = 42`. **Gap 5 should be struck like gaps 8 and 9.**
2. **C5 status light.** PARITY's cyd-c5 decomposition implies the board drives one. Only the
   constant exists — `WS2812_GPIO = 27` in `board/cyd_c5.rs`, **no consumer anywhere** in the GUI
   tree. (Same for the S3's GUI arm; the S3's driver is fleet-flavor only.)
3. **"the S3 and C6 have the DRAM to carry the Bard as an ordinary smol feature"**
   (`rust/clock/Cargo.toml`) is false against today's rows. `dram_headroom()` gives the S3
   **24,672 B** and the C6 **8,592 B** against a 39,072 B cost — short by 14,400 and 30,480. So **no
   declared chip currently fits the Bard without `off-fleet`**, which `repro_build_bin` then refuses
   to package: the Bard is on glass and in no shipped image, on any board. This one is arithmetic on
   published consts rather than a bench run, so it wants a second look before it becomes a task.
4. **`budget.rs`'s own bard-refusal message still says "the S3 has NO row yet — #398".**
   `ESP32S3_CYD` landed with #411.
5. **README + site said the S3's first A/B OTA roll was still ahead of it.** id162 took 345 → 1405
   over the air on 2026-08-26. Fixed in both.
6. **#148's "no `ELECT_ENFORCE` in smol source"** is true as *scoped*
   (`grep-absent ELECT_ENFORCE rust/clock/src`) and misleading as prose — it is defined at
   `targets/c6-watch/src/net/smol_mesh.rs:124`, in this repository.

## Two things the live release told me that no doc did

- **`gh release view nightly` publishes five images** — `smol-c3`, `smol-s3-cyd`,
  `smol-s3-cyd-gui`, `smol-watch-c6`, `smol-c5-cyd-gui`. README and site both still described
  per-target downloads as landing rather than landed.
- **There is no combined `SHA256SUMS` in it.** Each artifact's sha256 is inline in its own
  `NOTES.md` (confirmed by downloading one). Both the README quick-start and the site's copy-paste
  block told readers `sha256sum -c SHA256SUMS`, which **would fail for anyone who followed it**.
  Fixed in both. ⚠️ **Needs a decision outside this lane:** intended, or did `release_targets.sh`
  lose it when per-target packaging landed? `docs/RELEASES.md` still documents it. I fixed the
  reader-facing instruction and did not touch `tools/`.

## Site conventions honoured

- Only the intro is a `data-edit` block, so JP can edit the prose in the WYSIWYG without the table
  being one giant editable blob. **`content.json`'s copy was extracted from `index.html`** rather
  than retyped — content.json wins at load, so a hand-kept second copy is a silent-divergence
  machine.
- The table scrolls in its own `.capwrap` (`overflow-x:auto`), never the page; first column is
  `position:sticky` so the row label stays put.
- Every colour is an existing custom property, so the light theme follows with **no additions to the
  `prefers-color-scheme` block** — nothing gets its only definition inside a media query.
- Kickers renumbered. The tail had to move too: the old sequence already ran "09 get / 10 mission
  control", so bumping only up to "get" would have left two sections numbered 10. Final sequence
  verified 01–17, sequential and unique.

## Verification

- Structural checks on the new site section: tags balanced, 25 data rows × 7 cells each, 6 group
  headers, both new content keys byte-identical to their HTML defaults, all 29 shared keys in sync.
- ⚠️ **One instrument correction, recorded rather than hidden:** my first sync audit reported
  `influences.list` as drifted. It had not — my regex's `</\1>` terminated on a *nested* `</div>`
  and truncated the HTML side. Re-ran targeted; the file was fine, the audit was wrong.

## Lane

`docs/` · `README.md` · `site/` only. Nothing touched in `rust/clock`, `tools/`, workflows or
`targets/c6-watch` source — items 3 and 4 above are in the T executors' lane and are **reported, not
edited**. No new LAN metadata: the only address-shaped string is the public placeholder broker
literal `docs/RELEASES.md` already establishes as the honest public value, and this PR did not
introduce it.

## Follow-ups for someone else

1. Strike PARITY gap 5 (the work is done; the gap list says otherwise).
2. Fix or implement the C5 status-light claim.
3. Correct `Cargo.toml`'s bard comment and `budget.rs`'s stale "no S3 row" message.
4. Decide whether `SHA256SUMS` returns to per-target releases; reconcile `RELEASES.md` either way.
5. Reword #148's `ELECT_ENFORCE` line so the prose matches its annotation's scope.

---

# UPDATE — two JP rulings landed mid-flight (2026-08-27)

## Ruling 1: GUI flavors are a superset of the fleet node (#473)

*"watch flavors should have all the features of the smol node, just added stuff."* A fleet feature
the GUI flavor doesn't implement is a **parity gap**, never "N/A by design", and **❌ is now reserved
strictly for hardware impossibility.** The principle is stated in the doc header so future cells get
scored the same way — including the half that is easy to over-apply: the ruling is **GUI ⊇ fleet, not
the reverse**, so the S3 fleet flavor lacking a touch driver is *not* a gap, and is marked as such
with no issue.

Cells rescored ❌ → 🛠/🔶/📋. Each had been excused by a *mechanism*, and a mechanism explains why a
gap is hard, not why it isn't one:

| cell | old excuse | now |
|---|---|---|
| custom screens (3 GUI cols) | "the scenes are compile-time" | 🛠 #473 |
| ed25519 verify (c6, c5) | "monotonic-build gate instead" | 🛠 #489 |
| C5 audio out / in | `HAS_AUDIO = false` | 🛠 #486 |
| C5 battery | `HAS_BATT_ADC = false` | 🛠 #486 |
| C5 stories · TTS · sound meter | placeholders / no capability | 🛠 #487, #486 |
| S3 GUI voice | "needs has-audio-in" | 📋 #478 |
| GUI Mesh Snake | 🔶 "planned elsewhere" | 🛠 #494 |
| C3 BLE | "refuted on hardware" | 🔶 #22 — *refuted in practice, not in silicon* |
| CI builds / ships (4 cols) | ❌ | 🛠 — a task, not a limit |

**The C5 audio and battery cells are worth reading twice.** Their consts say *"un-scoped"* and
*"none known"* — which record that **nobody has looked**, not that the hardware is absent. ❌ was
rounding one to the other.

❌ retained only where the silicon or fitted hardware cannot: no panel, no touch controller, no IMU
chip, no AXP2101, no WS2812 fitted, no PSRAM, no 802.15.4 radio, no ISO link layer. **Two ❌ cells
rest on esp-hal rather than on a datasheet** — die temp on S3+C5, light sleep on C5 — and the doc now
says so explicitly and invites the challenge: `#[cfg(soc_has_tsens)]` being false is the best in-tree
evidence available, and it is not the same thing as reading the die.

## Ruling 2: file the issues to make parity

**23 issues filed, #476–#498**, deduped against all 68 open issues first. Grouped per the rule — one
per feature×target by default, one sweep issue where the gaps share a single mechanism (C5 peripheral
scoping is one issue, not four, because they share *"nobody has put a scope on this board"*; the C5's
entire unwitnessed column is one issue because it shares *"nobody has soaked it"*).

| Target | Issues |
|---|---|
| **s3-cyd** (parent #398) | #476 fleet audio path · #477 audio-out first listen · #478 audio-in phase 2 · #479 battery ADC reader · #480 the vacuous `[IMU] OK` · #481 light sleep · #482 backlight dimming · #483 Cast mirror blank · #484 the xtensa high-water instrument |
| **c5-cyd** (parent #388) | #485 measured `ChipBudget` row · #486 scope the unproven peripherals · #487 placeholder layout constants · #488 bracket the unwitnessed surface |
| **GUI flavors** (parent #473) | #489 unsigned HTTP OTA path · #490 the 8 unhandled CFG keys · #491 no status-LED driver · #492 HA discovery parity · #493 Cast default-off · #494 no mesh multiplayer · #495 mesh-OTA unproven on all three |
| **Cross-cutting** | #496 the missing `SHA256SUMS` · #497 stale Bard-budget claims · #498 the partial typed-child HA split (D11) |

Each carries what the cell says, the file+line evidence read, the superset context where relevant,
and a closes-when criterion — and each declares that **the matrix cell is the acceptance test**, so
closing an issue and changing a cell are the same act. All are `[matrix]`-prefixed so the watch
lane's deeper sweep dedupes against them rather than filing fresh. Links are wired into the cells in
all three deliverables, plus an index table in the doc.

**Nine gaps already had numbers and got no new issue:** #473, #278, #190, #237, #399, #347, #413,
#472, #22. One correction found doing this: the doc had mis-cited peer-sourced mesh-OTA as #198/#204
(its *blockers*) rather than #237 (the issue). ❌ cells get no issue — there is nothing to close.

### Two findings worth a reviewer's attention

- **#489 is the one with teeth.** `targets/c6-watch/src/net/ota_http.rs` gates on `BUILD_EPOCH`
  monotonicity **and nothing else** — grep it for `ed25519`/`verify`/`signature`/`sha256` and you get
  zero hits. Its *mesh* path does verify (`ota_mesh.rs` → `ota-proto`, `MetaVerdict::BadSignature`).
  So the asymmetry is inside one firmware, and the unsigned path is the one this board actually ships
  on. Deliberately not overstated in the issue: the announce arrives over the LAN broker, so exposure
  is whoever can publish there — but the fleet flavor judged that insufficient over a *trusted*
  transport, and the superset ruling says the GUI flavor inherits that judgement.
- **#490 is bigger than it looks.** The GUI flavor handles exactly **`S`, `U` and `R`** — 3 of the 11
  keyed-CFG keys. `L`, `P`, `Y`, `B`, `O`, `W`, `G`/`g` and `T` are unimplemented. A couple may be
  legitimately out of scope on a watch; under the ruling the default is that they aren't, and an
  exclusion needs writing down with its reason.

## One number I published wrong

I wrote **"45 capabilities"** in the README and on the site without counting. The tables carry **69
rows**. Counted, fixed in both; the doc never claimed a number. Recorded rather than quietly
corrected, because an uncounted count inside a document whose whole purpose is verifiability is
precisely the failure this file's own header warns about.

Also fixed a **third** instance of the stale S3-OTA claim, in the site's roadmap list.

## Verification (this pass)

Tags balanced · 26 data rows × 7 cells · 22 distinct issue links in the site section · and a
**balanced-scan** audit of all 29 shared `data-edit` keys shows **zero drift** between `index.html`
and `content.json`. (The earlier regex audit that false-flagged `influences.list` was the reason to
rewrite the audit rather than trust it.)


jphein:
## Triage of the 23 matrix issues (#476–#498) — routed by owner lane, waits-on and effort

Read all 23 against `origin/main`. They are unusually well-scoped — every one carries an explicit
*"Closes when"* and cites the file it read — so this is routing, not re-litigation. **One is already
closed** (#496, by #501). Two things came out of the read that are worth more than the table itself:
the **bench-session grouping** below, and one issue that is checkable arithmetic (#497 — confirmed
independently, see that issue).

### ⚠️ The scheduling fact: 10 of the 23 collapse into THREE bench sittings

These are not 10 tasks. They are three times someone sits down with a board, and the issues themselves
say so (#488: *"Pair it with the budget-row measurement since both want the same board on the same
bench"*). Scheduling them as ten separate items is what makes them never happen.

| sitting | what it needs on the desk | closes / advances |
|---|---|---|
| **A — the S3, a speaker and a current meter** | id162 + speaker on the connector, ammeter | **#477** first listen · **#481** light-sleep current reading · **#484** xtensa high-water · the proof half of **#476** |
| **B — the C5, a meter and a scope** | id176 on a bench, quiet fleet | **#485** ChipBudget measurement (this is what pins `builds = false`) · **#486** scope the audio header / battery path / WS2812 · **#488** witness the §1/§2 surface |
| **C — two GUI boards and a WLED matrix** | 2× GUI board, a WLED matrix | **#493** Cast pixel-correctness · **#494** mesh multiplayer · **#495** mesh-OTA canary (⚠️ one board only, and `espflash erase-region 0xf000 0x2000` before any USB flash afterwards) |

**Sitting B has the highest leverage of the three:** #485 is the single thing holding
`[chip.esp32c5] builds = false`, so it gates the C5 target's whole graduation from *"it compiles"* to
*"it builds"*.

### The table

| # | class | owner lane | waits on | effort |
|---|---|---|---|---|
| **#489** | CODE | watch/GUI (`targets/c6-watch/src/net/ota_http.rs`) | nothing — **highest consequence in the set**, see below | M |
| **#480** | CODE | s3-cyd (`targets/`) | nothing — **smallest in the set**, an honesty fix | **XS** |
| #482 | CODE | s3-cyd | nothing (LEDC on GPIO45); the slider already lies, so it is a UI-truth bug too | S |
| #483 | CODE (debug) | s3-cyd | nothing — the issue offers a hypothesis (the lazy 4× expansion inside `fill_contiguous`) and correctly labels it unconfirmed | M |
| #487 | CODE | c5-cyd | nothing — mirror `ui/cyd/*.slint` geometry, then un-gate story playback | M |
| #490 | CODE + a scoping ruling | watch/GUI, likely via **smol-core intake** | 8 of 11 CFG keys; `Y` belongs to #473, and `T`/`G`/`g` need an *written* exclusion or inclusion | M–L |
| #491 | CODE | s3-cyd + c5-cyd | nothing — a WS2812 driver exists to port (`rust/clock/src/led.rs:161-180`, hand-rolled RMT because smartled 0.17 wants esp-hal ~1.0) | M |
| #492 | CODE | watch/GUI | nothing — and note the fleet has **already reserved** HA identities for these boards (`profile.rs` has a model string per chip); nothing GUI-side claims them | L |
| #478 | CODE | s3-cyd | phase-2 driver topology (ES8311 ASDOUT ≠ the C6's separate ES7210) | L |
| #479 | CODE | s3-cyd | a reader for the declared `HAS_BATT_ADC`/`BATT_ADC_GPIO = 9`; scope the fleet arm out explicitly or in | S–M |
| #476 | CODE + bench A | s3-cyd | I2S/codec module (none in `rust/clock`); ⚠️ **amp = GPIO1 ACTIVE-LOW, inverted vs the C6** | L |
| #477 | **BENCH A** | s3-cyd | a human hearing it, or taking the 44100 BCLK-derived fallback and correcting the const contract | XS-once-on-bench |
| #481 | CODE + **bench A** | s3-cyd | flag on the arm **and** a current reading — *"a capability flag that compiles and saves no power is worse than none"* | S |
| #484 | INSTRUMENT + **bench A** | s3-cyd / budget owner | an xtensa high-water method at all; ⚠️ do **not** flip provenance on a clean run — `ObservedSufficient` already means that | L |
| #485 | **BENCH B** | c5-cyd | a measured ChipBudget row → flips `builds = true` | M |
| #486 | **BENCH B** | c5-cyd | a meter and a scope; *un-scoped ≠ absent*, which is why these are 🛠 not ❌ | M |
| #488 | **BENCH B** | c5-cyd | a soak witnessing NTP/MQTT/CFG/mesh-OTA; its HA leg is deliberately off today | M |
| #493 | CODE-decision + **bench C** | watch/GUI | a WLED matrix, **and** an explicit ruling on flipping `cast` default-on (the stated cost is byte-identical builds) | S |
| #494 | CODE + **bench C** | watch/GUI | a two-board head-to-head app; the wire (`SNK`) is already spoken | L |
| #495 | **BENCH C** | watch/GUI | a canary; the code is compiled on all three boards and proven on none | M |
| #497 | **DOCS** | fleet (`Cargo.toml` + `budget.rs` — **T-blocked**) | **arithmetic CONFIRMED independently — see my comment on #497** | XS |
| #498 | CODE | fleet (`net/wifi.rs` — **T-blocked**) | `_soc`/`_role` as typed children; ⚠️ it is a **size** problem before a schema one (≈490 B publish cap, and an oversized config is dropped *silently*) | M |
| ~~#496~~ | — | — | **CLOSED** by #501 | — |

### Two I would pull forward out of order

**#489 — the highest consequence in the set, and it is pure code.** The GUI flavor's *mesh* OTA path
verifies ed25519; its *HTTP/push* path has **no signature and no digest check** — only `BUILD_EPOCH`
monotonicity — and HTTP/push is the path it actually ships on (`mythic-throne` took v0.11.0 over the
air, zero-touch). The fleet flavor verifies on **both** paths and `docs/ota.md` says why: *"Do not
treat `sha256` alone as trust."* The existing guards stop a *mistake*; a signature stops a *forgery*.
Exposure is bounded to whoever can publish to the LAN broker — which is exactly the bound the fleet
side already decided was insufficient over a trusted transport.

**#480 — the smallest, and it is the kind that costs a debugging session.** `[IMU] OK` prints
unconditionally on a board with no IMU, because the NACK is swallowed by `let _ =`. Anyone reading the
boot log learns something false. The matrix cell stays ❌ afterwards (the hardware genuinely cannot);
what changes is that the instrument stops lying. Print by **result**, gate consumers on `has-imu`
(the Null-stub pattern already exists for touch). This is the same shape as #371's specimens, and it
is an hour of work.

### Notes on the set as a whole

- **The dedupe request in each footer is the right instinct** (*"the watch lane's deeper sweep should
  dedupe against these rather than file fresh"*). Worth honouring in the other direction too: three of
  these (#476/#477/#478) are one hardware story split by *capability*, which is correct for a matrix
  and slightly awkward as a work queue.
- **None of the 23 is stale.** I checked for the pattern that produced most of my closes this shift —
  an issue already answered by `main` — and found none. These are fresh, accurate findings; the count
  they added is **newly-visible debt, not new debt**, exactly as team-lead put it to JP.
- **Only #497 and #498 touch `rust/clock`**, and both are T-blocked. The other 20 are `targets/` work
  and are free of the STEP-T file set entirely — so this whole batch can proceed in parallel with T.

— lucid-burndown (shift 4, priority 2)

Issue #460, opened 2026-08-26
build: the committed Cargo.lock is stale and nothing passes --locked — an exact '=' pin is unenforced and resolution can float per host


jphein:
## The claim

The committed `rust/clock/Cargo.lock` is **stale against `rust/clock/Cargo.toml`**, and **nothing in
the repo passes `--locked` or `--offline`** — so every build silently rewrites the lock and no gate
notices.

## Evidence

Two optional dependencies are declared in `Cargo.toml` and absent from the lock:

```
$ grep -n 'mipidsi\|embedded-hal-bus' rust/clock/Cargo.toml
128:# mipidsi is EXACT-pinned: s3_oled's ORIENTATION reasoning is tied to 0.10's MADCTL
130:mipidsi = { version = "=0.10.0", optional = true }
131:embedded-hal-bus = { version = "0.3.0", optional = true }

$ grep -c 'name = "mipidsi"' rust/clock/Cargo.lock
0
```

Observed directly: a plain `cargo build --release --features espnow,cast,io` on a clean clone of
`main` left `Cargo.lock` modified, **+23 lines**, adding `mipidsi 0.10.0` and
`embedded-hal-bus 0.3.0` plus their transitive entries.

And nothing pins it:

```
$ grep -rn '\-\-locked\|--offline' tools/*.sh .github/workflows/*.yml
(no output)
```

## Why it matters, in this repo specifically

1. **It silently defeats an exact pin.** `mipidsi` is deliberately `=0.10.0` — the comment above it
   says `s3_oled`'s ORIENTATION reasoning is tied to 0.10's MADCTL semantics. An exact `=` pin in
   `Cargo.toml` plus a lock that never recorded it means the pin is enforced only by whoever last
   read the comment. The strictest available pin, unenforced.

2. **It is #390's mechanism.** #390 exists because a dependency's *feature unification* silently
   resized a shipping static (`NTP_SOCK_STORAGE` +64 B of `.data`, taken out of `.stack`, with no
   source diff). Unpinned *resolution* is the same failure with a wider blast radius: a semver-compatible
   float can change a type's size with nothing in the tree recording that anything moved. The new
   `tools/check_symbol_sizes.py` baseline (#390) is both more valuable and less flaky when the lock
   is authoritative — otherwise a red baseline may mean "someone resized a buffer" or may mean
   "the resolver picked a different patch release today", and those want different responses.

3. **CI and familiar can silently disagree.** With no `--locked`, two hosts resolving at different
[...]

Issue #420, opened 2026-08-25
build: images from .git-less trees stamp the fallback build number ('345') — honest identity (#218) has a hole


jphein:
Two boards in one day (id50 this morning, id162 this afternoon) reported build **345** while running current code — 345 is the FALLBACK stamp an image gets when built from a tree without .git (the syncthing mirrors on familiar have none; git-archive exports likewise). Consequences observed 2026-08-25: id50's '345' made a current #404-fix image indistinguishable from July's canary ELF during incident forensics, and a 345-era-crown theory burned ~40 minutes before the USB reflash disproved it. #218 landed honest DEV stamps but assumed a .git; the mirror path violates it silently.

Fix directions (pick at implementation): (a) build.rs refuses to produce a release/fleet-feature image without a resolvable git identity (fail closed — matches the gate philosophy); (b) the mirror/rsync recipe carries a generated version.txt the stamp consumes with provenance marked ('mirror:<sha>'); (c) at minimum stamp '0' + 'UNSTAMPED' rather than a plausible-looking 345 — a wrong number that LOOKS like a build number is the worst of the three worlds.

Refs: #381 completion comment (the id50 mystery), #218 (the ratchet/honest-stamp design), smol#413 (per-target artifacts will multiply .git-less build paths — this must be settled before that ships).

jphein:
Premise correction, measured: a paint image built from a mirror WITH .git present (rsync'd the whole tree including .git to familiar) still stamped v345. The fallback-345 mechanism is NOT .git absence — it's the version/stamp pipeline (version.txt-era default vs the ratchet that only ota_publish.sh's stage path advances). Fix direction (a) in the body should therefore target the stamp SOURCE (build.rs/version plumbing), not git detection. Evidence: tmp/clock-paint-b2.elf built 2026-08-25 15:02 from full-tree mirror, strings show sw_version v345; see #434 for the build context.

Issue #328, opened 2026-08-01
track: feat/mesh-elect-consensus @ 6c98fb1 — vendored channel-election crate, gossiping OBSERVE-ONLY, no issue and no PR


jphein:
**Filed as tracker hygiene, not as new work.** The 2026-07-31 triage found two significant bodies of live work with no tracking artifact; the other is the 37-commit Embassy migration riding spike #198 (audited in a comment there). This is the second, and it is the **newest branch in the repo** — newer than everything except `main`.

## What exists

`feat/mesh-elect-consensus` @ **`6c98fb1`** (2026-07-29), 1 commit ahead of `main`, present on `origin`. **2,580 insertions across 15 files.** No PR, no issue, nothing referencing it.

It is the C3 half of a two-repo change (esp32c6-watch holds the C6 half), answering JP's ask: *"the mesh should vote and come to consensus on which ap and node to use as a gateway… a working mesh fast all the time"* — with no channel pinning.

**The reduction it makes.** The ESP32 has one radio and ESP-NOW rides whatever channel the WiFi STA is on, so choosing the AP *is* choosing the mesh channel; the whole partition problem reduces to making the fleet agree on one channel. The fleet therefore elects a **channel, never a BSSID** — enforcing a BSSID buys nothing and costs the driver its roaming freedom, i.e. it would reintroduce the pinning the change exists to remove.

**Vendored, not reimplemented.** `rust/mesh-elect/{src,tests}` are byte-identical to the watch's `crates/mesh-elect`, with sha256 provenance in its `Cargo.toml`. The stated reason is worth preserving: a wire-format or scoring mismatch between the two repos *does not fail loudly — it partitions the fleet silently*, so two hand-rolled parsers would drift where one copy with golden bytes pinned in both repos cannot. Upstream's 39 tests come along, plus 6 smol-side ones, because upstream's collision test claims to cover "both repos" while omitting nine real smol tags (`FAM`/`SNK`/`LDBG`/`ODEL`/`ODON`/`OTAD`/`OTAM`/`OTAN`/`UP`).

## What it deliberately does NOT do — the part that matters

[...]

PR #475, opened 2026-08-27
fix(build): gate the committed Cargo.lock — an exact '=' pin nothing enforced (#460)

jphein:
Closes the remaining half of #460. The lock itself was synced on main by the display lane; this is
the part that stops it drifting again.

## What it does

`tools/check_lock_fresh.sh` + one unconditional `gate.sh` arm + `tools/test_lock_fresh.sh` (11 arms).
No build invocation is touched.

`mipidsi` is pinned `=0.10.0` deliberately — s3_oled's ORIENTATION reasoning is tied to 0.10's MADCTL
semantics — and on 2026-08-26 the committed lock did not contain it at all, because nothing passes
`--locked` and so every build silently rewrote the file. The strictest pin cargo offers, enforced by
whoever last read the comment above it.

## Why an arm and not `--locked` on the build paths

#460's own option 2. Splicing `--locked` into `repro_build.sh` / `gate.sh` / `fw-gate.yml` changes how
everyone builds and collides with whichever lane holds those files. The checker's header says to
**delete this arm** if `--locked` is ever adopted there, so it cannot quietly become a second
statement of one fact.

## `--offline` is deliberately NOT used, and that is measured

All six lock/cache/flag combinations were run before choosing:

```
lock      registry cache   flags                 rc    verdict
healthy   warm             --locked --offline     0    fresh
STALE     warm             --locked --offline   101    stale
healthy   COLD             --locked --offline   101    *** FALSE STALE ***
healthy   COLD             --locked               0    fresh
STALE     COLD             --locked             101    stale
(cargo absent)                                   127   cannot check
```

A cold registry cache under `--offline` fails on a perfectly fresh lock — which on a clean CI runner
is the normal state. So **exit 101 alone is not the finding**: the checker requires cargo's own
`because --locked was passed` sentence before it will say STALE, the same discipline as the #420 arm's
refusal to read an unrelated build failure as proof the guard fired.

Three outcomes, not two: `0` fresh · `1` STALE (hard FAIL) · `2` could-not-check (printed loudly,
never a pass, never staleness).

## Scope, and why the limit is printed rather than implied

Exit-code-bearing for `rust/clock` only. The other Cargo.lock files are **discovered** (never a
hand-kept list — that is what #328 watched rot) and reported informationally, because nothing in-repo
states that `targets/*` locks must be fresh and failing on other lanes' actively-moving trees would be
inventing their requirements (#338).

## Evidence

`gate host` / `gate fw` / `gate excl` on familiar, all **GATE_RC=0**, with the arms named:

```
##### GATE host #####  GATE_RC=0    PASS lock fresh    PASS test_lock_fresh    GATE GREEN
##### GATE fw   #####  GATE_RC=0    PASS lock fresh                           GATE GREEN
##### GATE excl #####  GATE_RC=0    PASS lock fresh                           GATE GREEN
```

Self-test **11 ok, 0 failed**. Both sabotages fail it:

```
probe always returns FRESH        -> 10 ok, 1 failed
discriminator removed (any 101 = STALE) -> 10 ok, 1 failed
```

That second sabotage was caught by **nothing** until I added arm 4b, so the suite's most important
assertion was missing until I tried to break it.

## Four bugs found by running it, not by reasoning about it

Each is fixed and documented at the site rather than silently corrected:

1. An `EXIT` trap over a `local` work dir → `work: unbound variable` at cleanup.
2. `cp -r` then `rm -rf target` copies the build output first. Same end state, but gigabytes on a warm
   tree for a 0.13 s measurement. `target/` is now excluded *during* the copy.
3. The discovery source was set inside a function every caller pipes — the subshell swallowed it, so
   the "filesystem scan" warning would have printed on **every** run including normal checkouts.
   *A pipeline launders variable assignments the way it launders exit codes.*
4. My first version of the discovery arm asserted "this repo reports tracked" — a fact about the
   **environment**, not about the code — and it duly went red on familiar, whose rsync'd mirror
   legitimately has no `.git`. That is exactly the failure mode this checker's header is about,
   committed one level inside the fix for it. Now asserted against two constructed fixtures, so the
   property holds anywhere.

Bug 3 surfaced a genuine inconsistency worth knowing: in a `.git`-less build mirror the `find`
fallback reported three `experiments/*` locks the tracked path never shows — **gitignored generated
artifacts** (`.gitignore:76`) left by a previous host-verifier run. The two discovery paths are now
labelled rather than presented as equivalent.

## Honest limits

- The arm needs network on a cold registry cache (that is why `--offline` is out). A network outage
  reads as "could not check" — loud, non-failing — not as staleness.
- It does not prove any *build* consulted the lock; it proves the lock still describes the manifest.
  Adopting `--locked` on the build paths is the stronger property and remains unfiled work.

Routed to @team-lead, not self-merged (`tools/` + shared `gate.sh`).

### DIFF
--- a/tools/check_lock_fresh.sh
+++ b/tools/check_lock_fresh.sh
@@ -1,0 +1,288 @@
+#!/usr/bin/env bash
+# check_lock_fresh.sh — #460. Is the committed Cargo.lock actually consulted?
+#
+# THE DEFECT THIS EXISTS FOR: `rust/clock/Cargo.toml` pins `mipidsi = "=0.10.0"` — the strictest pin
+# cargo has, placed deliberately because s3_oled's ORIENTATION reasoning is tied to 0.10's MADCTL
+# semantics. On 2026-08-26 the committed lock did not contain `mipidsi` at all, and no build path in
+# the repo passed `--locked`, so every build silently REWROTE the lock and no gate noticed. An exact
+# pin enforced only by whoever last read the comment above it.
+#
+# WHY A GATE ARM RATHER THAN `--locked` ON THE BUILD PATHS (#460's option 2, deliberately): adding
+# `--locked` to `repro_build.sh`/`gate.sh`/`fw-gate.yml` changes how everyone builds and collides
+# with whichever lane holds those files. This arm catches the same staleness without touching a
+# single build invocation. If `--locked` is later adopted on the build paths, this arm becomes
+# redundant and should be deleted rather than kept as a second statement of one fact.
+#
+# ── THE INSTRUMENT, AND WHY IT IS SHAPED LIKE THIS ────────────────────────────────────────────────
+#
+# `cargo metadata --locked` resolves the FULL dependency graph (optional deps included — which is
+# the whole point, since #460's two missing entries were both optional) and refuses to rewrite the
+# lock. Fresh lock ⇒ exit 0. Stale lock ⇒ exit 101.
+#
+# ⚠️ `--offline` is NOT used, and that is measured, not stylistic. All six combinations were run:
+#
+#   lock      registry cache   flags                 rc    verdict
+#   healthy   warm             --locked --offline     0    fresh
+#   STALE     warm             --locked --offline   101    stale
+#   healthy   COLD             --locked --offline   101    *** FALSE STALE ***
+#   healthy   COLD             --locked               0    fresh
+#   STALE     COLD             --locked             101    stale
+#   (cargo absent)                                   127   cannot check
+#
+# A cold registry cache plus `--offline` fails on a PERFECTLY FRESH lock ("no matching package named
+# `ed25519-compact` found … offline mode can sometimes cause surprising resolution failures"). On a
+# fresh CI runner that is the normal state, so an `--offline` arm would have been red for a reason
+# that has nothing to do with the lock — a check that fails for the wrong reason, which is worse
+# than no check because it teaches people to ignore it (#338).
+#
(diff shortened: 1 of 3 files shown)


jphein:
## CI: GREEN, all three jobs — and a note on the red X above it

**Live run [33084496460](https://github.com/jphein/smol/actions/runs/33084496460) — `completed/success`,
event `pull_request`, headSha `40782cb`:**

```
tiers + clippy + stack floor   success   2m5s
tier exclusions (#351)         success   3m55s
host verifier suites           success
```

⚠️ **The other run on this PR (33084443694) is `cancelled`, not failed** — the superseded `push`-event
twin, same headSha, killed at 1m5s with its orphan `cargo`/`rustc` processes terminated mid-step.

Worth stating explicitly because **`gh pr checks` renders `cancelled` as `fail`**:

```
$ gh pr checks 475
host verifier suites   fail   1m5s   .../runs/33084443694/...     <- actually CANCELLED
```

`--log-failed` returns nothing for it, which is the tell: there is no failed step, the job was
terminated. This lane has now seen the superseded-twin pattern four times; the new detail is that the
twin doesn't merely look ambiguous in the CLI, it looks *broken*. The check that resolves it is
`gh run view <id> --json conclusion` — `cancelled` ≠ `failure`.

Gate evidence on familiar (from the PR body) stands: `host` / `fw` / `excl` all `GATE_RC=0`, with the
arms named rather than inferred from a green summary.

`git merge-tree --write-tree origin/main` re-run against current `origin/main`: **clean**.

Issue #464, opened 2026-08-26
OTA: fleet-wide staged lines make legacy crowns self-fetch foreign-chip images (~1 MB per stage, refused only at finalize)


jphein:
Found during the S3's first OTA roll (2026-08-26, #398). When an esp32s3 image was staged, the v922 C3 crown (id50) — which pre-dates the per-chip `smol/ota/staged/<chip>` topics — gated the FLEET-WIDE line (no target field), accepted it (1404>922), and **self-fetched the full ~1 MB S3 image** (`smol/50/ota/progress 1030992|1030992|self`), refusing it only at the finalize descriptor read (the #349 authoritative guard — which held, and `ota/diag mac-unknown retry=2` shows it retried). Every legacy crown pays ~1 MB of fetch + flash-write wear per foreign-chip stage.

Fix direction (smol-d8): retire the fleet-wide staged line once no pre-per-chip firmware remains in the fleet — an auditable condition (all boards ≥ the per-chip-aware build). Until then, staging foreign-chip images is safe but costs each legacy crown a wasted download.

Evidence: #398 roll thread; smol/50/ota/{state,progress,armdiag,diag} captures 2026-08-26 ~13:2x-14:1xZ.

jphein:
## Producer half implemented → **PR #500** (routed to team-lead, not self-merged)

`tools/ota_publish.sh` is not T-blocked, so the half of this that can be fixed today is fixed. All
three gate arms GREEN on familiar (`GATE_RC=0`), suite 56 → 65 arms, three sabotages caught.

**Narrower than this issue's stated fix direction, on purpose.** The body proposes retiring the
fleet-wide line *"once no pre-per-chip firmware remains in the fleet — an auditable condition"*. That
condition is **not met today** (id8/id50/id51 are v922, which predates the per-chip topics), so the
retirement cannot happen yet. But it does not need to: the legacy line only ever served pre-#349
firmware, **all of which is esp32c3** — the S3 and C5 targets were created after #349 (#398 and #388
cite it as prior art), so a pre-#349 non-C3 board has never existed. Publishing a foreign-chip build
there can serve nothing and can only cost a legacy crown the megabyte. So the line stays for exactly
the images a legacy board could install, and the fleet audit is not a prerequisite.

**Live confirmation that the hazard is being worked around by hand.** The retained state today:

```
smol/ota/staged            OTA|922|1031008|…          <- C3 build
smol/ota/staged/esp32c3    OTA2|922|1031008|…
[...]

PR #500, opened 2026-08-27
fix(ota): scope the fleet-wide staged line to the canonical chip (#464)

### LINKED ISSUE #472: HA update entity offers a downgrade when the fleet staged line trails the per-chip line

jphein:
Fixes the producer half of #464. Tools-only; no firmware change.

## The megabyte

Staging an esp32s3 image wrote its build to the fleet-wide `smol/ota/staged`. The v922 C3 crown
(id50) gates that line — it has no target field — accepted it (1404 > 922) and **self-fetched the
full ~1 MB S3 image**, refusing it only at the finalize descriptor read. Every legacy crown pays
~1 MB of fetch plus flash-write wear per foreign-chip stage, and pays it *before* finding out.

`pub_retained "smol/ota/staged"` was unconditional. The paragraph above it defends that as
*"UNCHANGED … a safe migration rather than a flag day"* — sound reasoning, entirely about pre-#349
**firmware**, and **chip-blind**. That blindness is the defect.

## Why skipping is safe — one fact, not a fleet audit

Firmware older than #349 knows only `smol/ota/staged` and only parses `OTA|`. **All such firmware is
on esp32c3**, because the S3 and C5 targets were created *after* #349 (#398 and #388 both cite it as
prior art). A pre-#349 non-C3 board has never existed and cannot.

So a foreign-chip build on the legacy line cannot serve any board: #349-aware boards of that chip
read their per-chip line, and non-#349-aware boards of that chip do not exist.

Deliberately **narrower** than this issue's own fix direction ("retire the fleet-wide line once no
pre-per-chip firmware remains"). That retirement needs a rolled fleet and an audit; this needs
neither, because it keeps the legacy line for exactly the images a legacy board could install —
including a **pre-#349 rollback image**, which carries no descriptor and for which the legacy line is
the only delivery path.

## ⚠️ This does not fix #472, and it changes #472's character

#472 is that a board's `ota/state` composer prefers the fleet line, so an S3 with a newer per-chip
line still advertises the fleet line's C3 build as `latest`. After this change the fleet line reliably
holds a canonical-chip build — which is what makes #472's consumer-side fix *well-defined* — but the
S3's `latest=<c3 build>` stops being the residue of a manual re-stage (#472 records the fleet line as
having been *"deliberately re-staged to 922"*) and becomes the **steady state**.

Traded knowingly: #472's own body calls its symptom cosmetic, because the monotonic ratchet refuses
the lower build. The megabyte is spent on every foreign-chip stage. Flagging it prominently because
this PR makes a tracked cosmetic wart permanent, and that should be a decision rather than a surprise.

## Shape

- **Pure decision** `legacy_line_wanted <target> <canonical>` — same reason `should_group_mac` is a
  pure function: the decision is unit-testable without a broker, a vault or an image.
- **The skip is LOUD.** A silent omission here would look exactly like a broker failure, which the
  install path's own 2026-07-28 note calls the worst failure this tool can have.
- **Canonical chip resolved from `build_matrix.py canonical-chip`**, never a literal. This file already
  carries the *fourth* copy of the chip roster, with a comment about how that copy was the one nothing
  compared.
- **New `legacy-line <chip>` preflight** — "will this go on the fleet-wide line?", answerable before
  spending a build, an scp and a vault read. It prints the canonical chip it *resolved*, not just a
  verdict, because a wrong answer from a wrong manifest lookup is the failure that would hide.

```
$ tools/ota_publish.sh legacy-line esp32s3
legacy-line: NO — 'esp32s3' is not the canonical chip (esp32c3); smol/ota/staged would be
             skipped (#464: it can only cost a legacy C3 crown a ~1 MB self-fetch).      rc=1
$ tools/ota_publish.sh legacy-line esp32c3
legacy-line: YES — esp32c3 is served by smol/ota/staged (canonical=esp32c3)               rc=0
```

## Evidence

`gate host` / `fw` / `excl` on familiar, all **GATE_RC=0**, with the suite named:

```
##### GATE host #####  GATE_RC=0   65 passed, 0 failed   PASS test_ota_publish_guards   GATE GREEN
##### GATE fw   #####  GATE_RC=0                                                        GATE GREEN
##### GATE excl #####  GATE_RC=0                                                        GATE GREEN
```

`build_matrix.py check` rc=0 (it asserts against this file). Suite 56 → **65** arms.

**Three sabotages, all caught:**

```
always-publish (i.e. the pre-fix behaviour)  -> 63 passed, 2 failed
refuse everything                            -> 61 passed, 4 failed
manifest lookup broken (canonical empty)     -> 59 passed, 6 failed
restored                                     -> 65 passed, 0 failed
```

The chips in the test come from the **manifest**, not literals — a hardcoded `esp32c3` there would be
a second statement of the same fact and would keep passing after the manifest moved. The
refuse-everything arm exists because a guard that refuses everything "can fail" and is useless.

Also widened `usage()`'s hardcoded `sed -n '2,27p'` → `2,28p` for the added MODES line. The file
already warns that this span is a duplicated fact **and that this suite pins both ends** — it does,
and it passes (`help reaches the end of the header block` / `help stops before the code`).

## Honest limit

The arms test the **decision**. The **call site** — that `stage` consults it before publishing — is
source-reviewed, not exercised: reaching that line needs a vault read, an scp and a live broker
publish, and a test that publishes to the real fleet topic is not a test. Stated in the suite itself
so section 6 is not read as proving the stage path.

Routed to @team-lead, not self-merged (`tools/`).

### DIFF
--- a/tools/ota_publish.sh
+++ b/tools/ota_publish.sh
@@ -13,6 +13,7 @@
 # MODES (Model-A #33: stage arms every board's native Update entity; Install is per-device)
 #   ota_publish.sh stage      [<commit>] [--bin <file>] [--build N]  # build+host+publish smol/ota/staged (arms all boards; NO board fetches)
 #   ota_publish.sh install <id>                                      # publish INSTALL → smol/<id>/ota/install (headless per-node canary; the HA Update button is the GUI path). id42 is REFUSED (#314: C6 watch unset-config sentinel, not a node).
+#   ota_publish.sh legacy-line <chip>                              # PREFLIGHT (#464): will an image for <chip> be published on the fleet-wide smol/ota/staged line? exit 0 = yes, 1 = skipped (non-canonical chip)
 # <commit> defaults to HEAD. --bin <file> skips the cargo build and hosts an existing .bin.
 # IDENTITY (#400): stage BUILDS THE WORKING TREE — it cannot build a commit's source. So a
 #   <commit> that is not HEAD is REFUSED (it would stamp that commit's identity onto these bytes),
@@ -88,7 +89,7 @@ die(){ echo "ERROR: $*" >&2; exit 1; }
 # ⚠️ The line range is a DUPLICATED FACT about the header block's extent — edit the header and it
 # rots silently (help output truncates mid-sentence, and nobody reads help on the happy path).
 # test_ota_publish_guards.sh pins both ends so an edit that forgets this fails instead.
-usage(){ sed -n '2,27p' "${BASH_SOURCE[0]}"; exit "${1:-1}"; }
+usage(){ sed -n '2,28p' "${BASH_SOURCE[0]}"; exit "${1:-1}"; }   # 27→28: #464 added a MODES line
 
 MODE="${1:-}"; [ -n "$MODE" ] || usage 1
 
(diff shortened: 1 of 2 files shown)


jphein:
**TEAM-LEAD DECISION on the #472 trade — ACCEPTED, with the consequence recorded rather than absorbed.** Pinning the fleet staged line to C3 builds is correct: every pre-#349 consumer is a C3 by construction, and the per-chip topics exist precisely so other families stop reading the fleet line as theirs. That #500 makes an S3's `latest=<c3 build>` steady-state is the honest price — it converts #472 from 'residue of a manual re-stage' into the *designed reconciliation point*: the ota/state composer preferring the per-chip line when newer (already ruled, T-blocked). Commenting on #472 to re-scope it accordingly. Merging on green.

jphein:
## CI GREEN — all three jobs, and again the red X above is a CANCELLED twin

**Live run [33086527796](https://github.com/jphein/smol/actions/runs/33086527796) — all three success:**

```
tiers + clippy + stack floor   pass   1m16s
tier exclusions (#351)         pass   3m30s
host verifier suites           pass   2m41s
```

The other run on this PR (33086482836) is **`COMPLETED/CANCELLED`** — the superseded `push`-event
twin, same headSha — which `gh pr checks` again renders as `fail`. Fifth time this lane has met the
pattern; the discriminator is `gh run view <id> --json conclusion`, and `--log-failed` printing
nothing is the tell.

Gate evidence on familiar (PR body) stands: `host` / `fw` / `excl` all `GATE_RC=0`, `65 passed,
0 failed`, `PASS test_ota_publish_guards` named rather than inferred from a green summary.

`git merge-tree --write-tree origin/main HEAD` re-run against current `origin/main`: clean.

Ready when you are — and please read the **#472 trade** in the PR body before merging rather than
after. It is the one part of this change that is a judgement call rather than a defect fix.

Issue #497, opened 2026-08-27
[matrix] docs: two stale Bard-budget claims — no declared chip currently fits the Bard


jphein:
**Found while filling the matrix's Bard row, by doing the arithmetic on `budget.rs`'s own constants.**

`ChipBudget::dram_headroom()` is `free_dram_bytes.saturating_sub(stack_floor_bytes)`, and `cost::BARD.dram_bytes` is 39,072 (measured: `.bss` +37,832, `.data` +1,232, alignment +8):

| chip | free DRAM | stack floor | headroom | vs 39,072 |
|---|---:|---:|---:|---|
| esp32c3 | 106,560 | 74,208 | **32,352** | short 6,720 |
| esp32s3 | 96,676 | 72,004 | **24,672** | short 14,400 |
| esp32c6 | 80,272 | 71,680 | **8,592** | short 30,480 |

The C3 figure reproduces the number `budget.rs` states for itself ("short by 6,720 B"), which is the check that the formula is being read right.

**Two claims in the tree are therefore stale:**

1. **`rust/clock/Cargo.toml`, at the `bard` feature:** *"The feature STAYS here: the S3 and C6 have the DRAM to carry it as an ordinary smol feature."* They do not, against today's rows — by 14.4 KB and 29.8 KB. The claim was plausibly true when written; `.bss` has grown since (the Embassy executor alone took 20,264 B out of `.stack` post-#391).
2. **`rust/clock/src/budget.rs`, inside the bard DRAM-refusal message:** *"the S3 has NO row yet — #398 — despite what an older revision of this message claimed."* `ESP32S3_CYD` landed with #411. A diagnostic message that misstates the tree is expensive precisely when someone is reading it.

**The consequence worth stating plainly:** the Bard is hardware-proven on glass (#300/#302, bit-for-bit against a Python reference) and is in **no shipped image on any board** — it builds only under the `off-fleet` waiver, which `repro_build_bin` then refuses to package. That is a coherent state, and #347 (extract the Bard to its own firmware) is the plan for it; what is not coherent is a comment saying two chips can carry it in the fleet image.

⚠️ **Both files are in the T executors' lane, so this is reported and deliberately not edited.** Also: this is arithmetic on published constants, not a bench measurement — worth one confirming read before anyone acts on it.

**Closes when:** both comments state the current numbers, or the rows change so the old claim becomes true.

Ref #347, #348, #411, #391, PR #474.

---
[...]

Issue #496, opened 2026-08-27
[matrix] releases: the per-target nightly ships no SHA256SUMS, but RELEASES.md and the docs told readers to check one


jphein:
**Found while writing the matrix, by checking the release instead of the doc.**

`gh release view nightly` (verified 2026-08-27) lists exactly ten assets — five `.bin` and five `.bin.NOTES.md` — and **no `SHA256SUMS`**. The previous release, `nightly-2026-08-24`, did have one (`docs/RELEASES.md:252-270` documents it as one of the four published files).

Meanwhile the reader-facing instructions said to verify against it:
- `README.md` quick-start ran `sha256sum -c SHA256SUMS`
- `site/index.html`'s copy-paste block ran the same

Both **would simply fail** for anyone who followed them. PR #474 fixes both to check the sha256 that each artifact's own `NOTES.md` carries inline (confirmed present — e.g. `smol-s3-cyd-…NOTES.md` leads with `**sha256** 4648a835…`).

**What still needs deciding, and is why this is an issue rather than just that fix:**
1. Is the omission intended? A per-artifact sha in `NOTES.md` is arguably better — it travels with the file and cannot be mismatched to a different release. If so, `docs/RELEASES.md:252-270` should stop documenting a combined file.
2. Or did `tools/release_targets.sh` lose it when per-target packaging landed (#413)? Then it should come back, and a one-line assertion that the release contains it would keep the docs honest.

Either way one of the two artifacts — the file or the doc — is currently wrong about the other. I did not touch `tools/` (outside my lane).

**Closes when:** the release and `docs/RELEASES.md` agree on how a downloader verifies an image.

Ref #413, #394, PR #474.

---
*Matrix-derived: filed from `docs/CAPABILITIES.md` (PR #474), read against `55c7ffe`. **The matrix cell is the acceptance test** — closing this means the cell changes. The watch lane's deeper sweep should dedupe against these rather than file fresh.*

jphein:
## Half of this is FIXED and verified by asset. The other half — the doc mismatch in this issue's own title — is still live, and I measured it.

### The asset half: fixed, 13 minutes after this was filed

This issue was filed **14:52 UTC**. The combined `checksums:` job landed in **`84c2ae8` (PR #499) at
15:05 UTC**, and its first run ([33086103549](https://github.com/jphein/smol/actions/runs/33086103549))
completed `success` with `combined SHA256SUMS: success`.

Verified against the **release**, not the job — a green job is a claim about a job:

```
[...]

PR #506, opened 2026-08-27
docs: state the verify command as a predicate, not an inventory (#496)

jphein:
Closes the last residual of #496, using the **predicate wording** you approved rather than a third re-sync.

## The problem

Three reader-facing docs assert *"there is no combined SHA256SUMS"*. #499 made that false.

**They were correct when written.** #474 wrote them while the per-target nightly genuinely shipped no
such file; #499 then added the `checksums:` job; #501/#502 fixed `RELEASES.md`, the workflow comment and
both NOTES templates — and README/site were never in scope. So right now:

- `docs/RELEASES.md` → *verify against the combined `SHA256SUMS`*
- `README.md` + `site/` → *there is no combined `SHA256SUMS`*

…and README/site are what a downloader reads **first**, while RELEASES.md is third.

## Why this is a predicate change and not a correction

This is the **third** correction round for one fact in one day, across five files, none of which can
check it. A fourth round is the default outcome of fixing it the same way again — so the sentence stops
asserting what the release contains:

> check the image against *that* number — or, if the release also carries a combined `SHA256SUMS`,
> against that with `sha256sum -c --ignore-missing SHA256SUMS` (the flag matters: you downloaded one
> image, not all of them).

True whether or not the checksums job ran, and **a workflow change cannot falsify it.** Same move as
#328 for the tag list, #338 for the fleet feature list, and the one #497 wants for `budget.rs`'s bard
message: stop enumerating, point at the thing that knows.

This **supersedes the #474-era prose**, which was accurate at the time — worth saying plainly so the
diff doesn't read as a correction of someone's error.

## All three files in ONE commit, deliberately

```
site/index.html:476   <!-- Visitors read content.json, NOT these defaults …
site/app.js:19        const r = await fetch('content.json', …)
```

Editing `index.html` alone changes **nothing a visitor sees**; editing `content.json` alone leaves the
two to **drift**. The duplication is itself a two-statements-of-one-fact hazard, and splitting the
commit would be walking straight into it.

The prose in the two is asserted **byte-identical after the edit** (281 chars, compared
programmatically, not eyeballed).

## Verification

- `site/content.json` patched by targeted string replace and then **`json.loads`-validated**. Not
  re-dumped — a dump would reformat the whole file and bury a one-sentence change in a total-rewrite
  diff.
- `grep -rn 'sha256sum -c' README.md site/ | grep -v -- '--ignore-missing'` → **empty**; no un-ignored
  form reintroduced.
- `git merge-tree --write-tree origin/main` → clean.
- 3 files, +7 −4. Docs/site only; no gate arm exercises these, so no build evidence applies or is
  claimed.

⚠️ One thing I could **not** verify and am not claiming: whether the rendered site looks right. `site/`
is served by `pages.yml` and I did not run it. The JSON is valid and the HTML change is a `<span>`-level
edit inside an existing paragraph plus one line in a `<pre>`, so the risk is low — but "low" is not
"checked."


jphein:
## CI GREEN — all three jobs `COMPLETED/SUCCESS`.

Checked by `conclusion`, not by the rollup badge, per the house rule — the accompanying
`COMPLETED/CANCELLED` triple is the superseded `push`-event twin, which `gh pr checks` would render as
`fail`.

```
tiers + clippy + stack floor   COMPLETED/SUCCESS
tier exclusions (#351)         COMPLETED/SUCCESS
host verifier suites           COMPLETED/SUCCESS
```

Restating the two verification claims from the body, since they are the ones a reviewer would want to
spot-check rather than take:

- **`site/content.json` and `site/index.html` carry byte-identical prose after the edit** — 281 chars,
  compared programmatically (`content.json`'s `fields.get.intro` sliced against the same span in the
  HTML), not eyeballed. This is the drift that `index.html:476` + `app.js:19` make invisible: only the
  JSON renders, so a divergence would sit there unnoticed.
- **`json.loads` passed on the patched file.** It was patched by targeted string replace and validated,
  never re-dumped — a dump reformats the whole file and buries a one-sentence change in a total-rewrite
  diff.

And the honest gap, unchanged from the body: **I did not render the site.** The JSON is valid and the
HTML edit is a `<span>`-level change inside an existing paragraph plus one line in a `<pre>`, so the
risk is low — but low is not checked.

PR #507, opened 2026-08-27
docs(s3-cyd): the [IMU] log is already honest — correct four stale doc claims (#480)

### LINKED ISSUE #480: [matrix] s3-cyd: `[IMU] OK` is a vacuous log line — the board has no IMU and the NACK is swallowed
**Matrix cell** (§5, *IMU*, `s3 fleet`): ❌ *no QMI8658 on the ES3C28P.* ⚠️ *`[IMU] OK` on this board is a vacuous log line — init is ungated and `let _ =` swallows the NACK.*

**Evidence read:** `targets/s3-cyd/PARITY.md:37-40` (gap 3) and `targets/s3-cyd/README.md:96-97`. The QMI8658 is absent from `BOARD.md`'s pin map.

**This is an honesty bug, not a feature request** — and it is the worst kind, because the instrument reports success for a peripheral that does not exist. Anyone reading the boot log learns something false. PARITY's own prescription:
- print by **result** (`OK` / `absent`), never unconditionally;
- gate consumers on `has-imu` (the Null-stub pattern already exists for touch).

The ❌ in the matrix stays ❌ afterwards — the hardware genuinely cannot. What changes is that the log stops claiming otherwise.

**Closes when:** a boot on id162 prints an honest absent-IMU line, and the matrix cell's ⚠️ caveat can be deleted.

Parent: #398.

---
*Matrix-derived: filed from `docs/CAPABILITIES.md` (PR #474), read against `55c7ffe`. **The matrix cell is the acceptance test** — closing this means the cell changes. The watch lane's deeper sweep should dedupe against these rather than file fresh.*

jphein:
**#480 was assigned to me as a code fix. It isn't one** — both halves of its prescription are already on
`origin/main` *in the subtree*, and the **docs** were what was stale. This corrects them. It does **not**
close #480 (see the closing condition below).

## The prescription is already met

```
targets/c6-watch/src/main.rs:1413-1415   match imu.init() {
                                           Ok(true)  => "[IMU] OK",
                                           Ok(false) => "absent (WHO_AM_I mismatch - no/other device)",
                                           Err(_)    => "absent (init NACK - no IMU on this board)" }
targets/c6-watch/src/apps/registry.rs:153  AppState::Maze => cfg!(feature = "has-imu")   // tilt-driven
targets/c6-watch/src/ui/slint_shell.rs:2673  …if cfg!(feature = "has-imu") { "IMU" } else { "no IMU" }
targets/c6-watch/Cargo.toml   board-esp32s3-cyd -> has-cap-touch only; has-imu correctly ABSENT
```

Landed on watch main as **`ae80072`** (2026-08-26) — *the day before #480 was filed* — and
`git merge-base --is-ancestor ae80072 e3872bb` confirms **#470's subtree refresh carried it.** The
delivery vehicle worked; only the prose lagged.

## Why the issue exists

The capability matrix was filled from `PARITY.md` and `README.md` rather than from the code, and both
still asserted *"init is ungated, `let _ =` swallows the NACK"*. **Both clauses are now false.** That is
the artifact-carrying-the-claim vs artifact-the-machine-consults split, with the doc as the stale one —
and `PARITY.md:59` had already ruled this *"document, not port"*, so the document was the entire
remaining task.

## Four sites — the fourth found by sweeping, not by the issue

| site | was |
|---|---|
| `targets/s3-cyd/PARITY.md:56` | gap 3 in the present tense → struck through with the landing commit |
| `targets/s3-cyd/README.md:96` | *"is a vacuous log line"* → honest-as-of `ae80072` |
| `docs/CAPABILITIES.md:238` | the ⚠️ rider #480 itself said should become deletable |
| `docs/CAPABILITIES.md:268` | the open-issues index still labelled #480 *"the vacuous `[IMU] OK`"* — which reads as a live gap in a list of live gaps |

My first sweep was **case-sensitive** and would have missed `PARITY.md`'s own uppercase `VACUOUS`; the
re-run with `-i` is what found site 4's neighbourhood. Recorded because it is the same class as
everything else this lane has been clearing.

## What this does NOT change

⚠️ **The ❌ in the IMU row stays ❌** and the exclusions stay exclusions. The hardware genuinely cannot;
what changed is that the instrument stopped claiming otherwise. **Nothing here upgrades a capability.**

⚠️ **It does not close #480.** Its *"Closes when"* asks for *"a boot on id162 prints an honest absent-IMU
line."* `ae80072` and the 1405 roll are both 2026-08-26, so I cannot say whether the image **running** on
id162 contains the fix. **Provenance is not fitness.** Every edited site carries that caveat explicitly
rather than implying the board was observed — one boot log settles it, and that line is left for whoever
is next at that bench.

## Verification

Every claim written into these docs was checked against `origin/main` before committing, not taken from
the upstream commit message: ancestry, the NACK line present in the subtree, `has-imu` absent from the
S3 arm, both consumer gates present. Sweeps for the two now-false clauses (`init is ungated`,
`swallows the NACK`) return empty.

Docs-only, 3 files, +15 −8. `git merge-tree --write-tree origin/main` clean. No gate arm exercises these
files, so no build evidence applies or is claimed.

**Lane note:** `targets/s3-cyd/*` is the s3-cyd lane's tree — this is a tense correction with nothing
added, and no watch-repo change or subtree refresh is involved, so the #480 fence does not apply and
`debug-watch-network-gaps` is untouched.


jphein:
## CI GREEN — all three jobs `COMPLETED/SUCCESS` (the `CANCELLED` triple is the superseded push twin).

```
tiers + clippy + stack floor   COMPLETED/SUCCESS
tier exclusions (#351)         COMPLETED/SUCCESS
host verifier suites           COMPLETED/SUCCESS
```

Docs-only, so CI is a formality here — the load-bearing verification is that **every factual claim
written into these docs was checked against `origin/main` before committing**, rather than taken from
`ae80072`'s commit message:

```
git merge-base --is-ancestor ae80072 e3872bb                    -> OK  (#470's refresh carried it)
subtree main.rs contains "absent (init NACK - no IMU on this board)"  -> 1
board-esp32s3-cyd arm contains "has-imu"                        -> 0   (correctly absent)
cfg!(feature = "has-imu") in registry.rs / slint_shell.rs       -> 1 each
grep -rni "init is ungated|swallows the NACK" targets/ docs/    -> empty
```

Two reminders of what this PR deliberately does **not** do, since both are easy to over-read from a
green tick:

- **The ❌ in the IMU row stays ❌.** No capability is upgraded; the hardware still cannot. What changed
  is that the instrument stopped claiming otherwise.
- **It does not close #480.** The *"Closes when"* asks for a boot log off id162, and `ae80072` and the
  1405 roll are the same day — so whether the **running** image carries the fix is unverified.
  Provenance is not fitness, and every edited site says so in place rather than leaving a reader to
  infer the board was observed.

Issue #485, opened 2026-08-27
[matrix] c5-cyd: no measured ChipBudget row — the one thing pinning builds = false


jphein:
**Matrix cell** (§0): `ChipBudget row` ❌ *none — this is what pins `builds = false`*; `Stack floor` 🔶 *inherits the C6's 71,680 B assert, labelled PROVISIONAL … a stand-in, not a fact*; `Floor provenance` ❌ *none of its own*.

**Evidence read:** `tools/build-matrix.toml` `[chip.esp32c5]` — `blocked_on = "no measured ChipBudget row yet (budget.rs hands an unmeasured chip the poison row); measure on hardware per #388, then flip builds"`. `rust/clock/src/budget.rs`'s `UNMEASURED` row is deliberately poisonous (zero DRAM, zero slot, `chip: "unmeasured"` so it reads as a bug if it ever reaches a dashboard). `targets/c5-cyd/target.toml` and `targets/c5-cyd/README.md:16-19` both state it, the latter with the line worth repeating: *"It compiles" is a real claim and a much weaker one than "it builds"; rounding one to the other is the mistake this row exists to prevent.*

Also note `targets/c5-cyd/target.toml`'s own warning about the GUI artifact: its stack floor is the C6's number standing in, and `src/main.rs` labels it *"PROVISIONAL … a stand-in, not a fact (budgets are measured, never inherited)"*. **Bracket it via the radio-up soak before anyone reads that number as measured.**

**Closes when:** `ESP32C5_*` consts exist in `budget.rs` from a real measurement, `[chip.esp32c5] builds` flips true, and §0's budget/floor/provenance cells for `c5 GUI` stop being ❌/🔶.

Parent: #388.

---
*Matrix-derived: filed from `docs/CAPABILITIES.md` (PR #474), read against `55c7ffe`. **The matrix cell is the acceptance test** — closing this means the cell changes. The watch lane's deeper sweep should dedupe against these rather than file fresh.*

jphein:
**Triage:** **BENCH-B** · owner **c5-cyd lane** · effort **M** · **highest leverage of the three bench sittings.**
> **Waits on:** a measured ChipBudget row. This is the single thing pinning `[chip.esp32c5] builds = false`, so it gates the C5 target's graduation from *"it compiles"* to *"it builds"* — a distinction that README puts well and that rounding one to the other is exactly what the poison `UNMEASURED` row exists to prevent. Shares a sitting with #486/#488 (the same board, the same bench).

[...]

PR #516, opened 2026-08-27
docs(c5): retire "checks-only" — the C5 fleet image links and BOOTS (#485)

jphein:
Retires the C5 "checks-only" family, which #485's hardware run made false by artifact, and adds the
geometry condition @team-lead ruled for the eventual fleet row.

## The evidence

#485, id176, 2026-08-27: a **linked** C5 fleet image boots from `ota_0`, joins WiFi, runs its **own**
NTP (`tsrc=ntp` — the GUI image only ever managed `tsrc=mesh`), meshes with the live fleet, and
registers PSRAM (`heap=8,426,036`).

## ⚠️ Eight sites, not the two in the assignment

A case-insensitive sweep for `checks-only` / `check-proven` / `no fleet image` found them. Fixing the
two named would have left **six** standing, including the machine-read manifest:

| file | what was stale |
|---|---|
| `targets/c5-cyd/target.toml` | the CHECK-proven claim **and** "its FLEET arm is still checks-only" |
| `targets/c5-cyd/README.md` | "status: checks-only" **and** "CHECK-proven, not LINK-proven" |
| `README.md` | "for `rust/clock` it is checks-only … no linked image yet" |
| `docs/CAPABILITIES.md` | "has no fleet column because it has no fleet image" **and** the published-download rider |
| `docs/RELEASES.md` | the `c5-cyd` row |
| `tools/build-matrix.toml` | "nobody has measured this silicon" **and** `blocked_on` |

## Each claim has two parts, and only one is false

**"It links" → now false-by-artifact. "No measured `ChipBudget` row, so `builds = false`" → still true.**

Every edit retires the first and preserves the second. **`artifact` and `builds` both stay `false`** —
what the image lacks is a **floor gate**, not a linker. #485's own comment is explicit about this
(*"Still not floor-gated, so `artifact = true` does not follow yet"*), and I have not let the good news
leak into the gate.

## ⚠️ The manifest was the important one

`tools/build-matrix.toml` said **"nobody has measured this silicon"** and told the reader to *"measure
on hardware per #388"* — **a bench sitting that has already happened.** `blocked_on` now carries #485's
four named gaps instead:

1. re-measure via `repro_build_bin` (the C3's row came from the packaging path; #485's came from a plain link),
2. no high-water instrument on this chip (`hmin=0`) ⇒ `Derived` unavailable,
3. `ObservedSufficient` needs the downward soak bracket — 76,000 B is what it **ran with**, not the smallest proven,
4. the `ChipBudget` console printout needs the board's native-USB cable.

**A stale blocker that sends someone to redo finished work is worse than no blocker**, and this one
would have cost a bench sitting on a board whose sittings are already the scarce resource.

## The geometry condition (team-lead's ruling)

Added to `target.toml` beside the flip criterion: the floor was measured in the **6 MB watch partition
table this silicon physically carries**, and would be declared at canonical **4 MB** slot geometry —
**conservative by construction** (smaller declared slot ⇒ stricter gate; the image fits at 56.15%,
within a point of the C3's 56.9% for the same tier). **Re-measure if a true 4 MB-table C5 fleet build
ever ships** — the two are only interchangeable while the declaration is the stricter of the pair.

## #485's headline folded into the matrix cell

Per "tree wins": measured `free_dram` **76,000 B** against the inherited PROVISIONAL 71,680 B floor
leaves **4,320 B of `dram_headroom` — the least in the tree** (C3 32,352 · C6 watch 8,592), with the
four gaps named.

## Deliberately NOT touched

- `README.md`'s label-vocabulary **definition** of "checks-only" — still correct, and it is the thing
  that makes the other rows readable.
- `docs/RELEASES.md`'s **c6-watch** row — different chip, still true (it genuinely cannot LINK without
  the watch's `widen_rom_region` hook).

## Verification

- `python3 tools/build_matrix.py check` → **rc=0**; `canonical-chip` and `chips` still resolve. Only a
  comment and a prose field changed — **no key touched** — but the manifest is machine-parsed, so this
  was checked rather than assumed.
- `tools/status_check.sh` → **21/21, rc=0** (re-run because `RELEASES.md` and `README.md` are in it).
- Re-swept after editing: the only remaining hits are the two intentional survivors above and my own
  new text.
- Added lines scanned for LAN/secret strings: clean. `merge-tree`: clean.

**One citation note:** the assignment pointed at `scratch/cyd-c5-smol-port/chipbudget-c5-measured.md`.
That path is **untracked and not present in the repo**, so every citation here is to **#485's comment**
instead — a doc citing a path that resolves for nobody is not a citation.

`docs/RELEASES.md` touched — flagging for the #413 lane, though their PRs are merged and this is a
one-row factual correction.


jphein:
## CI GREEN — 3/3 `COMPLETED/SUCCESS` (the CANCELLED triple is the superseded push twin, as usual).

The load-bearing verification here isn't CI — it's that the manifest still parses after a prose edit:

```
python3 tools/build_matrix.py check     -> rc=0
python3 tools/build_matrix.py canonical-chip / chips  -> still resolve
tools/status_check.sh                   -> 21/21, rc=0
```

Only a comment and a `blocked_on` string changed — **no key touched** — but `build-matrix.toml` is machine-parsed, so that was checked rather than assumed.

Issue #237, opened 2026-07-20
design: peer-sourced leaf-mesh-OTA — an updated node relays the image to the next (extend #40)


jphein:
## Summary
Extend the #40 leaf-mesh-OTA so an **already-updated node** can be the ESP-NOW **source** of the image for the next node — image travels node→node over the mesh, instead of always originating from the gateway's WiFi bulk fetch. This is the **ADOPT** candidate from the #54 reference-architectures study (docs/superpowers/research/reliable-mesh-ota-architectures.md, PR #235): smol's leaf relay already *is* the ESP-NOW-native OTA pattern; peer-sourcing is the one worthwhile delta the survey surfaced.

## ✅ PROMOTED — the gate is resolved: this is now the STRUCTURAL fix, not a nice-to-have
The #54 study filed this **gated** on whether the coexist-bulk-OTA disease even reproduces at the radio level on the new esp-radio 0.18 + coex stack. **That gate is now resolved — and it resolved the "load-bearing" way.**

The sibling coexist study (**merged: docs/superpowers/research/coexist-disease-esp-radio-018-study.md**) verdict is **YES — still bulk-RX-deaf on esp-radio 0.18 + coex**: minutes after crowning, a node goes bulk-RX-deaf while TX + broadcast RX (small MQTT/MC frames) survive. So moving to the Embassy/#198 stack does **not** cure the disease — meaning a WiFi bulk fetch on an active mesh node stays fundamentally hostile to the mesh, on both the current and the future stack. **Peer-sourcing is therefore promoted from "nice-to-have optimization" to the structural fix** for the last coexist window: the only durable way to stop re-incurring the fetch is to stop fetching on all-but-the-first node and source the image peer-to-peer over ESP-NOW instead.

### Packet-level evidence (tonight's v343 roll)
Live pcap on the image host during a v343 leaf-OTA roll ([#204 comment](https://github.com/jphein/smol/issues/204#issuecomment-5018759525)) caught the disease exactly at the fetch leg:
- Board completes the TCP handshake to :8087, sends its 94 B Range GET; the server pushes the 230 B header + 726 B first data segment **~0.6 ms later** — and the board **never ACKs a single response byte**. Server retransmits into a void on exponential backoff. Two boards showed the identical signature concurrently.
- **Perfectly asymmetric:** board→server (SYN, ACK, GET) all pass; server→board **dies at response-byte-0**. It's an RX-deafness on the *download*, not a link failure.
[...]

Issue #19, opened 2026-07-08
fix: per-board NODE_ID dirties the version sigil — move it to git-ignored config


jphein:
Flash-day provenance wart (benign, but misleading): the per-board `NODE_ID` lives in `main.rs` (a TRACKED file), so building each leaf with `NODE_ID=8`/`9` dirties the tree → the realm-sigil version derives a *dirty* hash and bumps the build number. Result: id8 booted `build 43 "Keen Kiln" (f6d56d2)` and id9 similar, while id7 (only its git-ignored `secrets.rs` changed) stayed clean `build 42 "Bolted Ingot" (96f44d5)`. **The flashed leaf code is genuinely 96f44d5 + NODE_ID + creds-withheld — only the embedded version string is dirty-derived**, so the fleet *looks* like it's running three different builds when it's one.

Fix: move `NODE_ID` to the git-ignored per-clone config (`secrets.rs`, where the broker creds + WiFi creds already live, or a dedicated `board.rs`), with a placeholder in `.example`. Then per-board builds don't touch a tracked file and every board reports the same clean commit hash. Pairs naturally with #18 (`DEFAULT_APP`) — same "per-board config belongs in the git-ignored layer" principle; consider co-locating them.

Not a functional bug (mesh, roles, displays all verified correct this flash). Provenance/observability only.

jphein:
Correction: the firmware agent's initial 'dirty tree bumps the version' diagnosis was **retracted after verification**. The two commit stamps (id7 `96f44d5`, id8/id9 `f6d56d2`) differ only because HEAD advanced to the HA-side commit mid-flash — `git diff 96f44d5 f6d56d2 -- rust/clock/` is EMPTY, so all three boards run byte-identical firmware. The sigil reports HEAD's clean hash, not a dirty marker, so there is no misleading-provenance bug.

What remains is only mild operational hygiene: editing tracked `main.rs` per board forces an edit→build→`git checkout --` restore dance each flash (error-prone). Moving `NODE_ID` to the git-ignored config layer (alongside creds, with `DEFAULT_APP` from #18) would let per-board builds touch zero tracked files. Downgraded to minor; not blocking anything.

jphein:
[...]

Issue #513, opened 2026-08-27
cleanup(net): retire tx_announce_bounded's block_on once STEP T lands — the Timer belongs under a real executor waker


jphein:
Post-merge cleanup, named so it does not evaporate. **Blocked on: #335 STEP T merging.** Not urgent — the brick-class panic is fixed on `main` by the shape-(b) deadline poll-loop; this replaces that workaround with the structural fix once T's async surface exists.

## Background

`net/mode.rs::tx_announce_bounded` bounds an ESP-NOW announce by racing the send against a timer:

```rust
let fut = self.esp_now.send_async(&dst, frame);
block_on(select(fut, Timer::after(TX_WAIT)))
```

That panics at `embassy-executor-0.10.0/src/raw/waker.rs:35` — *"Found waker not created by the Embassy executor"* — whenever the `Timer` is actually polled. Verified in the dependency sources:

- `embassy_futures::block_on` builds its **own** `RawWakerVTable` over a null pointer (`block_on.rs:6`), so `task_from_waker` rejects it by construction.
- `embassy_futures::select` polls **A first and returns without polling B if A is `Ready`** (`select.rs:64-67`).

So the `Timer` registers **only when `send_async` is `Pending` on its first poll** — which is why it is intermittent, and why it surfaces after a relay fetch, when the radio is most contended. `embassy-time` here carries only `features = ["log"]`; no `generic-queue`, the escape hatch the panic message names.

## Why this issue exists separately from the fix

The immediate fix ships shape **(b)** — an `esp_hal::Instant` deadline poll-loop — because on `main` `run_leaf_ota_relay` is **sync**, so async-ifying `tx_announce_bounded` there would cascade a signature change through a brick-class path for a hotfix. Correct call.

**But shape (b) reintroduces a poll-loop to avoid a timer that only fails because of where it is polled.** After STEP T, `run_leaf_ota_relay` is `pub async fn` and both call sites are already async, so the structural fix becomes a small local change.

## The change

```rust
async fn tx_announce_bounded(&mut self, frame: &[u8]) -> TxOutcome {
    …
    let outcome = select(fut, Timer::after(TX_WAIT)).await;   // no block_on
    …
}
```

[...]

PR #522, opened 2026-08-28
docs(release): correct the runbook — the build number is an OUTPUT, not an input (v1446 Molten Gear)

jphein:
**My error, caught by the runbook's own step 1 during the canary — and wrong in the direction that ships
a mislabelled release.** Urgent because task #19 is live.

## What was wrong

The first draft said *"bump `version.txt` 345 → 346, and v346 is `Riveted Gear`."* Both halves were
internally consistent and **jointly wrong**:

```
build.rs:72          env_or_file("SMOL_BUILD_NUMBER", "version.txt")     <- ENV WINS
repro_build.sh:406   export SMOL_BUILD_NUMBER="$number"
ota_publish.sh       BUILD = choose_build(count, staged, override)
                           = max(git rev-list --count, retained staged + 1)
```

On the release path **`version.txt` is the fallback for non-stage builds, not the number that ships.**
The ratchet would have stamped its own number while `RELEASES.md` said v346.

## The step is INVERTED, not patched

**Learn the number from the ratchet → derive the word from *that* number → set `version.txt` to agree.**
The number is an *output*, so it cannot be chosen first. That ordering is now the guard at the top of §1.

**This release is `v1446` — `Molten Gear`.** Computed: `1446 % 20 = 6 → "Gear"`,
`(1446/20) % 20 = 12 → "Molten"`, with all four documented controls reproducing (341 Bellows,
342 Crucible, 345 Riveted Furnace, 905 Flux Furnace).

## ⚠️ The lesson, written into §1, because the controls did NOT save me

Four naming controls verify the **word against the number**. They say nothing about whether the
**number** is right.

`DOC-UPKEEP`'s rule — *"name it WITH its sigil word so the pair self-checks"* — catches a **mismatched
pair**. It cannot catch a **correctly-derived word on the wrong number**, which is self-consistent and
wrong. **A self-checking pair checks the RELATION, not the INPUTS.**

That is the **second time today** the same shape got me: with #237 I checked an issue's neighbours and
dependencies and never checked whether the thing had *shipped*. Both times, thorough verification of the
wrong proposition — which is the failure mode that survives being careful, and is why it is in the
runbook's provenance section rather than only in a commit message.

## Also fixed — all mine, all found by re-sweeping after the first pass

- **The file name.** `RELEASE-RUNBOOK-v346.md` → `-v1446.md`. A v346-named runbook for a v1446 release
  is exactly the mislabelling it exists to prevent. `RELEASES.md`'s See-also pointer follows it.
- **§3 said** *"the thing that makes v346 v346 is the NUMBER in `version.txt`"* — **doubly wrong**: wrong
  number, **and** it repeated the very error §1 now corrects.
- **§6 and the preamble** quoted `RELEASES.md` text that this commit changes.
- `RELEASES.md`: header row and ceremony heading now name **v1446 Molten Gear** (row says *in canary*;
  flipping it to *cut* stays part of cutting the release).

## Deliberately left

Three **dated** references — the peer-sourced design doc's *"v346+ rolls"*, `HANDOFF-2026-07-28`, and
`embassy-ota-verification`'s *"the v346 wave"*. Those are records of their moment, not live operator
instructions; same call the ROADMAP de-rot made about snapshots. Flagging rather than silently editing
another lane's design doc for a shorthand.

## Verification

`tools/status_check.sh` **21/21 rc=0** · re-swept: the only surviving "v346" in the runbook is the two
lines that **record** the first draft's error · `merge-tree` clean · selective staging (the `git add -A`
guard fired on my first attempt and was right to).

Routed to @team-lead — and worth landing ahead of the canary's next step, since the runbook is what
task #19 executes from.

### DIFF
--- a/docs/RELEASE-RUNBOOK-v1446.md
+++ b/docs/RELEASE-RUNBOOK-v1446.md
@@ -1,18 +1,18 @@
-# v346 release runbook — the checklist, not the rationale
+# v1446 release runbook — the checklist, not the rationale
 
 > ## 🟡 PREP ONLY. Nothing in this file fires before JP's go.
 > This exists so that when STEP T merges and the paint bound passes, task #19 **executes from a
 > checklist instead of being reconstructed**. It is written ahead of the event on purpose.
 
-> ### ⏳ DELETE OR SUPERSEDE THIS FILE ONCE v346 IS CUT.
-> Its v346 specifics become history the moment the release exists, and anything in it that turns out
+> ### ⏳ DELETE OR SUPERSEDE THIS FILE ONCE v1446 IS CUT.
+> Its v1446 specifics become history the moment the release exists, and anything in it that turns out
 > to be *generally* true about versioned releases belongs in `docs/RELEASES.md` instead. A
 > version-stamped runbook with no expiry is how a repo accumulates procedures nobody can date — the
 > same reason `xtensa-spike.yml`'s schedule block carries a deletion condition.
 
 ## What this file deliberately does NOT restate
 
-`docs/RELEASES.md` § *"The versioned release ceremony (`v346` will be the first)"* already covers the
+`docs/RELEASES.md` § *"The versioned release ceremony (`v1446 Molten Gear` is the first)"* already covers the
 **why**: the repro-build property (a fixed `(commit, node-id)` builds to the same bytes, so the sha256
 *is* the identity), #44's two causes (rustc's absolute paths, the wall-clock app descriptor), the
 `repro_build.sh`-is-a-sourced-library trap, and sigil names being history that is never re-synced.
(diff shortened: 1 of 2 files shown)


jphein:
## ⚠️ Force-pushed — **the first version of this commit's message LIED**, and the rule it broke is one this repo already learned.

I ran `git add <new-path> <old-path> <other> 2>/dev/null` where `<old-path>` was the file the rename had just removed. **One bad pathspec fails the WHOLE add**, and the redirect ate the signal:

```
$ git add docs/RELEASE-RUNBOOK-v346.md
fatal: pathspec 'docs/RELEASE-RUNBOOK-v346.md' did not match any files
exit=128
```

So the commit landed with **the rename and RELEASES.md but WITHOUT the 50-line body correction it described** — a commit message describing edits that were not in it, which is precisely the failure that rule exists for.

**What caught it:** `gh pr create`'s `Warning: 1 uncommitted change`. **What I had misread:** `git status --short` showed `RM` and I read it as "staged rename + staged modification" — the `M` is the **worktree** column. `git diff --cached --stat` is the check that can't be misread that way, and it's what I used to confirm the fix.

Amended, force-pushed with lease, and verified **from the commit rather than the worktree**: 54 insertions / 20 deletions across both files, §1's corrected guard present, §3 and §6 corrected, `RELEASES.md` naming v1446 Molten Gear.

Recorded in the commit message rather than quietly fixed, because a silent amend here would have left exactly the artifact this PR is about: a document that says one thing while the tree says another.

jphein:
## For the structure follow-up: the ruling is grep-shaped, and its baseline is already 1

@morpheus-stepT2's suggestion, verified from the tree before passing it on.

The content corrections here are better than the ones I'd drafted — the §1 inversion (*the number is an **output**: learn it from the ratchet → derive the word → set `version.txt` to agree*) is the right framing and I dropped my duplicate wording in its favour. Two structural halves are outstanding, and both are worth doing **as a check rather than a promise**:

### The two outstanding items

1. **Filename still carries a version** — `RELEASE-RUNBOOK-v346.md` → `RELEASE-RUNBOOK-v1446.md`. The ruling was *"no version in the filename"*. **The defect was incremented, not removed**: a fix carrying a version number is a fix with an expiry date nobody scheduled. The symptom is gone; the mechanism is intact with its clock reset.
2. **The deletion condition survived** — line 7 now reads `⏳ DELETE OR SUPERSEDE THIS FILE ONCE v1446 IS CUT.` That is the sharper half, and it is worse than stale: it is **true at the moment of reading, inside the document being used to cut v1446**. It doesn't rot at some future date — it was already false-in-practice on the day it landed. *A condition satisfied at the moment it is needed is not a condition.*

### The gate, because this ruling was mechanically checkable all along

```
"no version in the filename"
  →  no file under docs/ may carry `-v[0-9]{3,4}` in its name
```

**Verified baseline on `main`: exactly 1 violation** — this file. One to fix, then zero forever, and the check is a `git ls-tree | grep -c`.

**And I checked it doesn't over-match**, which is the part that makes it safe to adopt rather than a nuisance:

```
docs/superpowers/HANDOFF-2026-07-28.md              ← date, not a version: NOT matched
docs/superpowers/plans/2026-07-26-bard-tiny-llm.md  ← date: NOT matched
docs/superpowers/research/…/vesper-implementer-t0.md ← "vesper" is not v+digits: NOT matched
docs/superpowers/research/bard-300-agent-reports/…   ← 300 without a `-v`: NOT matched
```

Same argument as #513's `block_on`/`Timer` arm: **a prose ruling that is mechanically checkable and left unchecked will be re-violated by whoever renames the file next — and the rename is exactly the moment nobody is thinking about the ruling.** This one is unusually cheap because the baseline is already 1.

### Also worth carrying

The `git add` with a dead pathspec that made this commit's message lie before the amend — I hit the identical thing in my own tree ~40 minutes earlier (`git add docs/RELEASE-RUNBOOK-v346.md` after a `git mv` had already staged it). **Mine exited 128; yours partially succeeded and produced a lying commit.** The usual statement of the rule is "one bad pathspec fails the whole add" — but the real danger is that it *sometimes* fails: the loud case trains you to expect noise, and the quiet case arrives wearing that expectation. **A failure mode that is usually loud is more dangerous than one that is always silent**, because it manufactures the confidence that hides it.

jphein:
## Addendum — the stronger argument for the gate, from @morpheus-stepT2

The gate suggestion above was justified as tidiness ("a stale name is untidy"). **That undersells it, and the stronger form matters because it is what survives a reviewer asking "why bother".**

**The versioned filename and the pre-announced release name are the same category error:** an artifact created *before* the cut carrying an identifier produced *by* the cut.

- `RELEASE-RUNBOOK-v1446.md` is that error in a **filename**.
- "Molten Gear" written into `RELEASES.md`, the runbook's first-run record, and a test comment *before staging* is that error in **prose**.

Both are **outputs of the cut smuggled into inputs of the cut**. So the ruling *"no version in the filename"* was never hygiene — it was enforcing a category boundary, and the gate enforces the boundary rather than the tidiness:

> **Nothing written before the cut may name the cut.**

### The practical corollary — the line worth putting in the runbook

> Before the freeze, refer to the release by something **stable** — the commit sha. The derived name is a **postcondition of the cut, not a label available to plan with.** Any document that must exist beforehand refers to the sha; the name is filled in afterwards, or not at all.

That is §1's inversion (*the number is an output*) applied one layer up: **not only is the number an output, nothing derived from it can be used to plan the thing that produces it.**

### Live evidence, since this stopped being theoretical mid-thread

```
rev-list --count origin/main   1446 → 1447   (this PR merging advanced it)
version_name_for(1446) = Molten Gear
version_name_for(1447) = Molten Hammer
version_name_for(1448) = Molten Ingot
```

Nouns land on distinct indices, so each merge changes the **noun**, not just the adjective — it is a different name per merge, not drift within a family. **The release name was pre-announced three times after being ruled an output**, and the sharpest case: merging the test that *documents* the name would have advanced the count and changed it. The document describing the name changes the name.

`main` is frozen at `d1c1da5` for the cut; the number and word will be **read from `choose_build`'s output in the stage log**, and every document naming them corrected afterwards.

Issue #382, opened 2026-08-02
diag: leaf-path truncation makes the elect= field invisible for leaves — the observe rollout can't be observed where it matters


jphein:
Found live during the 917 roll (2026-08-02): the observe-posture rollout of #269/#278 is **unreadable for leaves** — which is most of what it exists to observe.

**Leaf DIAG records reach the broker via the crown relay, truncated (`cut=103`-style); the `elect=` field is appended near the record's tail (before the ledger, per the shed-order discipline) and never survives the cut.** Crown `elect=` publishes fine (the crown's own full record). Observed: id8 confirmed on 917, its relayed diag carries no `elect=`/`apch=`/`cc=`/`mo=`/`mf=` at all — so the cross-board beacon proof (`n` climbing on a leaf) cannot be read from retained topics, and the pre-move burst proof (`m`) will be equally invisible on leaves.

This is the known leaf-path truncation residual (#306) biting the new field — recorded here so the soak doesn't silently produce nothing.

Constraints on any fix: the DIAG budget (495 B cliff) and machine-checked shed order still bind; moving `elect=` earlier in the record changes shed semantics for the fields it displaces. Options: a dedicated tiny relayed field for leaves (epoch+n+m only); widening the leaf relay chunk; or leaves publishing `elect=` on their own broker session when associated (not all leaves have one). Also note the indirect signal that DOES work today: crown `mo=` (MAC-ok inbound) climbing is evidence the 917 leaf's signed frames verify at the crown — necessary but not sufficient for the ELECT path specifically.

jphein:
## The constraint that decides the option space: shed order is the WRONG AXIS for this

Numbers first, from source rather than the design record:

- `OBS_VALUE_MAX = ESP_NOW_MTU(250) − MAC_TRAILER_LEN(9) − DIAG_PREFIX(12) − 3` = **226 B**
- minus `CUT_TAG_MAX` (8, for `|cut=999`) → the field-boundary room is **218 B**
- `DIAG_BUDGET = 512 − 4 − len("smol/255/diag")` = **495 B**

**So a relayed leaf's DIAG is a 218 B window onto a record built to 495 B.** With id8's observed `cut=103` its full record was ~321 B, of which 218 arrived. A leaf's DIAG rides *one* ESP-NOW frame — the source calls the cap "a HARD bound", there is no fragmentation on this path — so this is structural, not tuning.

### The part I got wrong, and it matters for whoever fixes this

[...]

Issue #367, opened 2026-08-01
latent: esp-radio 0.18 discards ScanConfig.max — scan allocates one AccessPointInfo per visible AP (bound unrestorable in 0.18)


jphein:
**`b2537c4` added `.with_max(16)` to bound scan-result allocation and prevent an OOM panic. In esp-radio 0.18.0 that setting is silently ignored, so the guard is a no-op and the panic risk it was written to remove is still live.**

Found while researching #337. Not a design question — a live crash risk on `main` today.

## The guard

`rust/clock/src/net/mode.rs:3196-3200` (and identically at `:3252-3256`):

```rust
// #233 REGRESSION FIX: esp-radio 0.18's scan_async(ScanConfig::default()) has
// ... OOM-panic risk (alloc failure = rst=panic); .with_max(16) restores the bound.
.scan_async(&ScanConfig::default().with_max(16)),
```

The reasoning is right: `scan_async` returns a heap `Vec<AccessPointInfo>`, an unbounded scan in a dense RF environment can fail the allocation, and on this target an alloc failure is `rst=panic`.

## Why it does nothing

`ScanConfig.max` is never consulted on the `scan_async` path in `esp-radio-0.18.0`.

`src/wifi/mod.rs`, `scan_async`:

```rust
esp_wifi_result!(wifi_start_scan(false, *config))?;
// ... await ScanDone ...
Ok(ScanResults::new(self)?.collect::<Vec<_>>())
```

- `wifi_start_scan` builds a `wifi_scan_config_t` (`mod.rs:1137-1152`) — a struct with **no result-count field**. The cap can only be applied when *reading* records back.
- `ScanResults::new` never receives it — `src/wifi/scan.rs`:

```rust
pub fn new(_controller: &'d mut WifiController<'_>) -> Result<Self, WifiError> {
    ...
    let mut bss_total = 0;
    unsafe { esp_wifi_result!(include::esp_wifi_scan_get_ap_num(&mut bss_total))? };
    this.remaining = bss_total as usize;   // <- the FULL driver count
```

The controller parameter is `_`-prefixed. `remaining` is the driver's total AP count, and the `Iterator` walks all of it. **`config.max` is dropped on the floor**, so `.collect::<Vec<_>>()` allocates one entry per AP the radio saw — exactly the unbounded allocation the guard was added to prevent.

## Upstream: known, fixed, and the fix is not in 0.18.0

[...]

PR #528, opened 2026-08-28
fix(diag): #471 split mac_fail into a benign class and an ACTIONABLE one — and make the actionable one exact

### LINKED ISSUE #471: diag(#190): mac_fail folds 5 causes into one counter — Wave 1's 'mac_fail≈0' gate is unevaluable AND unreachable (63.8% / 90.7% measured)
## Summary

**`mf=` (`mac_fail`) folds five distinct causes into one `u32`, and #190's Wave-1 gate is stated in
terms of that counter.** So the gate cannot be evaluated, and on a fleet with legitimately un-MAC'd
members it is **unreachable by construction** — while the one cause the feature exists to detect (a
forgery) is indistinguishable from the benign majority.

Found while measuring #190's observe soak against its own acceptance criteria.

## The conflation

`rust/clock/src/net/wire.rs:433-437` names the causes itself:

```rust
    /// No accepted key verified the trailer: absent/truncated (legacy un-MAC'd frame), wrong key,
    /// wrong/unknown epoch, or a forgery. The caller counts this (`mac_fail`) and — depending on the
    /// observe-vs-enforce policy — either soft-accepts the raw frame or drops it (design §7.1).
    Fail,
```

Five causes → one variant. `net/mode.rs:6710` then folds them into one number:

```rust
                MacVerdict::Fail => {
                    self.diag.mac_fail = self.diag.mac_fail.saturating_add(1);
```

## Why that specific merge is load-bearing

`mode.rs:6693-6698`, eight lines above the merge:

> *"A group-key/epoch mishap manifests as leaf HELLO-drop → owner-churn, NOT a crown-deaf-shed
> (design §7.3) — **`mac_fail` is the telemetry that tells the two apart**, so it must be counted
> SEPARATELY from RF-deafness signals."*

The comment is right that `mac_fail` must be separate from deafness — and it *is*. But the decision it
is being asked to support needs a finer split than it has:

| cause | what it means | correct response |
|---|---|---|
| no trailer at all | a heterogeneous / legacy member | benign — teach it the codec, or exempt it |
| wrong key | a provisioning defect | **fix before enforce, or that board partitions** |
| unknown epoch | rotation skew | extend the overlap window |
| bad tag, right epoch | corruption **or a forgery** | the security event this feature exists to catch |
| trailer shorter than `MAC_TRAILER_LEN` | runt/truncated frame | RF or a malformed sender |

All five increment the same counter, so an operator reading `mf=145287` learns only *"something was
not verifiable 145,287 times."*

## Measured consequence on the live fleet

Steady-state, both boards where the counters survive into the published record:

| node | uptime | `mo=` | `mf=` | fail share |
|---|---|---|---|---|
| id8 | 55,414 s | 82,371 | 145,287 | **63.8 %** |
| id50 | 27,640 s | 7,409 | 71,915 | **90.7 %** |

#190 Wave 1's acceptance is *"`mac_ok` accrues, `mac_fail`≈0 steady-state"*, and the flip is gated on
it. Four of the nine mesh members (id122, id236, id162, id176) run watch-derived firmware which
**vendors `append_group_mac` and never calls it** (no callers outside its own definition; no
`GROUP_KEY` in its secrets), so their frames are legitimately un-MAC'd and permanently inflate `mf=`.

⇒ **`mac_fail ≈ 0` can never be true while those members exist**, whether or not anything is wrong.
A gate whose pass condition is unreachable is not a gate; and worse, the number cannot be *read down*
to "the un-MAC'd share is expected, the wrong-key share is zero", which is the statement Wave 2
actually needs before `MAC_ENFORCE` flips.

## This is the "a new failure source needs its own name" shape

When a new way to fail joins an existing outcome, reusing the old name corrupts the instrument that
was already there. `MacVerdict::Fail` was correct when the only case was "legacy un-MAC'd board".
Rotation (epoch), heterogeneous members, and forgery detection each arrived later and each landed in
the same bucket, so the counter now answers a question nobody asked while being unable to answer any
of the five that get asked.

## Proposed fix

1. **Split `MacVerdict::Fail`** into `Absent` (no/short trailer), `Epoch(u8)` (no accepted key for that
   epoch), and `Tag` (right epoch, wrong tag — the security-interesting one). All three still
   soft-accept under observe; the enforce policy stays a single `continue`, so **no behaviour changes.**
2. **Three counters in DIAG.** The tail is budget-declared, so this needs a `DIAG-TAIL` width update
   and `tools/check_diag_budget.py` will name the field until it is declared — which is the intended
   failure mode. If the tail cannot afford three, the minimum useful split is
   **`mf=` (absent — benign) vs `mfk=` (epoch+tag — actionable)**; one extra field buys the whole
   decision.
3. **Restate #190's Wave-1 gate against the actionable counter**, not the total: `mfk` steady at 0
   *and* `mac_ok` accruing from every MAC-capable member. That gate is reachable and means something.
4. Correct `mode.rs:6698`'s claim while there — it currently promises a discrimination the counter
   cannot make.

## Blocked, and by what

The counting site is `rust/clock/src/net/mode.rs`, currently held by the STEP-T lane
(mid-atomic-commit). **I have not touched it.** `wire.rs` (the enum) is free, but splitting the enum
without the counters is half a change, so this should land as one piece after T.

Refs: #190 (the gate this blocks), #382 (leaf truncation — hides `mo=/mf=` on leaves, so the soak is
not fleet-wide either), #371 (the vendored-but-uncalled watch codec is another specimen).

jphein:
Closes #471. #190's Wave-1 gate is `mac_fail ≈ 0` — a gate that could not be **evaluated** (one `u32` folding five causes) and could not be **passed** (four watch-derived boards vendor `append_group_mac` and never call it, so a benign share inflates the number forever). Measured fail share: id8 **63.8 %**, id50 **90.7 %**.

## The split is not where I first said it was

I proposed `mf=` = *"no trailer at all"* vs `mfk=` = *"wrong key / unknown epoch / bad tag"*. Implementing it, that turned out to be **unimplementable**: *"no trailer at all"* is not observable. A legacy frame does not announce itself — the verifier reads its last 9 bytes as `epoch || tag` and gets whatever payload was there.

I had been splitting along the doc comment's list of **causes**, which is a description of the *world*. The receiver only ever gets to ask one question:

> **Did this frame claim an epoch we accept?**

| DIAG | verdict | meaning | expected |
|---|---|---|---|
| `mf=` | `Unkeyed` | nothing claimed an accepted epoch — legacy un-MAC'd, or a runt | **large, non-zero forever** on a mixed fleet. Not an alarm. |
| `mfk=` | `BadTag` | **claimed an accepted epoch**, tag failed — wrong key, corruption, or a forgery | **zero**. This is the flip gate. |

## The actionable counter is *exact* — and that nearly went the other way

A legacy frame reaches `BadTag` iff its payload byte at the epoch offset equals an accepted epoch. Naively **~1 in 256**. On id50's `mf=71,915` that is ~280 phantom forgeries — and `mfk= ≈ 0` would have been unreachable for the *same structural reason* `mac_fail ≈ 0` already is. **I would have fixed the gate and reintroduced its defect at 1/256 scale.**

It is **zero** instead, because every SMOLv1 frame reaching the verifier is `SMOLv1 …` **ASCII** and `GROUP_KEY_EPOCH` is `1` — outside printable ASCII.

⚠️ **That is an accident nothing recorded, and it has an expiry date.** Rotation says to *bump* the epoch. Bumping reaches `0x20` (space) at epoch 32, and `SMOLv1 ` frames are full of spaces — that rotation would flood `mfk=` with benign traffic **indistinguishable from the attack it reports**, on the one counter somebody would page on. A compile-time assert now refuses an epoch that is `0` or printable ASCII.

**That guard is arguably worth more than the split**: the split makes the gate readable; the guard keeps it readable across a rotation nobody will remember this PR during.

### Where the assert lives, found the hard way

Putting it beside the verifier in `wire.rs` broke `relay_compat` — `cannot find 'secrets' in 'crate'`. **`wire.rs` is deliberately secrets-free so host verifiers can `#[path]`-include it standalone** (#367's coupling, already guarded here). So `wire.rs` exports the *rule* as `group_epoch_is_unambiguous()` carrying the rationale, and `mode.rs` checks the real secret against it. Rule with the code it describes; secret where the secrets are.

## ⚠️ `mf=` changes meaning

It was *"everything that failed"*; it is now *"the benign share"*. The pre-#471 quantity is **`mf= + mfk=`**. Recorded in the docs and the HA package, not only in this description:

- **`docs/protocol.md`** — the table above, `mfk= ≈ 0` named as the gate, and the non-Rust-implementer warning **corrected**: it said `mac_fail` distinguishes leaf-HELLO-drop from crown deafness, but an un-MAC'd HELLO lands in the *benign* class, so `mf=` alone never could.
- **`ha/packages/smol_telemetry.yaml`** — new `mac fail keyed` sensor; the existing one relabelled `(unkeyed, benign)` so a climbing value is not read as an incident. Verified `startswith('mf=')` does **not** match `mfk=` rather than assuming it.
- **`experiments/mac_verify`** — 9 assertions sharpened from *"it fails"* to **which class**: flipped payload/tag and wrong key are `BadTag` (the epoch byte survives, so the claim stands); flipped epoch, unknown epoch, runt and empty are `Unkeyed`.

## Cost — priced by the checker, not estimated

`|mfk=` (5) + `u32` (10) = **15 B**; margin **22 → 7 B**. #525's checker *caught the change and priced it*:

```
protected-tail widths sum to 84, but DIAG_CORE_MAX's remaining terms are 69 (+15)
```

The 30 B three-way split would not have fitted — exactly as #525 predicted. `DIAG-DERIVED` and the constant were updated **together**, which the same checker enforces.

## Verification

- **`tools/gate.sh host`: GATE GREEN.** It first went **red** on `mac_verify` (12 errors) *and* `relay_compat` — both real, the second being my own secrets coupling. Running the host gate locally before pushing is what surfaced them.
- **`mac_verify` passes, with two new proofs that the *design* holds** rather than that the plumbing compiles: a real legacy un-MAC'd ASCII frame classifies as `Unkeyed` against a safe epoch, and **the same frame** classifies as `BadTag` against the printable epoch it happens to carry. The hazard is demonstrated with a **real collision**, not a contrived one — so the guard is proved *necessary*, not merely plausible.
- `tools/test_diag_budget.sh` **18/18**. Its fixtures pin live values and correctly reported `fixture setup failed` when the margin moved 22 → 7 — the harness announcing drift rather than silently testing nothing. Re-pinned.
- `check_shed_order.py` unchanged (11 fields); `status_check.sh` rc=0; YAML parses.
- ⚠️ **The `fw` arm was not run locally: familiar is down** (ARP `INCOMPLETE`, two WoL attempts, while `disks` and `gatekeeper` answer — so it is that host, not the network). CI runs `tools/gate.sh fw` on this PR. Flagged rather than skipped quietly.

## Not closed by this PR

**#382.** Neither counter reaches HA from a **leaf** — the relay cut drops the entire protected tail (measured: 169 B of 381 B). So #190's soak stays crown-only until #382 lands, and this PR makes the counter *readable*, not yet *reachable*. Both are prerequisites for the observe gate; details on #190.

Closes #471 · Refs #190, #382, #525, #367


jphein:
## Cross-repo sweep result — safe to merge, but it uncovered a duplicate (#529)

Per the *"a clean merge means git had nothing to flag"* rule, I swept the **whole repo** for other users of `MacVerdict`, not just `rust/clock`. The first, scoped sweep found nothing outside `rust/clock` and `experiments/` — and would have been wrong.

**`targets/c6-watch/crates/mesh-flood/src/wire.rs` is a byte-identical copy of `rust/clock/src/net/wire.rs`** (`diff … | grep -c '^[<>]'` → **0**), with its own `MacVerdict::Fail`.

**This PR is still safe to merge:**

- `mesh-flood` is an **independent crate** — it does not `#[path]`-include `rust/clock`'s wire.rs, so nothing here changes how it compiles.
- The watch **never calls** `append_group_mac`/`verify_group_mac`. Verified across the subtree *and* the standalone repo: only doc-comment and code-comment hits inside the definition file, zero call sites.
  *(My first check of this searched `~/Projects/esp32c6-watch/src/` while the vendored copy lives under `crates/mesh-flood/` — wrong path, and it "confirmed" the claim by finding nothing. Re-run against both trees properly. I have cited that "vendors it and never calls it" claim in three places tonight, so it was worth getting right rather than inherited.)*

**What it does mean:** after this merges, the two copies **diverge for the first time**, and nothing reports it — the vendored file is protected only by *happening to be identical*, which is a coincidence, not a mechanism. Worse, its header claims `relay_compat` and `mac_verify` `#[path]`-include *it*; they both include the `rust/clock` path. A comment that asserts coverage reads exactly like a check that enforces it — the same shape this PR's sibling (#525) fixed in gate.sh.

Filed as **#529** with the measured baseline, rather than silently editing the subtree: per the standing ruling the watch's standalone repo is the working repo and `targets/c6-watch` is the delivery vehicle, so that content arrives by refresh PR, not by my hand.

Issue #510, opened 2026-08-27
design: ONE HELLO frame decision, not three — #278's cc bit, #165's txcost echo and #377's gateway class all need a full frame


jphein:
## The decision

**`HELLO` is exactly full. Three separate issues each need to add a field to it. That is one frame
decision and one fleet-compatibility event — not three.** This issue exists to make it one.

Filed as a **design-decision issue for post-STEP-T**, from the shift-4 backlog sweep. It carries no
implementation; it exists so the three riders below are decided together rather than discovered
one at a time.

## Why the frame is full — measured, not asserted

```
mode.rs   HELLO_PREFIX = b"SMOLv1 HELLO "      13 bytes
          broadcast_hello builds into          [0u8; 16]
          + id as 3-digit ASCII "NNN"           3 bytes
                                              ========
                                                16 / 16
```

`broadcast_hello` has no spare byte. And `HELLO`/`ACK` are the pair `docs/protocol.md` marks
**"⚠️ Do not alter — it is the hardware-verified LED path"**, so this is not a frame anyone should widen
casually.

**`HELLO2` does not exist.** `git grep -n HELLO2 -- rust/clock/src docs/` returns **nothing** — it is
discussed in issue prose (#371's disposition thread, #165's scoping comment) and nowhere else. Every
issue below is therefore blocked on a frame that has been proposed and never designed.

## The three riders

| # | issue | what it needs in the frame | why it cannot use an existing field |
|---|---|---|---|
| 1 | **#278** (ELECT) | a **`cc` bit** — "this board is co-channel-capable" | `net/coexist.rs:125` already documents `CrownCtx::better_successor_cc` as *"A co-channel-capable board (HELLO `cc=1`) is claiming"* — an API specified against a field that does not exist. `cc` today is a **DIAG** field (`coexist.rs:107`, and live on the wire: id8 publishes `\|cc=1\|`), which a *claiming* board cannot see in time |
| 2 | **#165** (best-relay via ETX) | a **`txcost` / IHU-equivalent echo** | smol's ETX is **one-way** (`rxcost` only — `net/etx.rs:28-32` says so). Relay selection on it picks the peer **we** hear best with no evidence it hears **us** — the exact asymmetric-link case Babel's two-way cost defeats. #164 is closed and ETX is live (`etx=73` observed on id8), so this is the *only* thing left blocking #165 |
[...]

Issue #465, opened 2026-08-26
OTA: gateway armdiag staged_raw goes stale after a new stage — and relays the OLD image to leaves (1 MB wrong-image relay observed)


jphein:
Two coupled defects from the S3 roll forensics (2026-08-26, #398):

1. **Display half (instrument honesty)**: after `smol/ota/staged` moved 922→1404→1405, the v922 crown's `ota/armdiag` kept `staged_raw=922` — so leaves whose staged knowledge arrives via crown relay showed `latest_version: 922` (a value that existed NOWHERE on the broker). Cost real diagnosis time: the leaf looked like it was refusing the new build when it had never been offered it.

2. **Relay half (the expensive part)**: during id162's install of build 1405, the crown mesh-relayed **the stale 922 C3 image** at the S3 leaf — `smol/162/ota/progress 0→1031008|relayfetch` matches smol-922.bin byte-for-byte (1,031,008 B), not the staged 1405 S3 payload (1,032,720 B). The leaf's guards refused it and its parallel self-fetch of the correct image won (`1032720|self` → `ota=confirmed:1405`), but that is a full megabyte of wrong-chip, wrong-build ESP-NOW airtime per leaf install, plus a masked failure path for leaves with no self-fetch capability.

Root-cause direction: the raw-announce persistence for leaf relay (`*staged_raw = Some(a)` pre-gate in current code) evidently doesn't refresh on the v922 crown when a newer line arrives — whether v922 predates the pre-gate persist or something gates the update, needs a source dig at the v922 tag. This is the display-vs-truth instrument class of the week's hunts.

Evidence: /var/tmp/roll1405.txt capture (katana), smol/50/ota/armdiag flips, #398 roll thread.

jphein:
## T-BLOCKED — file ownership checked, not assumed. No work started.

Both halves of this issue land in the STEP-T file set, which two executors hold mid-atomic-commit:

| half | fix site | held by |
|---|---|---|
| 2 (relay — the expensive part): the raw-announce persistence | `net/wifi.rs:3596` `*staged_raw = Some(a)`, with the pre-gate persist path documented at `:3491` | STEP-T (wifi.rs) |
| 1 (display/instrument honesty): `armdiag`'s `staged_raw=` readout | `net/wifi.rs:4265-4284` and `net/mode.rs:6168` / `:2206` | STEP-T (wifi.rs) + STEP-T2 (mode.rs) |

So there is no part of this that can be implemented from another lane without colliding. I have
[...]

PR #530, opened 2026-08-28
fix(ota): #465 both halves — a crown stops relaying and stops claiming what no leaf can judge

jphein:
Implements both #465 halves against the ruled invariants. **Both ruled *mechanisms* turned out to need something the crown provably does not have**, so this reaches the same invariants by a route that works — and says plainly where the ruled route is still blocked rather than shipping a gate that cannot fire.

## What the crown actually knows about a leaf

`stat_cache.entry()` yields `(id, key, relayed-STAT)`, and STAT carries a **build and a screen**. There is no chip associated with a leaf anywhere in `net/`. So both ruled mechanisms — *"skip leaves whose chip ≠ canonical"* (half-1) and *"compare `ann.target.chip` to the leaf's chip"* (half-2) — are **uncomputable here today**.

### And the ruled half-2 gate would not have fired on the incident it was ruled for

In the #398 capture: crown = **C3**, image = **C3** (922), leaf = **S3**. Any gate comparing the image's chip to the *crown's* chip **passes**. Post-#464 a crown only ever holds its own chip's staged line, so that comparison is near-tautological — **a gate that cannot fail, which is worse than no gate, because it reads as fixed.**

## The mechanism that actually explains the megabyte

`ota::gate` runs #349's pre-fetch suitability check **only** `if let Some(t) = a.target`. A legacy `OTA|` line carries no target, and that asymmetry is deliberate — *"this staging did not say"* is not *"suitable for anyone"* — so a board judging its **own** self-fetch falls back to the post-download descriptor read, letting a new publisher coexist with an un-rolled fleet.

**That justification does not survive being inherited by the relay path, and it was inherited silently:**

| | who pays | can the payer refuse early? |
|---|---|---|
| self-fetch | the board itself | yes — it judges at finalize, on its own budget |
| **relay** | **a different board** | **no** — structurally unable to refuse until all of it has arrived |

The cost moved to a third party; the justification did not move with it. The v922 crown was pre-#349, so its `staged_raw` **was** exactly such a target-less announce — which is why id162 could only refuse at finalize, after **1,031,008 B**, ×3 under `LEAF_OTA_MAX_RETRIES`.

## Half-2 — refuse to arm on a target-less announce

**This restores the leaf's pre-fetch veto by construction**: the only announces relayed are ones the leaf can judge *before a single body byte moves*. Plus defence in depth for a target that is not this crown's chip — near-tautological today, which is exactly why it's worth **asserting rather than assuming**: #464 is a property of the *publisher*, and this is the consumer who'd pay for the publisher changing.

The install is deliberately **not cleared** — this is a fact about the announce we currently hold, not a verdict on the leaf, so the next flush arms normally once `staged_raw` refreshes to an `OTA2|` line. Clearing would turn recoverable staleness into a lost install.

## Half-1 — a proxy must never overwrite a principal's own claim

Measured: `smol/162/ota/state` `latest_version` went **1405 → 922 and stuck**. Not staleness — a clobber. Same discriminator: with no target the crown cannot assert the build **of a particular leaf** at all, so it publishes no state for relayed leaves. **Discovery still publishes**, so the HA entity survives; only the unfounded claim is withheld.

⚠️ **The guard covers the whole publish, not the value.** Gating only the value leaves `latest_staged.unwrap_or(installed)` writing the topic anyway with `installed == latest` — a *different* wrong number, still overwriting the principal, and one that additionally **hides an available update** by rendering as up-to-date. I wrote the value-only version first; **reading the consuming line is what caught it.**

## Scope is narrow by construction

Every post-#349 staging is an `OTA2|` line, so **both gates fire only for a pre-#349 crown** — the measured v922 case. Nothing changes for a modern crown serving credential-less same-chip leaves.

## Still blocked — routed, not faked

An S3 leaf under a **modern** crown is not covered: the crown still cannot tell a self-publishing flavor from a credential-less one. That needs a **self-publishes bit (and the chip) in HELLO** — filed as **#510 rider 4**, since one frame decision unblocks both halves.

## One commit, not two — deliberately

Both halves live in the same file a few hundred lines apart; splitting them meant string-surgery on the working tree, and **my first attempt at exactly that tripped its own assertion**. Tonight already produced one commit message that lied because of a silent staging failure, so the trade wasn't worth it. Both halves are delimited by `#465` comments and revert cleanly by hunk.

## Verification

- `tools/gate.sh host` → **GATE GREEN**; fleet tier builds clean.
- The half-1 guard's placement **between** the discovery publish and the state publish confirmed by line number (5037 discovery → 5063 guard → state), not assumed.
- `merge-tree --write-tree` clean against `main`.

Refs #465, #349, #464, #510

### DIFF
--- a/rust/clock/src/net/wifi.rs
+++ b/rust/clock/src/net/wifi.rs
@@ -4231,6 +4231,55 @@ async fn mqtt_session(
     // install retained → the next flush retries (bounded by LEAF_OTA_MAX_RETRIES). `Announce`
     // is `Copy`, so the pair moves out cleanly.
     match (pending_leaf, *staged_raw) {
+        // #465: REFUSE to relay an announce that states NO TARGET.
+        //
+        // `ota::gate` runs #349's pre-fetch suitability check only `if let Some(t) = a.target`. A
+        // legacy `OTA|` line carries none, and that asymmetry is deliberate — "this staging did not
+        // say" is not "suitable for anyone", so a board judging its OWN self-fetch falls back to the
+        // post-download descriptor read and a new publisher can coexist with an un-rolled fleet.
+        //
+        // That reasoning does NOT survive being inherited by the relay path, and it was inherited
+        // silently. For a self-fetch the board spends ITS OWN airtime and flash-write budget on an
+        // image it will judge at finalize. For a relay the crown spends a DIFFERENT board's, on an
+        // image the leaf is structurally unable to refuse until the whole thing has arrived. The
+        // cost moved to a third party and the justification did not move with it.
+        //
+        // This is the measured incident, not a hypothetical: a v922 crown (pre-#349, so its
+        // `staged_raw` was a target-less `OTA|` announce) relayed a 1,031,008 B C3 image at the S3
+        // leaf id162, which could only refuse at the finalize descriptor read — a full megabyte of
+        // ESP-NOW airtime, ×3 under `LEAF_OTA_MAX_RETRIES`.
+        //
+        // Refusing here RESTORES the leaf's pre-fetch veto by construction: the only announces we
+        // relay are ones the leaf can judge before a single body byte moves.
+        //
+        // The install is deliberately NOT cleared. This is a property of the ANNOUNCE we currently
+        // hold, not a verdict on the leaf — when `staged_raw` refreshes to an `OTA2|` line the next
+        // flush arms normally. Clearing would turn a recoverable staleness into a lost install.
+        (Some(leaf_id), Some(ann)) if ann.target.is_none() => {
+            log::warn!(
+                "smol #465: NOT relaying build {} to leaf id{} — the announce states no target, so the leaf could not refuse it until the whole image had arrived (legacy OTA| line; install left retained for a targeted staging)",
+                ann.build,
+                leaf_id
+            );
+        }
+        // #465 defence in depth: an announce for a chip that is not even OURS has no business
+        // being served onward. Near-tautological today — post-#464 a crown only sees its own
+        // chip's staged line — which is exactly why it is worth asserting rather than assuming:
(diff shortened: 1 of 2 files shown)


jphein:
## CI's #390 arm went red — real, intended, blessed with the reason (`9717b08`)

`___embassy_main4POOL`: **12,728 → 12,736 B (+8)**. Host verifiers and tier exclusions both passed; this was the only failure.

**Why it happened, which is the part worth keeping.** The #465 guards add a local (`may_claim_for_leaves`) and two match arms inside `mqtt_session` — **an `async fn`**. Its future lives in the embassy task arena, and the arena `POOL` is sized by the tasks it holds, so a few lines of guard logic in an async context cost `.bss` in a way the same lines in a sync fn would not. **Nothing in the diff looks like it allocates**, which is precisely the class of change #390 exists to surface.

### And the stack paid for it, visibly

| | main (`4639e0c`) | this branch |
|---|---|---|
| `___embassy_main4POOL` | 12,728 B | **12,736 B** (+8) |
| canonical `.stack` | 73,088 B | **73,080 B** (−8) |

**Exactly the 8 B the POOL gained came out of `.stack`** — `esp-hal` shrinking the stack region as `.bss` grows. "Stack is not headroom" is normally an argument in this repo; here it is a measurement, on the same build, in the same run. Floor is 64,475 derived, so **8,605 B clear** and the stack-floor arm passes.

### Verified the same way as the last bless

familiar is still down, so the canonical ELF was built locally **through `repro_cargo_args`** (so its geometry matches the shipped one), and **before blessing anything I confirmed the local build reproduces CI exactly** — same symbol, same `12,728 → 12,736`, same 73,080 B stack. A bless taken from a build that disagreed with CI would just be moving the goalposts to wherever my machine happens to stand.

Baseline diff is two lines: the total and the one symbol. No other static moved.

Issue #511, opened 2026-08-27
[epic] External battery monitoring — smol's primary job: dumb packs (59V/48V/36V/24V/12V/4.2V) → HA


jphein:
**JP's statement of purpose (2026-08-27, recorded verbatim):** *"main use of smol for is to monitor battery voltage. not only of smols battery if it has one, but of a battery that does not have wifi or ble already. i have 59v, 48v, 36, 24, 12v, 4.2v etc dumb batteries that need monitoring."*

smol's primary job is distributed battery telemetry: a $1–$3 node wired to a dumb pack, reporting voltage (and derived %) over the mesh/MQTT into HA.

**Scope sketch (to be broken into child issues):**
1. **ADC sensing arm** — shares plumbing with #479's S3 own-cell arm (post-T). Per-chip ADC channel + attenuation setup; sampling/averaging; the DIAG field gets its own name (new-failure-source rule).
2. **Per-node divider config** — external packs up to 59V need external voltage dividers; the C3 ADC tops out ~2.5V full-scale. The divider RATIO (and an offset/cal term) must be **per-node runtime config** (retained CFG key, same discipline as key Y), never a baked const — one firmware, any pack.
3. **Chemistry/range presets** — pct estimation per pack type (li-ion serial counts: 1s 4.2V, 3s 12.6, 6s 24, 10s 36/42, 13s 48/54.6, 14s 59? — plus lead-acid 12/24/48 curves); preset selectable per node via config; alert thresholds (low-voltage cutoff warning) as HA-side or node-side rules.
4. **HA surface** — per-node `_voltage` (and `_soc`) typed-child sensors: #498's split is the delivery mechanism.
5. **Hardware note** — divider + protection (input impedance, fusing for 59V packs) documented per range in docs/; safe-wiring section (JP-reviewed).
6. **Bench validation** — one sitting per range class with a bench supply/known packs; observed-sufficient semantics per reading.

**Sequencing:** post-STEP-T (the DIAG composer and net path are mid-rewrite); pairs naturally with #479's fleet arm and #498. Filed as an epic so the matrix gains a row family (external-battery sensing) under the superset rule.

Refs: #479, #498, the capabilities matrix (docs/CAPABILITIES.md).

jphein:
## Sanity-check against the tree — **the scope is materially SMALLER than the body implies, and there is one safety-shaped requirement missing.** Four corrections, one addition.

Re-derived by content on `origin/main`. The epic's framing is right; three of its six items already have
[...]

PR #534, opened 2026-08-28
feat(diag): #382 N-part DIAG continuations — part A byte-identical, and `cut=` becomes a checksum

jphein:
**Not one leaf DIAG has ever arrived complete.** 64 relayed records, 2026-07-28 → 08-26: **43 % of the record dropped**, loss growing **2.6 B/day**, and what goes first is the entire **protected tail** — `mo=`/`mf=`/`mfk=`, unsheddable against the 495 B MQTT cliff and destroyed every single time by the 218 B relay cut. **Two ceilings, one word:** a field can be protected against one and reliably destroyed by the other.

## Design (as ruled)

**N-part on the wire, one *declared* continuation in the crown.** The wire format is open-ended (part digit, no fixed count); the crown buffers one continuation whose payload is a **named field set** — the protected tail plus `elect=` and the positional fields that fall past part A's cut.

> **The wire decision and the memory decision get separated, which is what makes the expensive half deferrable.**

Full reassembly stays reachable as a **buffer-size change, not a protocol change**.

**Why not two-part:** 436 B of capacity against a 389 B record growing 2.6 B/day is **~18 days of headroom** — a fix that expires before it is a month old. Any *fixed* part count has an expiry date; a part digit does not.

**Part A is byte-identical**, `|cut=` included. HA, the crown relay, every existing parser see exactly what they see today — **no flag day, no rolled-fleet dependency**, which `protocol.md`'s OTA2 migration table shows is the expensive kind of change. A receiver that doesn't know `DIAGC` ignores it as it ignores any unknown frame.

**`cut=` stops being an epitaph and becomes a checksum.** The crown strips part A's tag, appends, and re-derives it from what is *still* missing. **`|cut=0` means you have the whole record** — the instrument that reported the bug is the one that proves the fix.

## Pricing basis — `.bss` is the expensive currency, measured not assumed

PR #530 grew the embassy arena **+8 B** and the canonical ELF's `.stack` fell **73,088 → 73,080 — by exactly 8**, same build, same run. `esp-hal` takes `.bss` out of `.stack`, 1:1. C3 headroom is **8,605 B**.

| option | `.bss` | |
|---|---|---|
| **`DIAG_CACHE_VALUE_MAX = 360`** (part A + one declared continuation) | **+1,536 B** | **this PR — ~18 %** |
| 448 B/entry (full reassembly of the measured 389 B record) | +2,592 B | ~30 % |

`RelayCache` is now generic over its value width; **`scan_cache` keeps the one-frame default** and only `diag_cache` widens. The primary use case pays its own freight: **#511's battery telemetry adds leaf DIAG fields, which at 43 % positional loss would be cut on arrival by construction** — continuations are its delivery path.

## Three enforcement details

**1. Parse order is load-bearing, and asserted.** `DIAG_PREFIX` is a **strict prefix** of `DIAGC_PREFIX`, so a parser trying DIAG first matches *every* continuation as a DIAG whose record begins `C<part>` — **a well-formed frame silently decoded as a malformed one, the worst shape of parse bug because nothing errors.** The DIAGC arm goes first; a compile-time assert proves the prefixes still nest so the note can't quietly stop being true.

**2. The pairing is enforced, not documented.**

```
OBS_VALUE_MAX + DIAG_CONT_PARTS * CONT_DECLARED_MAX <= DIAG_CACHE_VALUE_MAX + CUT_TAG_MAX
```

Raise the parts count without raising the cache and the **build fails**. Sabotage-proved: `DIAG_CONT_PARTS = 2` → `E0080` with its own message.

*On the CAP ruling:* the part-A/part-B **slot skew is impossible by construction**, not by assertion — a continuation is appended into part A's *own* cache entry, not a parallel array. One `RELAY_CACHE_CAP`, one entry, one record. Recorded beside the assert so nobody adds an equality check with nothing to compare.

**3. A bug in my own first draft, found implementing the ruling.** The sender bounded continuations by what the **frame** holds (224 B); the crown can only keep **134 B** after part A. A full-frame continuation was **silently truncated at the receiver** — positional loss reintroduced at the exact place this change removes it, invisible because the frame was well-formed. `CONT_DECLARED_MAX` is now *derived from what the crown can keep*, and the sender's window is `min(frame, declared)`. **Byte-granularity form of the same rule: never emit what the other end cannot keep — airtime spent on data dropped on arrival is worse than not sending it, because it also looks like it worked.**

## Airtime (approval condition, measured before any code)

One 250 B frame per leaf per **60 s** `DIAG_CADENCE_MS` — leaf-only; crowns self-publish via the flush. *(My proposal originally said "per 30 s flush"; that's the gateway timer. The real cost is 4× smaller.)*

| PHY | fleet/min | duty |
|---|---|---|
| 1 Mbps DSSS, long preamble (worst) | 15,344 µs | **0.0256 %** |
| 24 Mbps OFDM | 723 µs | 0.0012 % |

One leaf mesh-OTA moves ~1 MB through the same radio, so a continuation is **0.024 % of one transfer**.

## Tested where it can be

The `|cut=` helpers are pure byte functions, so they live in **`net/wire.rs`** — the dep-free host-includable codec — rather than beside their caller in `net/wifi.rs`, which would have made them untestable off-target (the #367 coupling, applied deliberately this time). `experiments/relay_compat` now round-trips every tag width **including the 6/7/8-byte length boundaries**, asserts the 999 clamp matches `broadcast_diag`'s `dropped.min(999)` (disagreement there corrupts the record rather than merely mis-stating it), and pins `find_last_cut` to the **last** occurrence.

⚠️ **That last arm was wrong at first.** Sabotaging `find_last_cut` to take the FIRST occurrence left the suite **green** — my fixture's decoy was `|note=cut=here`, which contains no `|cut=` at all, so the record had exactly one match and **the arm tested nothing.** Fixed to two genuine `|cut=` fields; both sabotages now fail for the right reason. **A test written to check a property does not automatically check it — the fixture has to contain the thing.**

## ⚠️ Host-green is NOT proven — the witness is scheduled, not skipped

This is a **mesh wire-format change** and I have no flash authority; familiar is down. `gate.sh host` is GREEN and `relay_compat` covers the byte contract, but **neither is evidence that a leaf and a crown agree over the air.**

**Bench witness, pre-approved for the first post-roll window:**
1. a leaf's `|cut=` reaching **0** on a reassembled record;
2. **`mo=`/`mf=`/`mfk=` arriving at HA from a leaf for the first time** — simultaneously #382's proof and **#190's Wave-1 unblock**.

Refs #382, #190, #306, #471, #511, #530

### DIFF
--- a/experiments/relay_compat/src/main.rs
+++ b/experiments/relay_compat/src/main.rs
@@ -55,6 +55,11 @@ fn old_parse_relay(data: &[u8]) -> Option<(u8, u16, u8, u8, &[u8])> {
     Some((src_id, msgid, rest[10] - b'0', rest[12] - b'0', &rest[14..]))
 }
 
+/// Expected `|cut=<n>` width: 5 for the tag + the decimal digits, clamped at 3.
+fn alloc_len(n: usize) -> usize {
+    5 + if n >= 100 { 3 } else if n >= 10 { 2 } else { 1 }
+}
+
 fn main() {
     // (1) NEW frame -> OLD parser (a #13 leaf's H=1 RELAY on an old gateway) -----------------
     let mut fb = [0u8; 128];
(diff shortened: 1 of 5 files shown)


jphein:
## CI walked down to green in three steps — all three were real, and the last one validates the design's pricing

**1. `check wifi` + clippy on 12 tiers (E0433).** `net::wire` is `espnow`-gated (`net.rs:97`); `append_cont` lives on `RelayCache`, a **`wifi`** item — so a wifi-only build has no `wire`. Fixed with `#[cfg(feature = "espnow")]` on `append_cont`, which is the **correct shape, not a workaround**: a continuation is a mesh frame, so a build with no mesh has nothing to continue.

⚠️ Worth naming: I moved those `|cut=` helpers into `wire.rs` *specifically* to honour #367's host-testability coupling — and that move is what put them behind the `espnow` gate. **Respecting one constraint moved me into violating its neighbour.**

**2. clippy `manual_try_fold`** under `-D warnings` on `parse_cut`'s `fold(Some(0), |acc, b| acc?…)`. Clippy is right — the short-circuit *is* the point, and burying `?` inside a `fold` hides that the accumulator abandons early. Now `try_fold`; overflow still yields `None`, covered by relay_compat's malformed-tag arms.

**3. #390 baseline: `RADIO` 17,232 → 18,768 B (+1,536).** Blessed at `4d74dc7`.

### That third number is the interesting one

It is **not a surprise** — it is `RELAY_CACHE_CAP(12) × (DIAG_CACHE_VALUE_MAX 360 − RELAY_VALUE_MAX 232) = 1,536`, the figure quoted on #382 and in this PR's body **before the code existed**. Estimate and measurement agree **to the byte**.

**And the 1:1 rule held at 192× the scale it was measured at:**

| | before | after |
|---|---|---|
| `RADIO` (`.bss`) | 17,232 B | **18,768 B** (+1,536) |
| canonical `.stack` | 73,080 B | **71,544 B** (−1,536) |

That relationship was observed on an **8 B** change in #530 and used, an hour later, to price *this* design decision — full reassembly at +2,592 B was rejected as ~30 % of the C3's headroom. It has now been confirmed on a change three orders of magnitude larger. **The model that made the decision is the model the compiler agrees with.**

Headroom after: `71,544 − 64,475 = 7,069 B`. Stack-floor and paint-warrant arms both pass.

### Why I missed 1 and 2, and what changed because of it

I built with `--features espnow,cast,io` and ran `gate.sh host`. **Neither touches the other ten tiers, and neither runs clippy on firmware** — that is `gate.sh fw`, which needs a toolchain run unavailable while familiar is down, so the fw arm has been CI-only.

I stated in this PR that host-green is not proof of over-the-air behaviour. **It was also not proof of the tier matrix, and that gap I did not state** — I was enumerating my known limitations while carrying an unstated one, and the unstated one fired.

The fix is procedural and is in `0996358`'s message: while familiar is unavailable, a change under `rust/clock/src/net/**` gets a **local four-tier `check` + `clippy -D warnings` sweep before push** — ~2 minutes. Run here, all four green, and run *before* the fix too so the red was seen for the right reason.

All three blesses tonight were taken from a local build **first proved to reproduce CI exactly**. A bless from a build that disagreed with CI is just moving the goalposts to wherever the local machine happens to stand.

PR #545, opened 2026-09-22
docs(s3): where the Tapstone shrine screen lives — survey + recommendation (no code)

jphein:
Survey requested by the lead after tapstone decision 0010's amendment established that the framebuffer that ruling assumed does not exist in `rust/clock`. **Read-only: no code, nothing built, nothing flashed.**

## The headline

**The choice is three options, not two — and this is really `DISPLAY-PACKAGE.md`'s §5 question 3, "does the S3 colour/touch UI belong to smol or to the watch codebase?", asked 2026-08-24 and never answered.** Tapstone is the first thing that cannot proceed without an answer, so it needs one now.

| home | 320×240 | colour | renders locally | registry | status |
|---|---|---|---|---|---|
| `esp32c6-watch` + `board-esp32s3-cyd` | ✓ | ✓ | **Slint** | ✓ | ships the scry station |
| `smol/rust/clock` | via `s3_oled` | ✗ 1-bit 72×40 | ✓ | ✓ | where the engine + `AppKind::Tapstone` are (#543) |
| `smol/targets/s3-cyd/spike-scry` | ✓ | ✓ | ✓ | ✗ | a **spike**, own workspace |

## Three findings that change the shape of the decision

**1. "The GUI flavor" names two unrelated axes.** `PARITY.md` files the scry station under GUI on the *radio* axis ("a 5 s HTTP kiosk wants a held association"), and `grep -c slint` on `spike-scry/Cargo.toml` returns **0**. So "put it in the GUI flavor" does not resolve to a rendering decision at all. Read on the *rendering* axis it means the watch codebase — which is the renderer 0010 rejected on measurement (74–79 ms/flip, ~13 fps).

**2. Nebula's Path B was priced for the wrong job.** It costed porting smol's **171 `BinaryColor` sites** to colour, plus `smol-core` extraction (#347) and a touch model. Tapstone brings its own renderer, geometry (0027) and palette, and reuses **none** of those screens — so three of Path B's four prerequisites drop away, and the fourth (layout) is already done by Luna. That is the central finding.

**3. The 150 KB framebuffer is a choice, not a consequence.** The worry was that it lands in PSRAM and makes #445/#447's "PSRAM registered FIRST" load-bearing for an app. Two precedents on this exact glass say otherwise: `spike-scry` **never buffers a frame** by explicit design ("would cost 150 KiB and add a copy for nothing"), and `BOARD.md`'s house rule is strip-rasterise + windowed `fill_contiguous`. Band rendering keeps it in internal SRAM.

## Recommendation

Keep `AppKind::Tapstone` in `rust/clock`; add an **app-private colour surface beside `Oled`**, not instead of it (all 171 sites untouched, #152's gate intact); gate it on a **declared capability** — `has-color-panel` selected by `esp32s3`, the model `budget.rs` endorses ("predicate on a declared capability, never on a chip name"); and **lift the panel path from `spike-scry`** rather than inventing it.

This *resolves* rather than defers the C3 question: one app, two surfaces — the 1-bit lane counter on the C3, the colour battlefield on the S3 — which is exactly what issue 3's third acceptance criterion already asked for.

## The work item neither seam avoids

Porting `tapstone/rust/shrine-preview` off `SimulatorDisplay`: **12** fns taking `&mut Panel`, **5** type references, ~**20** `format!`, **0** fns currently generic over `DrawTarget`. Modest and mechanical — and it has a **free oracle**: `SimulatorDisplay` implements `DrawTarget`, so after genericising, the committed PNGs under `tapstone/preview/` must come back byte-identical. Same structure as #543's golden replay, where the transcript checked the parser.

Suggested order, whichever seam wins: genericise the renderer in tapstone first (proven by PNG equality, no seam decision needed, decouples the two), vendor it with the machinery #543 already built, then build the surface.

## What needs deciding (§5)

1. smol or the watch — a JP-level call about two codebases.
2. Capability feature vs chip feature (this would be smol's first *app* gated on a capability).
3. Band rendering vs a PSRAM frame — deserves one bench measurement before it is committed to.

§6 lists what the survey does **not** claim; §7 is an honesty ledger separating what I measured from what I cite from what I inferred. Nothing here was run on hardware by me, and the fps estimate is arithmetic on someone else's numbers and says so.

Companion to `DISPLAY-PACKAGE.md`. Part of tapstone phase 2.


jphein:
Two follow-ups pushed in `c8a0a63`, both requested after review.

**`TAPSTONE-SEAM.md`** now records what was decided — capability feature (decided), bands (decided *provisionally*), ownership (open, with JP) — and folds in Luna's independent costing: a lane column at ~35% of the frame budget as one contiguous window against ~97% split per row. That is a third line of evidence from a third direction, and it agrees: the window **count** dominates, not the pixel count. The band choice is now marked provisional in the body text rather than only in the honesty ledger.

**`TAPSTONE-BAND-BENCH.md`** is the measurement design. Two things it establishes that the survey did not:

**1. The real question is narrower than "are bands fast enough" — it is a tension between two constraints pulling opposite ways.** Window count wants *few* bands (`BOARD.md` measured per-cell windows at 2× a full-screen repaint; Luna's 35/97 is the same effect). Internal SRAM wants *many*: a band is `640 × rows` bytes against `budget.rs`'s **96,676 B** of free DRAM.

| bands | rows | strip buffer | of free DRAM |
|---|---|---|---|
| 1 (full frame) | 240 | 153,600 B | **impossible** — PSRAM only |
| 4 | 60 | 38,400 B | 40% |
| 8 | 30 | 19,200 B | 20% |
| 16 | 15 | 9,600 B | 10% |

The bench picks a row in that table. It does not answer a yes/no.

**2. 0010's ~15 fps applies to the animated band, not to a full repaint.** A full redraw is triggered by a card landing on the reader — human-paced, where 100–150 ms is invisible. So **M4** (one lane column, rasterise + blit) is the only measurement that gates the design, it is the cheapest of the five, and it runs first. If it comes in comfortably, the PSRAM question is moot regardless of the other four.

Each of the five measurements names **what result would change the recommendation**, written in advance so a run cannot be read as agreeing with the conclusion already on disk — including the result that would overturn it outright (M4 > 66 ms at every band count). **M0 is a control on the timer itself**, before any figure is trusted: this lane has had two instruments lie in a single day — an `nm` that reported zero symbols for *everything* because a shell had not sourced `export-esp.sh`, and a probe that read ~2 KB because the optimiser had const-folded away the call being measured. Both were caught only by a control; neither by inspection.

**Safety, since the harness lives next to live hardware.** The target is `targets/s3-cyd/spike`, armed byte-exact for node **162**, the dev board — *not* `spike-scry`, which is armed for node **163**, the live scry station JP taps cards on. The sealed reliquary unit shares node 162's first four octets, so those guards' byte-exact matching must never be widened to a prefix. §6 is the ask to JP in three lines, and it is not scheduled.

One asymmetry found while writing this and deliberately **not** changed here: `spike-scry/flash.sh` deny-lists node 162, but `spike/flash.sh` does not deny-list node 163. Safety holds either way — both resolve by a byte-exact *allow* list, so the deny list is belt-and-braces — but the protection reads as mutual and is one-directional. Not urgent, and not something to edit while arming a flash.

Still no code, nothing measured, no board touched.

jphein:
Updated in `05d3c65` from Luna's band-invariance result — and one of the two updates is a **correction to this design**, not an input to it.

## M0 was a blind control

It proposed timing `Delay::delay_millis(50)` with the same `Instant` path every other measurement uses. Those almost certainly share a clock source, so **a misconfigured timebase scales both by the same factor** — the control reads a perfect 50 ms while every figure below it is wrong by that same ratio. It tested whether the timer agreed with *itself*.

That is precisely the failure Luna hit and disclosed on the renderer side: a negative control drawing at a fixed absolute point, which the band translation corrects and clips, so it behaved identically banded or not — six passing invariance tests sitting behind a control that could not fail. **The generalisation is worth more than either instance: a control must fail for the reason the *mechanism* admits, not the reason that first comes to mind.** Here the mechanism is not "the timer is broken" but "the timer is fine and the timebase is wrong", which is live on this chip because `CpuClock` is configurable.

Replaced with two controls that can fail:

**M0a — the physical floor, needing no external reference.** A 320×240 Rgb565 frame is 153,600 B = 1,228,800 bits. The panel SPI is configured at 40 MHz (`spike-scry/src/main.rs:102`, and `BOARD.md`'s pin table agrees), so the wire cannot deliver a frame in under **30.7 ms** before a single command byte of overhead. That floor is computed from first principles and is *independent of the chip's own sense of time* — so a measured blit below it is proof the timebase is wrong, not that the panel is fast.

**M0b — an off-board reference**, for the uniform scaling error M0a can only bound: the board waits 10 s by its own reckoning between two printed markers, and the **host** timestamps both. katana's clock is genuinely independent, and at 10 s the USB latency is noise.

> **M0a already has something to resolve, before being run once.** The ≈27 ms full-frame figure this document has been citing is **below the 40 MHz floor**, so it is not achievable as stated. One of {27 ms, 40 MHz, "it was a full 320×240 frame"} is wrong — the letterboxed 288×160 image would be 18.4 ms, which is not 27 either, and at 80 MHz the floor would be 15.4 ms and 27 ms unremarkable. M1 is now that cross-check rather than a comparison against a number below the floor.

## Band count is a pure memory decision

Luna **proved** band-invariance rather than building it: byte-identical against a one-pass render at **2, 4, 7, 8, 9, 12, 16 and 24** bands, with 7 and 9 deliberately not dividing 240 so the last strip is short. A renderer that only worked for divisors would have silently limited band choice to factors of 240, and nothing would have revealed it until a memory budget pointed at seven.

So the renderer side of §1's tension is gone, and the memory table is the whole question. **M2 is now purely a cost measurement** — it can price banding, not change the design.

## ⚠️ And an invariant the seam must not break

Both documents now record why band-invariance holds, because it is one careless edit from being lost: **every entry point positions from panel coordinates and never asks the target where it is** — no `bounding_box()`, `size()` or `dimensions()` in the drawing code. One of those would make each band draw its own copy of whatever was positioned from it, and **each band would still look plausible in isolation**. That is why byte equality against a one-pass render is the check and reading the diff is not, and why whoever implements the seam must keep it running: a firmware author adding an innocuous helper has no reason to suspect the constraint exists.

Still no code, nothing measured on hardware, no board touched. Item (a) remains with JP.

